PR #269, opened 2026-04-18
fix: add missing default_voucher_series column to company_settings

jakobwennberg:
## Summary
- Frontend (`/settings/bookkeeping`), `types/index.ts`, and `lib/api/schemas.ts` have all referenced `company_settings.default_voucher_series`, but no migration ever added it. Saving the "Standardserie för verifikationer" setting failed with a PostgREST schema-cache error.
- Adds the column as `text NOT NULL DEFAULT 'A'` with a CHECK constraint mirroring the existing Zod validation (`^[A-Z]\$`).
- Ends with `NOTIFY pgrst, 'reload schema'` per CLAUDE.md rules.

## Why this slipped through
The column was referenced in the type, schema, and form without a matching DDL change. Every hosted user who opened the Bookkeeping settings page and hit Save would have hit this.

## Production state
Migration **already applied** to prod (`pwxtzglxptnnvjrpixpg`) to unblock a reporting user mid-session. This PR is the code-of-record catch-up — the file just needs to land in `supabase/migrations/` so other environments (self-hosted, preview branches, fresh boot) pick it up.

## Test plan
- [ ] Open `/settings/bookkeeping`, change the Serie dropdown from A to (e.g.) B, click Save — confirm it saves without error.
- [ ] Reload the page — confirm the new value persists.
- [ ] Try writing `'AA'` or `'1'` via direct SQL — confirm the CHECK constraint rejects it.
- [ ] Fresh Supabase boot (or branch) applies this migration cleanly without needing `IF NOT EXISTS` escapes.

### DIFF
--- a/supabase/migrations/20260418150000_company_settings_default_voucher_series.sql
+++ b/supabase/migrations/20260418150000_company_settings_default_voucher_series.sql
@@ -14,6 +14,20 @@
 ALTER TABLE public.company_settings
   ADD COLUMN IF NOT EXISTS default_voucher_series text NOT NULL DEFAULT 'A';
 
+-- ADD COLUMN IF NOT EXISTS silently skips the whole column definition
+-- (including NOT NULL) when the column already exists. Re-apply the
+-- constraint explicitly so environments that received the column
+-- out-of-band still end up with the same schema.
+ALTER TABLE public.company_settings
+  ALTER COLUMN default_voucher_series SET DEFAULT 'A';
+
+UPDATE public.company_settings
+  SET default_voucher_series = 'A'
+  WHERE default_voucher_series IS NULL;
+
+ALTER TABLE public.company_settings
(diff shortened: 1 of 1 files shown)


jakobwennberg:
Good catch from Greptile — verified prod (`is_nullable: NO`, default `'A'::text`) so nothing's diverged today, but pushed a hardening commit so the migration can't silently skip `NOT NULL` / `DEFAULT` on any environment that got the column out-of-band:

- `ALTER COLUMN ... SET DEFAULT 'A'` — idempotent, reasserts default
- `UPDATE ... SET default_voucher_series = 'A' WHERE IS NULL` — backfill so `SET NOT NULL` can't fail on legacy rows
- `ALTER COLUMN ... SET NOT NULL` — idempotent, reasserts NOT NULL

All explicit, all idempotent. Fresh boots go through the original `ADD COLUMN` path unchanged.

PR #308, opened 2026-04-22
Prompt to activate missing BAS accounts at commit

### LINKED ISSUE #304: Accounts not found
When a user tries to book to an account that is not yet enabled, they get an error, but no prompt to enable the account. Fix this in the best UX way possible

jakobwennberg:
Closes #304.

## Summary

Booking to a BAS account that isn't active in the user's chart used to throw a generic 400 `"Account(s) not found: 5010"` with no way forward — users had to leave the form and activate via `/bookkeeping` → BAS-katalog, then re-enter everything. This change turns that into a one-click activate-and-retry dialog right where the user is.

## How it works

**Backend — structured error contract**
- `lib/bookkeeping/errors.ts` *(new)* — `AccountsNotInChartError` carrying the missing numbers + `code='ACCOUNTS_NOT_IN_CHART'`, plus `accountsNotInChartResponse()` that returns a 400 with `{ error: { code, message, account_numbers } }`.
- `lib/bookkeeping/engine.ts` — `resolveAccountIds` now filters on `is_active=true` (so deactivated accounts are treated the same as never-added ones) and throws the typed error instead of a string.
- `lib/core/bookkeeping/storno-service.ts` — same treatment for the corrected-entry path.
- Every API route that calls the engine catches the typed error and returns the structured response (journal-entries POST / reverse / correct; transactions book / match-invoice / match-supplier-invoice / uncategorize; invoices mark-paid; supplier-invoices POST / mark-paid / credit; salary runs correct; import/opening-balance/execute; pending-operations/commit).
- `/api/bookkeeping/accounts/activate` now also reactivates existing-inactive rows (UPDATE `is_active=true`), not only INSERTs. Returns `{ activated, reactivated, skipped, unknown }`.
- New `GET /api/bookkeeping/accounts/bas-lookup?numbers=…` resolves BAS names server-side so the client doesn't need to bundle the full 1,276-account catalog.
- `lib/errors/get-error-message.ts` — maps `ACCOUNTS_NOT_IN_CHART` to a clear Swedish message for the fallback toast path.

**Frontend — shared UI**
- `components/bookkeeping/ActivateAccountsDialog.tsx` *(new)* — lists missing accounts with BAS names (`5010 · Lokalhyra`), flags any unknown/non-BAS numbers separately, single **Aktivera och bokför** action.
- `lib/hooks/use-submit-with-account-activation.ts` *(new)* — wraps an async submit. On `ACCOUNTS_NOT_IN_CHART` it opens the dialog, activates on confirm, then retries the original submit automatically so the user never re-enters data.
- `components/bookkeeping/AccountCombobox.tsx` — relaxed to accept any typed 4-digit numeric value (not just items in the active chart). The dialog handles everything the combobox lets through.
- `components/bookkeeping/JournalEntryForm.tsx` — wired to the hook + dialog via the existing review/confirm flow.

## UX

Type `5010` into a line, fill amounts, click **Granska & skapa** → **Bekräfta & skapa**. Instead of the old error toast you see:

```
Aktivera konton
Följande konton behöver aktiveras i din kontoplan innan bokföringen kan slutföras.
  5010   Lokalhyra
[Avbryt]  [Aktivera och bokför]
```

One click activates the account and posts the verifikation. Zero navigation away, zero re-entry.

Non-BAS numbers (e.g. `9999`) get flagged as *okända konton* and the confirm button is disabled — the dialog never silently creates a row for an invalid BAS number.

## Scope / out of scope

- The dialog is wired only into `JournalEntryForm` for now. Supplier-invoice approval, invoice mark-paid, transaction booking etc. all *return* the same structured error — they currently surface it as the Swedish toast *"Följande konton behöver aktiveras: …"* via `getErrorMessage`. Wiring the dialog there is a small additive follow-up.
- The `AccountCombobox` dropdown still only lists active accounts. Showing inactive / not-in-chart BAS entries inline with an "Aktivera" action is a richer but larger change — the infrastructure (hook + dialog + endpoints) is in place to do it in a later PR.
- No DB migrations, no RLS changes.

## API contract change

Routes that call the engine and hit an inactive/missing account used to return `{ error: "Account(s) not found in chart of accounts: 5010" }`. They now return:

```json
{
  "error": {
    "code": "ACCOUNTS_NOT_IN_CHART",
    "message": "Följande konton behöver aktiveras: 5010",
    "account_numbers": ["5010"]
  }
}
```

`getErrorMessage()` recognises both shapes. No existing test was asserting on the old string (whole suite still green). External consumers reading `body.error` as a string will now see `[object Object]` for this one error code — if any exist, they can read `body.error.message` or upgrade to handle the structured code.

## Commits

- `feat: prompt to activate missing BAS accounts at commit` — the change itself
- `docs: sync CLAUDE.md with current codebase state` — unrelated doc catch-up that had been sitting in the working tree (updated counts, new integrations, salary/providers modules, extension list, etc.)

## Test plan

- [x] `npm run lint` — clean for touched files
- [x] `npx tsc --noEmit` — clean for touched files
- [x] `npx vitest run` — 2277 / 2277 pass
- [x] `npm run build` — succeeds
- [ ] Manual: create a manual journal entry referencing an unactivated BAS account (e.g. `5010`), confirm the dialog appears, confirm activation + retry posts the verifikation
- [ ] Manual: type `9999` (non-BAS), confirm it shows as *Okänt konto* and the confirm button is disabled
- [ ] Manual: deactivate an already-active account via `/bookkeeping` > BAS-katalog, then try to book to it — should trigger the same activation flow (reactivation path)
- [ ] Manual: cancel the dialog mid-flow — no partial state, form stays open, no error toast

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -6,7 +6,7 @@ gnubok is a Swedish-focused accounting SaaS for sole traders (enskild firma) and
 
 **Tech stack**: Next.js 16.1.5 (App Router), React 19.2.3, TypeScript 5 (strict), Zod 4, Supabase (PostgreSQL + RLS + email/password + TOTP MFA auth), Tailwind CSS 4 + shadcn/ui, Vercel hosting, Docker (self-hosted).
 
-**Integrations**: Enable Banking (PSD2), TIC Identity (company lookup), Anthropic SDK, OpenAI (embeddings), Resend (email), Sentry (error tracking), Svix (webhooks), web-push (notifications), JSZip (archive export), sharp (image processing), Framer Motion (animations).
+**Integrations**: Enable Banking (PSD2), TIC Identity (company lookup), Anthropic SDK, AWS Bedrock (`@aws-sdk/client-bedrock-runtime` for inbox smart-match), OpenAI (embeddings), Resend (email), Sentry (error tracking), Svix (webhooks), web-push (notifications), Upstash Redis + Ratelimit, Google Drive (cloud backup via OAuth), JSZip (archive export), sharp (image processing), Framer Motion (animations), Recharts (charts), PDF.js (`pdfjs-dist`), `@react-pdf/renderer` (invoice PDFs), xlsx, fuse.js (fuzzy search), ics (iCal feeds).
 
 **Path alias**: `@/*` maps to the project root. **Language**: All code, comments, and commit messages in English. **License**: AGPL-3.0-or-later.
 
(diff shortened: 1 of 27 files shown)


jakobwennberg:
Pushed `83e0625` addressing the review. Summary of what I applied vs. pushed back on:

### Applied

**Greptile P1 #1 — `canConfirm` guard** (`components/bookkeeping/ActivateAccountsDialog.tsx`)
Confirm is now disabled when any entered number isn't a valid BAS account. The dialog never activates a partial set.

**Greptile P1 #2 — non-blocking regression** (`app/api/pending-operations/[id]/commit/route.ts`)
`commitSendInvoice` and `commitMarkInvoiceSent` no longer re-throw `AccountsNotInChartError`. Invoice delivery proceeds as before; the JE creation stays non-blocking. `TODO` comment left for when the activation dialog is wired into these flows.

**Compliance — `reverseEntry` blocked by inactive accounts** (`lib/bookkeeping/engine.ts`)
`resolveAccountIds` gained an `includeInactive` option. `reverseEntry` passes `true` so storno of a committed entry goes through even when a user has since deactivated one of its accounts — required by BFL 5 kap 5§. `createDraftEntry` keeps the default (`false`) so new bookings to inactive accounts still trigger the activation dialog.

**Compliance — orphan supplier invoice on JE failure** (`supplier-invoices/route.ts`, `supplier-invoices/[id]/credit/route.ts`)
On `AccountsNotInChartError` we now delete the just-inserted supplier invoice row (items cascade-delete) before returning the structured 400. Previously the invoice survived with no JE, understating ingående moms on the momsdeklaration.

### Pushing back

**Compliance — `correctEntry` leaves an orphan reversal**
Re-read `lib/core/bookkeeping/storno-service.ts:246-251`. The existing `try/catch` around the correction step already calls `cancelEntry(reversalEntry.id)` on any failure, so the state stays consistent: original is `posted`, reversal is `cancelled`, no orphan. The typed error propagates cleanly through that catch.

**Compliance — `mark-paid` routes create paid-but-no-JE state**
In both `app/api/invoices/[id]/mark-paid/route.ts` and `app/api/transactions/[id]/match-invoice/route.ts`, JE creation runs *before* the invoice status update, and my `AccountsNotInChartError` catch returns early. The invoice never gets marked paid if the JE fails. Comment at line 100 of `mark-paid` makes this invariant explicit: *"Create journal entry FIRST — only mark paid if accounting succeeds"*.

**Compliance — voucher gap risk**
Reviewer themselves noted "in practice no gap occurs today". `AccountsNotInChartError` is thrown by `createDraftEntry` before the `commit_journal_entry` RPC is called, so the voucher counter is never incremented on this path. Fragility concern is valid long-term but not a merge blocker.

### Deferred to follow-ups

- **Period-lock awareness on activation** — flag/warn when a user activates an account during a locked period (BFNAR 2013:2 kap 8). Additive.
- **Explicit audit-log line for reactivation** — `chart_of_accounts` is covered by the generic audit trigger already, but a dedicated event type would help downstream readers. Additive.
- **Wiring the activation dialog into supplier-invoice, invoice-send, and transaction-match surfaces** — called out in the PR description as out-of-scope. The structured 400 + Swedish fallback toast ship in this PR; the richer dialog follows.

### Tests

2285 / 2285 passing. Test mocks for `chart_of_accounts` updated for the new query chain (`eq.in.eq` after the conditional `is_active`).

jakobwennberg:
Amended `fcb6db7` addressing the second compliance review pass.

### Applied

**`commitSendInvoice` / `commitMarkInvoiceSent` non-blocking + visible** (`app/api/pending-operations/[id]/commit/route.ts`)
The two catches stay non-blocking (reverting them was the right call for Greptile P1 #2), but they now append an `InvoiceJournalEntrySkipped` event to `processing_history` on every silent skip. The event carries `invoice_id`, `operation`, `reason_code`, and — when it's an activation failure — the missing account numbers. Operators / bokföring consultants can query `processing_history` for `InvoiceJournalEntrySkipped` and re-book the missing verifikation before the momsdeklaration is filed. The breadcrumb is best-effort (append errors are logged-and-swallowed) so it never blocks invoice delivery.

**Supplier invoice rollback extended to all JE failures** (`supplier-invoices/route.ts`, `supplier-invoices/[id]/credit/route.ts`)
Previous commit only rolled back on `AccountsNotInChartError`. The catch now rolls back on *any* JE failure — period lock, balance trigger, DB error — and returns a clear Swedish error instead of logging silently. Keeps supplier invoices and their registration JE atomic from the user's perspective: either both land, or neither does. Closes the underreporting gap on 2440 / 2641 for any JE-creation failure mode.

### Pushing back on

**`correctEntry` reversal sub-step needs `includeInactive: true`**
The reversal sub-step in `storno-service.ts:111-125` doesn't call `resolveAccountIds` at all — it copies `account_id` directly from the original lines. The `.eq('is_active', true)` filter at line 171 applies only to the *new corrected lines* (step 2 of the flow), which is the right scope — the user is submitting new line data there, and the activation dialog should fire via the already-wired `correct` route. The reviewer's recommended fix ("reversal sub-step: includeInactive=true, corrected lines: active-only") is already what the code does.

### Deferred to follow-ups

- Period-lock awareness on the activate endpoint (BFNAR 2013:2 kap 8)
- Wiring `ActivateAccountsDialog` into the pending-op flows for `send_invoice` / `mark_invoice_sent` so they can be made blocking again with a one-click recovery path

### Tests

2285 / 2285 passing.

PR #391, opened 2026-05-05
feat(skatteverket): rewrite AGI flow against real Skatteverket RAML

jakobwennberg:
## Summary

- The previous AGI client posted JSON to URL paths that don't exist on Skatteverket's gateway and used invented field names. Rewritten against the official RAMLs in `dev_docs/`: `arbetsgivardeklaration-inlamning(1.7.7)` and `arbetsgivardeklaration-hantera-redovisningsperiod(1.2.8)`.
- Real flow: `POST /underlag` (XML) → `GET /underlag/{id}/kontrollresultat` → `POST /underlag/{id}/spara` → `POST /arbetsgivare/{x}/redovisningsperioder/{y}/skapaGranskningsunderlag` (returns Mina Sidor BankID deep-link) → `GET .../kvittenser` (after signing). Plus `POST .../las` / `lasUpp` on the hantera API.
- Drops `agi-mappers.ts`; `lib/salary/agi/xml-generator.ts` already produces schema-valid XML. The extension now feeds `agi_declarations.xml_content` straight to `POST /underlag`.
- Adds the **`agd`** OAuth scope, confirmed verbatim from SKV's *Tjänstebeskrivning Arbetsgivardeklaration inlämning v1.7*, §4.1.2.2 — the 403 example reads `"description": "The required scope agd has been requested for that access token."`
- Adds a Skatteverket connect step to `NewUserChecklist` alongside SIE/old-system import and bank, marked optional. New `hasSkatteverketConnected` flag on `OnboardingProgress`.

## What changed

**Extension (`extensions/general/skatteverket/`)**
- `lib/agi-client.ts` — full rewrite, 9 functions matching real endpoints
- `types.ts` — replaced fictional AGI types with real schemas (kontrollresultat, granskningsunderlag, kvittenser, error envelope)
- `lib/api-client.ts` — `skvRequest` accepts a `contentType` option (AGI needs `application/xml`)
- `lib/oauth.ts` — `agd` appended to `DEFAULT_SCOPES`
- `index.ts` — replaced 8 fake AGI handlers with 9 routes mirroring the real flow (`/agi/submit`, `/agi/kontrollresultat`, `/agi/spara`, `/agi/underlag` DELETE, `/agi/sparad` DELETE, `/agi/granskningsunderlag`, `/agi/kvittenser`, `/agi/las`, `/agi/lasUpp`); kept local `/agi/status`
- `manifest.json` — new optional base-URL env vars
- `lib/agi-mappers.ts` + its test — **deleted**

**UI**
- `components/salary/AGIPanel.tsx` — new state machine (`underlag_submitted` → `awaiting_signing` → `signed`); kontrollresultat polling with cap; normalized `KontrollFinding` rendering
- `components/onboarding/NewUserChecklist.tsx` — Step 3 "Anslut Skatteverket" (extension-gated, optional); shows green completed state when connected
- `components/dashboard/DashboardContent.tsx` — passes `hasSkatteverketConnected` through
- `components/settings/SkatteverketConnectPanel.tsx` — added `agd` → "Arbetsgivardeklaration" label

**Server**
- `app/api/salary/runs/[id]/agi/submit/route.ts` — orchestrator points at `/agi/submit`; comment updated
- `app/api/salary/runs/[id]/agi/submit/__tests__/route.test.ts` — updated expectation
- `app/(dashboard)/page.tsx` — counts skatteverket_tokens for the active user
- `types/index.ts` — `OnboardingProgress.hasSkatteverketConnected: boolean`

## Caveats

- **Existing connections lack the `agd` scope.** Anyone connected before this PR will get `403 invalid_scope` on AGI calls and must reconnect via `/settings/skatteverket`. The connect panel's missing-scope warning today only flags `skattekonto`; extending it to also flag `agd` is a small follow-up.
- **Defaults still point at `api.test.skatteverket.se`.** Production needs the env vars set: `SKATTEVERKET_API_BASE_URL`, `SKATTEVERKET_AGD_INLAMNING_API_BASE_URL`, `SKATTEVERKET_AGD_PERIOD_API_BASE_URL`, `SKATTEVERKET_SKATTEKONTO_API_BASE_URL`, `SKATTEVERKET_OAUTH_BASE_URL`.
- **No dedicated unit tests on the new `agi-client.ts`.** The orchestrator integration test covers the happy path indirectly. Worth a follow-up.
- `<a>` (not `<Link>`) on the onboarding connect button is intentional — the authorize endpoint 302-redirects cross-origin to `peroauth2.skatteverket.se`.

## Test plan

- [ ] `npx tsc --noEmit` clean for changed files (verified locally — pre-existing failures in unrelated test fixtures only)
- [ ] `npx vitest run extensions/general/skatteverket app/api/salary/runs/[id]/agi` — all 47 tests pass (verified)
- [ ] `npx eslint` — zero errors on touched files (verified; remaining warnings pre-existing)
- [ ] Manual smoke: connect via BankID in test env, generate AGI from a salary run, click "Skicka in underlag" → verify `inlamningId` returned, `kontrollresultat` polled to `DONE_SUCCESS`, then "Skapa signeringslänk" returns Mina Sidor URL
- [ ] Manual smoke: dashboard onboarding gate shows the new Skatteverket step, click → BankID round-trip → returns to dashboard with step 3 marked completed
- [ ] Reconnect an existing connection and verify `agd` shows up among the granted scopes in `/settings/skatteverket`


jakobwennberg:
Pushed `39350f1` addressing the review.

**Greptile P1 findings**

- ✅ **#2 INCORRECT_DATA link unreachable** — fixed. New destructive-styled block in `AGIPanel` renders the felrapport URL when `submission.status === 'underlag_rejected'`, distinct from the happy-path signing block.
- ✅ **#3 `/agi/underlag` DELETE doesn't clear state** — fixed. Handler now accepts an optional `period` query param and clears `agi_submission_{period}`. Falls back to scanning recent submission keys for a matching `inlamningId` when caller didn't pass period.
- ❌ **#1 Wrong base URL for `agiTaBortSparadInlamning`** — false positive. Cross-checked both RAMLs: `taBortSparadInlamning` is defined in `arbetsgivardeklaration-inlamning(1.7.7)`, not in `hanteraredovisningsperiod`. The bot misread my own JSDoc grouping. I tightened the comment block so future reviewers don't make the same misread.

**Swedish compliance bot findings**

- ✅ **AGI deadline guard removed** — fixed. Re-added the `['review', 'approved', 'paid', 'booked']` salary-run status check inside `loadAGIXml` so the direct-call path enforces it too.
- ✅ **`agi_declarations.status='exported'` premature** — fixed. Moved from `/agi/submit` to `/agi/spara`. Status now flips to `exported` only after the underlag is durably saved into Eget utrymme. `/agi/spara` accepts `salaryRunId` in the body for the happy path and scans `agi_submission_*` state as a fallback.
- ✅ **`salary_runs.agi_submitted_at` premature** — fixed. Removed from the orchestrator. The `/agi/kvittenser` handler now stamps it to `kvittens.signeradTid` (mirroring SKV's own timestamp) when it pins the receipt to the matching `agi_declarations` row.

**Acknowledged but out of scope for this PR**

- Correction (rättelse) detection — this depends on `xml-generator.ts`'s `_isCorrection` parameter, which is currently unused (the underscore tells the truth). Pre-existing bug; will file a separate issue.
- Hela-kronor rounding test coverage — `xml-generator.ts` does its own rounding via `formatAmount`. Worth a smoke test on `xml_content` containing no decimals; will file separately.
- Conditional `agd` scope gating — would require a probe before the OAuth redirect to know whether the company has employees. Will evaluate as a separate UX iteration.

jakobwennberg:
Pushed `4a3e07d` addressing the Swedish compliance review.

**Resolved**

- ✅ **`status='exported'` semantically misleading mid-flow** — added migration `20260505100000_agi_declarations_pending_signature.sql` extending the CHECK constraint with `pending_signature`. `/agi/spara` now sets `pending_signature` (underlag in Eget utrymme, not yet filed); `/agi/kvittenser` promotes to `submitted` when a `uuidKvittens` is observed. `exported` is preserved for the legacy manual XML download path.

- ✅ **`agi_submitted_at` only set on user-triggered poll** — `AGIPanel` now schedules background `/agi/kvittenser` polls at 30s, 2 min and 5 min after the granskningsunderlag link is generated. Cancels remaining polls once a kvittens is observed. Cleans up timers on unmount via `useRef` + `useEffect`. The "Hämta kvittens" button still works as an explicit recovery path.

- ✅ **`agd` scope mismatch surfaces as generic 403** — new `MISSING_SCOPE` code on `SkatteverketAuthError`. `skvRequest` now detects `invalid_scope` in the 403 body (per SKV's Tjänstebeskrivning v1.7 §4.1.2.2) and throws with a Swedish message directing the user to disconnect + reconnect via Settings. Mapped to HTTP 401 in `handleSkvError` so the frontend treats it as a token-level problem.

- ✅ **Deadline text** — softened slightly. The canonical reference (`.claude/skills/swedish-payroll/references/agi-filing.md:13-17`) confirms the standard 12th-of-next-month deadline applies regardless of company size; the 17th is only the January/August carve-out for employers with turnover ≤ 40 MSEK. Updated copy: `"Deadline: 12:e i månaden efter utbetalning (17:e i januari/augusti för arbetsgivare med omsättning ≤ 40 MSEK)."`

**Acknowledged but out of scope**

- **`submitted_by` vs `signed_by_personnummer`** — splitting these requires an `agi_declarations` schema add. Will file as a separate ticket; the current `submitted_by = ctx.userId` plus the `signeradAv` mirror in the local submission state covers the audit need for single-operator companies (the dominant case).
- **FK570 specifikationsnummer test coverage** — already covered in `lib/salary/__tests__/agi-xml.test.ts:194` (`'preserves Specifikationsnummer FK570 per employee'`) plus the `BetalningsmottagarId FK215` and `KontantErsattningUlagAG FK011` tests next to it. The deleted `agi-mappers.test.ts` only asserted personnummer-decryption error handling, which moved to the XML generator. No regression.
- **Stale XML if payroll amended after generation** — pre-existing concern with the XML-generation route, separate from this PR. Will file separately.

Tests: 47/47 pass. Lint and typecheck clean for touched files.

jakobwennberg:
Pushed `f9d5582` addressing the latest compliance review.

**Resolved**

- ✅ **Background polls don't satisfy the audit trail (BFNAR 2013:2 / BFL 5 kap 5§)** — added `/api/extensions/skatteverket/agi/kvittenser/cron`, scheduled every 2h in `vercel.json`. It walks every `agi_declarations` row in `pending_signature` status, fetches `kvittenser` via the company's stored token, and on a hit promotes to `submitted` + stamps `salary_runs.agi_submitted_at` from `kvittens.signeradTid`. The client-side timers from round 1 stay as the fast-path UX, but the cron is now the authoritative source for the audit trail.

- ✅ **`AGIStatus` type out of sync with the migration** — extended the union in `types/index.ts:2475` to include `'pending_signature'` with documenting comments per state.

- ✅ **Stale `'exported'` comment in `AGIPanel.tsx`** — updated to reference `'pending_signature'`.

**Audited and dismissed (with rationale)**

- **Downstream consumers of `status='exported'`** — grep across the repo: zero code paths read this string as a "filed" signal. Only references are the migration, the type union, and one (now-fixed) stale comment. The legacy meaning is preserved in the new union for forward compatibility, but no consumer currently relies on it.

- **Youth-rate Ruta 062 / `avgifterUnderlagUngdom` coverage** — the deleted test was checking a JSON field I'd invented; the actual SKV `huvuduppgift.json` schema has no per-bracket basis fields (`summaArbAvgSlf` is the only avgifter total). SKV derives the rate per IU from the `betalningsmottagarId` personnummer. `lib/salary/agi/xml-generator.ts` correctly omits the fictional field. No regression.

- **Correction (rättelse) detection** — for AGI, "rättelse" is expressed by reusing the same `Specifikationsnummer` (FK570) per employee on re-submission, which `xml-generator.ts:317` does (it pulls `specification_number` from the stable employee row). There's no top-level rättelse boolean in the AGI XML schema; the `_isCorrection` parameter on `generateAGIXml` is genuinely unused (the underscore signals it). The only unhandled rättelse case is removing an employee from a re-submitted period, which would need `Borttag` (FK205); filing as a separate ticket.

- **Deadline-text per-employer** — keeping the current `(17:e i januari/augusti för arbetsgivare med omsättning ≤ 40 MSEK)` parenthetical. Personalising to company turnover would require pulling that figure on render and is borderline product-decision; the current copy clearly scopes the exception to the qualifying group.

**Filed as separate follow-up tickets**

- `submitted_by` vs `signed_by_personnummer` column split (schema add).
- `Borttag` (FK205) emission for employee removal in rättelse re-submissions.
- Stale-XML guard if payroll is amended after `xml_content` generation.

Tests: 47/47 pass. Lint and typecheck clean.

jakobwennberg:
Pushed `8586e0d` addressing the round-3 review.

**Resolved**

- ✅ **`submitted_by` left NULL on cron-reconciled rows** — cron now writes `submitted_by: token.user_id`. The token row was created when the operator authenticated with BankID, and `kvittens.signeradAv` refers to the same person, so this is the correct natural-person mapping for the audit trail. Closes the BFL 5 kap 6§ / BFNAR 2013:2 kap 8 gap.

- ✅ **`/agi/spara` could regress an already-promoted row** — added `.in('status', ['generated', 'exported'])` predicate so a delayed `/agi/spara` after the cron has already flipped to `'submitted'` is a no-op. behandlingshistorik moves forward only.

- ✅ **DONE_REJECTED / DONE_FAILED never updated `agi_declarations.status`** — `/agi/kontrollresultat` handler now flips the row to `'rejected'` on terminal failure, using the same cached-submission-state lookup `/agi/spara` already uses. Same monotonicity guard prevents regressing a successfully-filed row.

- ✅ **Deadline criterion: omsättning → lönesumma** — SFL 26 kap's actual criterion is total taxable wages, not turnover. Updated `AGIPanel.tsx` pendingText. Note: the internal reference at `.claude/skills/swedish-payroll/references/agi-filing.md:16-17` uses colloquial "turnover" and is also wrong by the same standard — could fix in a follow-up but out of scope here.

**Disagreed**

- ❌ **`agiSkapaGranskningsunderlag` "wrong base URL"** — false positive. Both RAMLs expose `skapaGranskningsunderlag`. The inlamning extern.raml explicitly lists it in the operations table at line 31 and routes it at line 59. The hantera RAML lists it too (line 30). Either base URL works; using the inlamning base keeps the spara → granskning → kvittenser sequence on a single API surface, which is what the inlamning RAML's own "Filing flow" documentation describes.

Tests: 47/47 pass. Lint and typecheck clean.

jakobwennberg:
Pushed `6b05a33` for the round-4 review.

**Resolved**

- ✅ **`agi_submitted_at` falsified when `signeradTid` absent** — both `/agi/kvittenser` handler and the kvittens cron now leave the column NULL when SKV's kvittens omits `signeradTid`, with a `console.warn`. The status flip to `'submitted'` still happens (the audit gap was the timestamp only). Closes the BFNAR 2013:2 kap 8 / BFL 5 kap 6§ honesty gap on filing-moment recording.

- ✅ **No proactive UI for missing `agd` scope** — `SkatteverketConnectPanel` now warns when stored scopes lack `agd`, mirroring the existing `skattekonto`-missing pattern, and adds the same condition to the "Anslut igen" CTA. `AGIPanel` surfaces a banner in the connected state with a link to `/settings/skatteverket`. Users with pre-`agd` tokens get nudged to reconnect before submission, not at it.

- ✅ **Granskningsunderlag `isError` mixed HTTP 409 with payload string** — refactored to key solely on `tillstand`. Only `LOCKED_FOR_SIGNING` / `UNLOCKED` route to `awaiting_signing`; everything else (`INCORRECT_DATA`, `RECEIVING`, `CALCULATING`, `SIGNING`) routes to `underlag_rejected`. A future SKV addition like `RECEIVING` returned at HTTP 200 won't slip through as signable.

**Repeat finding (already addressed in round 3)**

- **Deadline text conditional rendering** — bot continues to flag the `pendingText` parenthetical as risk for above-threshold employers. The criterion was already corrected to `lönesumma` (per SFL 26 kap) in round 3, and the parenthetical scopes the exception explicitly to "≤ 40 MSEK". Personalising per-company would require pulling annual taxable wages on render, which is a product decision rather than compliance — leaving as-is.

Tests: 47/47 pass. Lint and typecheck clean.

jakobwennberg:
Pushed `a70e038` for round 5.

**Resolved**

- ✅ **`agi_submitted_at` shouldn't be NULL when filing succeeded** — round 4 had us leave it NULL on missing `signeradTid` to avoid falsifying the signing moment; round 5 (correctly) points out that NULL hides the filing entirely. Resolution: presence of `uuidKvittens` proves SKV signed and accepted the AGI, so we stamp `signeradTid || now()` and warn-log when the fallback is used. Applies to both the `/agi/kvittenser` handler and the kvittens cron.

- ✅ **`signeradAv` (BankID signer's personnummer) wasn't preserved** — added the full kvittens (uuidKvittens, signeradAv, signeradTid, arbetsgivare, period, underlag) to `agi_declarations.response_data`. The column already exists (no migration). `submitted_by` remains the auth.users UUID we have on hand — useful for app-level traceability — and `response_data.signeradAv` carries the legally load-bearing identity for BFL 5 kap 6§ / BFNAR 2013:2 kap 8. Cron rows are tagged `reconciledBy: 'cron'` so the audit trail distinguishes path of arrival.

- ✅ **`/agi/spara` couldn't recover from `'rejected'`** — added `'rejected'` to the monotonicity guard. The xml-route reuses the same `agi_declarations` row when re-generating XML, so this is the realistic re-submission recovery path. `'submitted'`/`'accepted'` are still blocked.

- ✅ **Misleading "fix in Mina Sidor" comment** — Mina Sidor doesn't expose in-place editing of saved underlag. Updated `agiSparaUnderlag` JSDoc to describe the actual rättelse flow (regenerate XML with same FK570 specifikationsnummer per employee → resubmit).

- ✅ **Deadline copy** — rephrased to `"vars sammanlagda lönesumma understiger 40 MSEK"` per the reviewer's suggested wording. Same semantics, cleaner Swedish.

**Disagreed (note)**

- The deadline-text "logical inversion" claim was based on a misread — `≤ 40 MSEK` clearly excludes large employers, so the prior text wasn't actually inverted. Adopted the rephrasing anyway because it reads better.

Tests: 47/47 pass. Lint and typecheck clean.

cc: this branch is now 7 commits deep on round-after-round compliance review feedback. The remaining bot findings are converging — most rounds are now repeats with finer phrasing, plus the bot's two self-reversals on `agi_submitted_at = NULL` vs reconciliation-time. Happy to continue iterating, but flagging that we may have hit the point of diminishing returns; calling for a human review pass would help confirm what's truly blocking vs cosmetic.

jakobwennberg:
Pushed `bd7293e` for round 6.

**Resolved**

- ✅ **`/agi/spara` shouldn't accept `'exported'` rows** — verified by grep that no code path currently writes `status='exported'`; the value lives in the schema for the legacy manual-download path that no longer has a writer. Tightened the monotonicity guard to `.in(['generated', 'rejected'])`. Recovery from rejection still works; conflation with the dormant `'exported'` state is now closed (BFL 5 kap 6§ chain-of-custody concern).

- ✅ **Deadline copy "per år" qualifier** — added so the 40 MSEK threshold can't be misread as per-payment.

**Acknowledged but not actionable in this PR**

- **`agi_submitted_at` could remain NULL if both cron and interactive handler fail** — flagged for follow-up. Reasonable mitigations include a low-priority alert when `pending_signature` rows age past N days, or a fallback write on next user login. Out of scope here.

- **`submitted_by` records token-owner; downstream UI must surface `signeradAv`** — this is forward-looking guidance for code that doesn't yet exist (audit-trail UI / exports). The data is preserved in `response_data.signeradAv`; future audit-trail UIs need to read both fields. Will note in the follow-up tickets.

- **Repeat findings: växa-stöd / youth-rate handling, personnummer decryption regression** — both addressed in earlier rounds. `xml-generator.ts:290` calls `decryptPersonnummer` and throws on failure; the agi-xml test suite covers the FK570 contract; the per-rate breakdown the bot worried about doesn't exist as a JSON field in SKV's actual huvuduppgift schema (rates are derived per IU from `betalningsmottagarId` personnummer).

Tests: 47/47 pass. Lint and typecheck clean.

This is the 8th commit on this branch in response to bot feedback. Compliance findings are now in the cosmetic / forward-looking territory; flagging again that a human review pass would be useful before further iteration.

PR #506, opened 2026-05-16
fix(docker): self-hosted build-from-source startup failures

alfjesus:
These three issues surfaced when running

      docker compose -f docker-compose.yml -f docker-compose.build.yml up --build

  i.e. building the image locally rather than pulling
  ghcr.io/erp-mafia/gnubok:latest. Pulling the pre-built image masked
  them; rebuilding from source exposed each one in turn.

  - Dockerfile: COPY of /app/public was missing --chown=nextjs:nodejs. The nextjs user could not sed-substitute the branding placeholder in public/sw.js, so the entrypoint exited with "Permission denied" before the server ever started. Introduced when sw.js placeholder substitution was added; the existing pre-built image predates it.

  - docker-entrypoint.sh: placeholder substitution only ran over /app/.next/static (client bundles). Next.js inlines NEXT_PUBLIC_* values into the server bundle too, so server code received the literal "__NEXT_PUBLIC_SUPABASE_URL__" string at runtime and /api/health returned 503 with "Invalid supabaseUrl: Must be a valid HTTP or HTTPS URL." Broadened the find to /app/.next so server chunks under /app/.next/server are patched as well.

  - docker-compose.yml: the cron service is always built locally (docker/cron.Dockerfile), so this hit even when the app image was pulled. Supercronic's built-in process reaper crashes with "Failed to fork exec: no such file or directory" when running as PID 1 on alpine, putting the container into a restart loop. Set init: true so Docker injects tini as PID 1 and supercronic detects it is not PID 1 and skips its broken reaper path.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -51,7 +51,7 @@ RUN addgroup --system --gid 1001 nodejs && \
     adduser --system --uid 1001 nextjs
 
 # Copy standalone output
-COPY --from=builder /app/public ./public
+COPY --from=builder --chown=nextjs:nodejs /app/public ./public
 COPY --from=builder --chown=nextjs:nodejs /app/.next/standalone ./
 COPY --from=builder --chown=nextjs:nodejs /app/.next/static ./.next/static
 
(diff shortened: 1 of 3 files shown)


alfjesus:
Most of these changes was commited by #524 . Closing this

PR #757, opened 2026-06-19
Realtime updates for dashboard transactions

esaiaswestberg:
# Summary
Keep the dashboard transaction badge and the `/transactions` page in sync with the database while the user stays on the site.

This PR adds a shared browser Supabase hook and uses it to subscribe to `postgres_changes` on `public.transactions`, scoped to the active company. When transactions are inserted, updated, or deleted, the dashboard nav refreshes the uncategorized count and the transactions page refreshes its list and total uncategorized count without requiring a full page reload.

# What changed
- Added `useRealtimeSupabase` in `lib/hooks/use-realtime-supabase.ts` so realtime consumers share the same stable browser client setup.
- Updated `components/dashboard/DashboardNav.tsx` to listen for `transactions` changes for the current company and refresh the uncategorized badge live.
- Updated `app/(dashboard)/transactions/page.tsx` to subscribe to the same table and refetch the visible transaction list plus uncategorized count when a relevant realtime event arrives.
- Kept the transactions page refresh path scoped to transaction data only, so realtime updates do not re-fetch unrelated Skatteverket rows on every change.
- Added a Supabase migration to publish `public.transactions` to `supabase_realtime`.

# Why
The uncategorized badge already seeded correctly on first render, but it became stale during a visit whenever transactions changed elsewhere. The same problem affected the transactions page itself. This fix makes both surfaces react to the database instead of relying on a manual refresh.

# Implementation notes
- Realtime is narrowed on the client with `company_id=eq.<companyId>` so the browser only subscribes to the active company from `CompanyContext`.
- The nav still uses the server-provided count as the initial value, then updates from realtime.
- The transactions page still does its initial server-driven load, then switches to realtime refetches after mount.
- The subscription uses `postgres_changes`, not polling or time-based refreshes.

# Access control
This change does not introduce a new path to read another company's transactions.

- The actual authorization boundary stays in Postgres RLS on `public.transactions`. The current multi-tenant policy is `transactions_select` with `company_id IN (SELECT public.user_company_ids())`.
- Supabase Realtime only delivers row change events that the authenticated session is allowed to read through that RLS policy.
- The `company_id=eq.<companyId>` clause in the realtime subscription is a client-side scoping filter only. It narrows the stream to the active company after the server has already applied RLS.
- The active company id itself is resolved server-side in [`app/(dashboard)/layout.tsx`](/home/esaiaswestberg/Documents/repositories/accounted/app/(dashboard)/layout.tsx) via `getActiveCompanyId(...)`, then passed into the client context and reused by the dashboard nav and transactions page.
- In practice, a malicious user can change the client-side filter in their own browser, but that does not bypass the database. They still only receive rows from companies they are entitled to access under RLS.

# Validation
- Ran `npm run lint -- components/dashboard/DashboardNav.tsx app/(dashboard)/transactions/page.tsx lib/hooks/use-realtime-supabase.ts`
  - Passed with 3 pre-existing warnings in `app/(dashboard)/transactions/page.tsx`.
- Ran `npm run check:guards`
  - Passed.
- Ran `npm run check:lint`
  - The wrapper failed with `eslint produced no JSON output`.
- Ran `npm run test`
  - Passed: `460` files, `5993` tests.
- Ran `npm run test:pg`
  - Failed immediately because no Vitest project matched `pg-real` in this checkout.
- Verified the new Supabase publication migration is idempotent.

# Notes
- The migration must be applied on the hosted Supabase project for the realtime events to flow.
- This change is intentionally narrow: it covers the dashboard nav badge and the `/transactions` surface only.

### DIFF
--- a/app/(dashboard)/transactions/page.tsx
+++ b/app/(dashboard)/transactions/page.tsx
@@ -1,10 +1,9 @@
 'use client'
 
-import { useState, useEffect, useMemo, useRef } from 'react'
+import { useState, useEffect, useMemo, useRef, useCallback } from 'react'
 import { AnimatePresence } from 'framer-motion'
 import { useSearchParams } from 'next/navigation'
 import { useTranslations } from 'next-intl'
-import { createClient } from '@/lib/supabase/client'
 import { Button } from '@/components/ui/button'
 import { Badge } from '@/components/ui/badge'
 import { Dialog, DialogContent, DialogHeader, DialogTitle } from '@/components/ui/dialog'
@@ -56,6 +55,7 @@ import type {
 } from '@/types/skatteverket'
 import { findBankSkvCounterparts } from '@/lib/skatteverket/bank-counterpart'
 import { useCompany } from '@/contexts/CompanyContext'
+import { useRealtimeSupabase } from '@/lib/hooks/use-realtime-supabase'
 import { getErrorMessage } from '@/lib/errors/get-error-message'
 import { formatCurrency, formatDate } from '@/lib/utils'
 import type { TransactionCategory, CreateTransactionInput, Invoice, Customer, SupplierInvoice, Supplier, VatTreatment, EntityType, LinePatternEntry, BookingTemplateLibrary } from '@/types'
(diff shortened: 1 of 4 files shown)


esaiaswestberg:
Sorry for missing this, has had a lot going on 😅

Fun to see that the implementation was sound. Just glad to be contributing.

PR #767, opened 2026-06-22
feat(api): v1 endpoints to stamp invoice inbox items as consumed

flojon:
## Summary

- Extended `POST /api/v1/companies/{companyId}/documents/{id}/link` with an optional `inbox_item_id` field — when supplied, stamps `invoice_inbox_items.created_journal_entry_id` after the document link succeeds (best-effort; failure is logged but does not fail the request)
- Added new `POST /api/v1/companies/{companyId}/inbox-items/{id}/stamp` for stamping an inbox item independently of the document link — useful when the link was already made without `inbox_item_id`
- Both use `documents:write` scope and require `Idempotency-Key`
- Stamp endpoint is strictly idempotent: same JE → 200, different JE → CONFLICT

## Test plan

- [ ] `POST .../documents/{id}/link` with `inbox_item_id` → document linked + inbox item disappears from todo
- [ ] Same call repeated → 200, no duplicate write (`.is('created_journal_entry_id', null)` guard)
- [ ] `POST .../documents/{id}/link` without `inbox_item_id` → existing behavior unchanged
- [ ] `POST .../inbox-items/{id}/stamp` → stamps item, returns `{ id, created_journal_entry_id }`
- [ ] Stamp with same JE → 200 (idempotent)
- [ ] Stamp with different JE when already stamped → CONFLICT
- [ ] Unknown inbox item or JE → NOT_FOUND
- [ ] Missing `Idempotency-Key` → 400

### DIFF
--- a/app/api/v1/companies/[companyId]/documents/[id]/link/route.ts
+++ b/app/api/v1/companies/[companyId]/documents/[id]/link/route.ts
@@ -4,7 +4,12 @@
  * Link an already-uploaded document to a journal entry (and optionally a
  * specific line). Wraps lib/core/documents/document-service.linkToJournalEntry.
  *
- * Body: `{ journal_entry_id: UUID, journal_entry_line_id?: UUID }`.
+ * Body: `{ journal_entry_id: UUID, journal_entry_line_id?: UUID, inbox_item_id?: UUID }`.
+ *
+ * When `inbox_item_id` is supplied the matching invoice_inbox_items row is
+ * stamped with `created_journal_entry_id` so it drops out of the active inbox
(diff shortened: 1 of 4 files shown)


flojon:
#769 dropped these changes so they can merge separately. Will try to fix the v1 contract guards.

flojon:
Both fixed in 8d00822:

1. **Envelope contract** — wrapped with `dataEnvelope(InboxItemStampedResponse)` and imported from registry.
2. **Spec snapshot** — also found that the stamp route was missing from `load-routes.ts` entirely (so the snapshot test was passing by not seeing the endpoint at all). Added the import and refreshed the snapshot to 102.

PR #769, opened 2026-06-24
fix(transactions): resolve bank account from cash_account_id in booking dialog

flojon:
## Summary

- Adds `cash_account_id` to the transactions list API response so booking dialogs can look up the correct ledger account per transaction
- Extracts `resolveAccount` (previously private to `MatchVoucherDialog`) to `lib/cash-accounts/resolve-account.ts` with 6 unit tests, so both booking dialogs share the same resolution logic
- `TransactionBookingDialog` now fetches `/api/cash-accounts` on open and pre-fills the bank leg with the transaction's actual ledger account (e.g. 1940 for a savings account) instead of the hardcoded `'1930'`
- Both the blank booking flow and the template-based flow are fixed
- Form is held until the fetch resolves so user input is never discarded by a remount
- Stale fetch races fixed in both `TransactionBookingDialog` and `MatchVoucherDialog` via cancellation flags

## Out of scope (follow-up)

- `BulkBookDialog` and `BookDirectlyDialog` still hardcode `'1930'` — pre-existing, same root cause, separate PR

## Test plan

- [ ] Open booking dialog for a transaction on a non-1930 account (e.g. savings mapped to 1940) — bank leg shows 1940, not 1930
- [ ] Open booking dialog for a transaction on the primary 1930 account — still shows 1930
- [ ] Open booking dialog with a preselected template on a non-1930 account — settlement line shows the correct account
- [ ] Switch quickly between transactions in different accounts — no stale account from a previous dialog open
- [ ] `npx vitest run lib/__tests__/resolve-account.test.ts` — 6/6 pass
- [ ] `npx vitest run app/api/transactions/__tests__/route.test.ts` — all pass

### DIFF
--- a/app/api/transactions/route.ts
+++ b/app/api/transactions/route.ts
@@ -64,7 +64,7 @@ export async function GET(request: Request) {
 
   let query = supabase
     .from('transactions')
-    .select('id, date, description, amount, currency, amount_sek, exchange_rate, reference, journal_entry_id, reconciliation_method, is_ignored')
+    .select('id, date, description, amount, currency, amount_sek, exchange_rate, reference, journal_entry_id, reconciliation_method, is_ignored, cash_account_id')
     .eq('company_id', companyId)
 
   // unmatched and reconciled are mutually exclusive — unmatched wins if both set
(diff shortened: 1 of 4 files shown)


flojon:
Sorry claude messes up the branches sometimes. I have cleaned up this PR so it only contains the intended changes. (So dropped inbox stamp endoints and overdue CAS guard)

flojon:
Working on it :)

PR #810, opened 2026-06-28
fix(sie): add ?encoding=cp437 for legacy bookkeeping software

flojon:
## Summary

- Removes the spurious `#FORMAT PC8` tag from the default UTF-8 SIE export — declaring CP437 while serving UTF-8 caused mojibake when importing into software that respected the header.
- Adds `?encoding=cp437` to both `/api/reports/sie-export` and `/api/v1/.../reports/sie-export`. When set, the response is properly CP437-encoded bytes with `#FORMAT PC8` in the header, as required by legacy desktop software (Visma Administration, BL Administration, etc.).
- Default behaviour (UTF-8, no format tag) is unchanged for modern cloud-to-cloud use.

## Test plan

- [ ] `npx vitest run lib/reports/__tests__/sie-export.test.ts` — all 14 pass
- [ ] Default export: UTF-8, no `#FORMAT PC8`, Swedish chars correct in a text editor
- [ ] `?encoding=cp437`: `#FORMAT PC8` present, å/ä/ö correct when imported into CP437-expecting software

### DIFF
--- a/app/api/reports/sie-export/route.ts
+++ b/app/api/reports/sie-export/route.ts
@@ -38,7 +38,7 @@ export const GET = withRouteContext(
         emit_format_pc8: useCP437,
       })
 
-      const body = useCP437 ? encodeSIEToCP437(sieContent) : sieContent
+      const body = useCP437 ? Buffer.from(encodeSIEToCP437(sieContent)) : sieContent
       const contentType = useCP437 ? 'application/octet-stream' : 'text/plain; charset=utf-8'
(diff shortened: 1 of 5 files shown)


flojon:
Done — rebased onto main (the `obEntryId` / periodEntries hunks dropped cleanly since git saw they were already in main). Wrapped the `Uint8Array` in `Buffer.from()` in both routes; 16 SIE tests still pass.

PR #813, opened 2026-06-28
feat(arcim): import Bokio underlag and link to verifikat (#786)

jhagberg:
## Summary

Implements the Bokio slice of #786: migrate the receipts/underlag attached to each verifikat. Neither the SIE GL import nor the arcim entity import carries them, so imported books end up with vouchers that have no underlying document. This adds an optional, re-runnable step that pages the Bokio `/uploads`, resolves each receipt's target verifikat, and archives it through the existing document service, linked to the journal entry.

Greenlit in #786; design follows @jakobwennberg's notes there.

## How the match works (validated on live data)

A Bokio upload only carries the journal entry's GUID, which we never imported. The SIE import *does* preserve Bokio's human voucher number on `journal_entries.source_voucher_series` / `source_voucher_number`. So the resolution is:

`upload.journalEntryId → Bokio journalEntryNumber ("V342") → gnubok verifikat where source_voucher_number matches`

Bokio restarts numbering at **V1 every fiscal year**, so the match is keyed on **(fiscal period, series, numeric part)**, not the raw string — the fiscal period is found from the Bokio entry's date. I confirmed the `#VER` equality the issue flagged against my own migrated 2021 books: of 484 vouchers, the 156 carrying receipts all resolve exactly, agreeing on date *and* amount, zero collisions.

## What's here

- **`lib/providers/bokio`** — `BokioClient.getBytes()` (binary download, same rate-limit + retry envelope as `get()`), and an `attachments` resource module: paginated uploads list, a GUID→voucher index built from `/journal-entries`, and per-upload download. `pageSize` is 100; the declared file type comes from the upload's `contentType` (the download is `application/octet-stream`).
- **`importProviderDocuments`** — bulk-reads the uploads, the Bokio voucher index, and gnubok's fiscal periods + verifikat + existing document hashes, then resolves each receipt **in memory** (no per-item N+1). Idempotent on `(company_id, sha256)` so re-runs don't duplicate — which matters because a receipt linked to a posted verifikat is undeletable (BFL 7 kap 2§ / WORM triggers). Best-effort: a per-receipt failure is counted and logged, never thrown. `dryRun` reports the match plan without downloading or writing.
- **`POST /import-documents`** — modelled on the existing `/reconcile` route, kept off the migration's critical path because the Bokio document API is rate-limited (200 req/60s) and a full sweep issues hundreds of download calls (would blow the 300s migration window).
- **`PROVIDER_IMPORT_DOCUMENTS_FAILED`** structured error.

## Design notes / decisions

- **Linking path** — uses `uploadDocument(..., { journal_entry_id })`, i.e. the same `lib/core/documents/document-service` module as #804's `linkToJournalEntry`, writing the same `document_attachments.journal_entry_id` column. Create-and-link is atomic this way (no transient unlinked row); it doesn't fork the link logic. Happy to switch to an explicit `linkToJournalEntry` two-step if you'd rather the two PRs converge on one call — say the word.
- **`journal_entry_id` link only** for v1 (per the issue) — most underlag sit behind plain expense verifikat on cash method, so the verifikat link covers them; the invoice path would be a redundant subset.
- **Idempotency** — content hash, not the `sha256_hash` unique index (which doesn't exist). There's no `metadata` column on `document_attachments`, so the Bokio upload id can't be stashed there; the trade-off is that a re-run re-downloads bytes to compute the hash before skipping. Noted as a v1 limitation — a future per-source marker column would let re-runs skip without downloading.
- **Unmigrated years** fall out naturally: a receipt whose Bokio date is outside every gnubok fiscal period resolves to nothing and lands in `unmatched`, so this is safe to run before all years are imported.
- v1 is **Bokio-only**; other providers are a no-op (not an error), so a mixed-provider caller can invoke it unconditionally.

## Test plan

- [ ] `extensions/general/arcim-migration/__tests__/import-documents.test.ts` (vitest): resolve+link with `upload_source=api`, sha256 idempotent skip, unmatched (wrong fiscal year), dryRun (no I/O), non-Bokio no-op, best-effort failure continuation.
- [ ] Could not run `vitest`/`tsc` locally (no `node_modules` in my checkout) — relying on CI `core-build`. `fetchAllRows` usage mirrors the existing no-cast pattern in `bulk-reconcile-supplier-vouchers.ts`.

### DIFF
--- a/extensions/general/arcim-migration/__tests__/import-documents.test.ts
+++ b/extensions/general/arcim-migration/__tests__/import-documents.test.ts
@@ -43,6 +43,7 @@ function rangeMockSupabase(byTable: Record<string, unknown[]>): SupabaseClient {
       select: () => node,
       eq: () => node,
       not: () => node,
+      order: () => node,
       range: () => Promise.resolve({ data: byTable[table] ?? [], error: null }),
     }
     return node
@@ -150,6 +151,28 @@ describe('importProviderDocuments', () => {
 
     expect(result).toMatchObject({ provider: 'fortnox', scanned: 0, linked: 0 })
     expect(mockFetchUploads).not.toHaveBeenCalled()
+  })
+
+  it('counts a receipt as unmatched when its journalEntryId is not in the Bokio voucher index', async () => {
+    // e.g. an unparseable journalEntryNumber — must be reported, not dropped.
+    const supabase = wireBokio()
+    mockFetchVoucherIndex.mockResolvedValue(new Map<string, BokioVoucherRef>())
+
+    const result = await importProviderDocuments({ supabase, companyId: COMPANY, userId: USER, consentId: 'c1' })
+
+    expect(result).toMatchObject({ scanned: 1, linked: 0, unmatched: 1 })
+    expect(result.unmatchedSamples[0]).toMatchObject({ uploadId: 'up-1', voucher: '(unresolved)' })
+    expect(mockDownload).not.toHaveBeenCalled()
+  })
+
+  it('handles a company with no uploads as an all-zero no-op', async () => {
+    const supabase = wireBokio()
+    mockFetchUploads.mockResolvedValue([] as never)
+
+    const result = await importProviderDocuments({ supabase, companyId: COMPANY, userId: USER, consentId: 'c1' })
+
+    expect(result).toMatchObject({ scanned: 0, linked: 0, skipped: 0, unmatched: 0, failed: 0 })
+    expect(mockUpload).not.toHaveBeenCalled()
   })
 
   it('is best-effort: a failed receipt is counted, not thrown, and the sweep continues', async () => {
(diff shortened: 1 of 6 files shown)


jhagberg:
Ran an independent review pass over the diff and pushed two fixes in `32bbda7a`:

- **Stable pagination order** — added `.order('id')` to the paged `journal_entries` / `document_attachments` / `fiscal_periods` reads. `fetchAllRows` pages with `.range()`, and without a deterministic order PostgREST can skip/repeat rows once a table crosses one page (this company's `journal_entries` will exceed 1000 across several migrated years), which would have defeated both voucher resolution and the sha256 dedup.
- **No silently-dropped receipts** — every upload carrying a `journalEntryId` now stays in scope; one whose Bokio entry number didn't parse (or resolves to no verifikat) is counted as `unmatched` instead of vanishing from the best-effort report. Added unresolvable-ref and zero-uploads tests.

One residual gap I left deliberately, flagging for your call: the `(company_id, sha256)` idempotency is a pre-insert snapshot check, so two **concurrent** `/import-documents` runs for the same company could both miss and both insert — and the rows are immutable. There's no unique index on `(company_id, sha256_hash)` to lean on (you noted the hash isn't a dedup key), and adding one would affect every upload path, not just this one, so it felt out of scope for this PR. Given the route is manually triggered per company the window is small, but if you'd want a partial unique index or an advisory lock around the sweep, happy to add it here or in a follow-up.

PR #829, opened 2026-06-29
fix(ci): switch compliance review to pull_request_target so fork PRs can post comments

flojon:
## Summary

- `on: pull_request` gives fork PRs a read-only `GITHUB_TOKEN` even when `pull-requests: write` is declared — this is a GitHub security restriction added in 2021
- Switching to `pull_request_target` runs the job in the base-repo context so the token has write access to post/update the review comment
- Added `ref: ${{ github.event.pull_request.head.sha }}` on the checkout step so the PR's actual code is reviewed, not the base branch

## Test plan

- [ ] Open or sync a fork PR and confirm the compliance review job posts its comment successfully
- [ ] Confirm the diff reviewed is from the PR branch, not main

### DIFF
--- a/.github/workflows/swedish-compliance-review.yml
+++ b/.github/workflows/swedish-compliance-review.yml
@@ -1,7 +1,7 @@
 name: Swedish Accounting Compliance Review
 
 on:
-  pull_request:
+  pull_request_target:
     types: [opened, synchronize, reopened]
 
 permissions:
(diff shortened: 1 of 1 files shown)


flojon:
Closing in favour of #830, which implements the same fix using a safer two-stage `workflow_run` pattern that avoids the `pull_request_target` + fork-head-checkout supply-chain risk.

PR #985, opened 2026-07-10
fix(transactions): stop defaulting supplier-invoice payment account to a stale private-funds setting

flojon:
## Summary
- `match-supplier-invoice` (POST + preview) defaulted the credited cash account from `company_settings.last_supplier_payment_account`, a sticky setting written by the manual mark-paid "betald med privata medel" flow (private-funds payments, e.g. crediting 2893 skuld till aktieägare).
- Once that setting held `2893` from an unrelated private payment, every later match against a **real** bank transaction reused it instead of the transaction's actual bank account, silently booking genuine bank payments as shareholder-loan repayments (root cause of a real misbooked GleSYS payment: 2440/2893 instead of 2440/1930).
- Fix: resolve the credit account from the matched transaction's own `cash_account_id -> cash_accounts.ledger_account` (falling back to `1930` when unlinked), mirroring the existing settlement-account lookup already used in `transactions/[id]/categorize/route.ts`. The resolution logic is now shared via `resolveSettlementAccount` (`lib/bookkeeping/settlement-account.ts`), consumed by the POST route, its preview, and `categorize/route.ts`. Neither match-supplier-invoice route reads `last_supplier_payment_account` anymore; that setting stays scoped to seeding the manual mark-paid UI's default account picker.
- The resolved `paymentAccount` is now threaded into **all three** journal-entry branches (pure-SEK clearing, FX via `createSupplierInvoicePaymentEntry`, and cash-method via `createSupplierInvoiceCashEntry`) — the FX/cash-method branches were initially scoped out but were fixed in a later commit once it became clear they already accepted the parameter and just weren't receiving it.
- **Error-handling change on the lookup itself**: `resolveSettlementAccount` now distinguishes a *hard query error* from a *missing mapping*. A found row with no `ledger_account` still warns and falls back to `1930` (a data-integrity gap, not a query failure). A hard Supabase error on the `cash_accounts` lookup now throws `BookkeepingDatabaseError` instead of silently defaulting to `1930` as it used to — the throw isn't caught locally, so it propagates to `withRouteContext`'s/`withApiV1`'s outer catch-all and surfaces as a structured `500` (caller can retry). Rationale: an explicit `cash_account_id` almost certainly resolves to a non-1930 account, so degrading to 1930 on infra flakiness would reproduce this PR's own bug class, just triggered by a DB hiccup instead of a stale setting — a retryable failed request beats a wrong verifikat that needs a storno (BFL 5 kap).
- A column comment (migration `20260711140000`) documents that `last_supplier_payment_account` must never be read to resolve a matched transaction's settlement account.
- Remaining out-of-scope gaps (tracked separately, not touched here): the customer-side match-invoice route still hardcodes `1930` unconditionally; the `/api/v1` (MCP-facing) match-supplier-invoice and categorize routes don't yet use `resolveSettlementAccount` (tracked in #1000).

## Test plan
- [x] `npx vitest run "app/api/transactions/[id]/match-supplier-invoice" app/api/transactions app/api/supplier-invoices lib/bookkeeping` — all passing, including regression tests for the stale-2893 scenario, non-1930 linked cash accounts, and a cash-account lookup error
- [x] `npm run check:guards` — passes
- [x] `npx eslint` on all changed files — no issues
- [x] `npx tsc --noEmit` — no errors

## Summary by CodeRabbit

- **Bug Fixes**
  - Supplier invoice matching now credits the bank account linked to the matched transaction.
  - Previews use the same settlement-account resolution as completed matches.
  - Transactions without a linked cash account continue to use the standard fallback account.
  - Database lookup failures now stop processing safely without creating journal entries.
  - Transaction categorization applies the resolved settlement account consistently.

- **Tests**
  - Added coverage for linked, alternate, missing, and invalid settlement-account scenarios.


flojon:
Followed up on the compliance-bot note about \`last_supplier_payment_account\` scope. Audited every reference in the codebase — outside migrations/types/tests, it's read or written in exactly 5 places, and all of them are the manual "betald med privata medel" mark-paid flow:

- \`app/api/supplier-invoices/[id]/mark-paid/route.ts\` — writes it, but only when the caller explicitly picks a \`payment_account\` (skipped on the MCP/agent path, which sends none — no unwanted churn).
- \`app/api/supplier-invoices/[id]/mark-paid/preview/route.ts\` — reads it as the picker's default when no \`payment_account\` is passed yet.
- \`app/(dashboard)/supplier-invoices/[id]/page.tsx\` — frontend dialog, pre-fills the picker.
- \`lib/bookkeeping/supplier-payment-lines.ts\` — mentioned only in a comment explaining the historical bug.
- \`match-supplier-invoice/route.ts\` + \`preview/route.ts\` — mentioned only in comments now, explaining why they no longer read it (this PR's fix).

No other matched-transaction code path reads this setting. Also confirmed the \`/api/v1\` (MCP-facing) supplier-invoice mark-paid route doesn't touch this column at all — it hardcodes the credit to 1930 unconditionally, so the sticky setting is unreachable outside the dashboard's manual mark-paid dialog. Blast radius is fully contained; the column comment added in the latest commit documents this for anyone reaching for the field in the future.

flojon:
@coderabbitai All three nitpicks from your review are addressed:

1. **Duplicated settlement-account resolution** — extracted into \`resolveSettlementAccount\` (\`lib/bookkeeping/settlement-account.ts\`), now shared by match-supplier-invoice POST, its preview, and \`categorize\`. (34d5d35)
2. **Missing POST test for a \`cash_accounts\` lookup error** — added; asserts fallback to 1930 and that the warning logger fires. (7cbbec1)
3. **Missing preview parity test for a non-1930 linked account** — added, matching the equivalent POST-route case. (7cbbec1)

flojon:
Re the Swedish Accounting Compliance Review — the remaining items are now addressed:

- **FX branch / cash-method branch still defaulting to 1930**: \`createSupplierInvoicePaymentEntry\` and \`createSupplierInvoiceCashEntry\` already accepted an optional \`paymentAccount\` parameter, but the route only threaded the resolved value into the pure-SEK clearing branch. Both branches now receive it too — one-line fix per call site, no new code path. (2112701)
- **Silent fallback risk**: \`resolveSettlementAccount\` now also warns (and still falls back to 1930) when \`cash_account_id\` resolves to a row with no \`ledger_account\` — previously that only warned on a hard query error, so a bound-but-empty row failed silently. (2112701)
- **\`last_supplier_payment_account\` scope documentation**: added a column comment via migration (20260711140000) stating it must never be read to resolve a matched transaction's settlement account. Full audit of every remaining reference to that column is in the comment above.

flojon:
Merge conflict resolved and last outstanding review comment addressed:

- **Merge conflict**: `DECISIONS.md` had diverged (both this branch and `main` appended new entries after the same line). Resolved by keeping `main`'s already-merged entries in place and appending this PR's entries after them — no content lost, no reordering of already-merged history. `app/api/transactions/[id]/categorize/route.ts` auto-merged cleanly (no overlap between this PR's settlement-account change and main's other edits there).
- **Last open CodeRabbit nitpick** (2026-07-12 review round): `lib/bookkeeping/__tests__/settlement-account.test.ts` asserted the thrown error type via `constructor: BookkeepingDatabaseError` inside `toMatchObject`, which the bot correctly flagged as non-idiomatic. Switched to `rejects.toBeInstanceOf(BookkeepingDatabaseError)` plus a separate `toMatchObject` for the `operation`/`message` fields.
- **PR body**: updated to reflect that the FX and cash-method branches were fixed in the final commit (previously the body still said they were out of scope), per @jakobwennberg's note above.

All previously-raised nitpicks were already resolved in earlier commits (confirmed by CodeRabbit's own follow-up). Verification: `vitest run` across match-supplier-invoice/transactions/supplier-invoices/bookkeeping (all passing), `tsc --noEmit` clean, `eslint` clean, `check:guards` passing.

flojon:
One more thing worth calling out explicitly for reviewers, since it's a behavior change beyond the original bug fix: **`resolveSettlementAccount`'s error handling for a failed `cash_accounts` lookup changed from silent-fallback to hard-fail.**

- **Before this PR's earlier commits**: any error resolving the settlement account (including a transient DB error) would warn and silently default to `1930`.
- **Now** (`lib/bookkeeping/settlement-account.ts`): only a *missing* `ledger_account` on an otherwise-found row still warns-and-falls-back to `1930` (genuine data-integrity gap, not a query failure). A **hard query error** (`error` truthy from Supabase) now throws `BookkeepingDatabaseError('resolve_settlement_account', ...)` instead.
- That throw is **not caught locally** in either the POST or preview route — it propagates out of the handler to `withRouteContext`'s (and `withApiV1`'s, for the categorize v1 caller) outer catch-all, which converts any `isBookkeepingError()` throw into a structured `500` via `errorResponse`/`v1ErrorResponse`. So a transient Supabase blip on this one lookup now fails the whole match/categorize request (caller can retry) rather than completing the booking against a possibly-wrong account.

Rationale (also in `DECISIONS.md`, 2026-07-12 entry): once a transaction has an *explicit* `cash_account_id`, it almost certainly resolves to a non-`1930` account, so degrading to `1930` on infra flakiness reproduces the exact misbooking class this PR exists to fix — just triggered by a DB hiccup instead of a stale setting. A failed request that the caller can retry is preferable to a wrong verifikat that needs a storno to correct (BFL 5 kap). Covered by the `throws a BookkeepingDatabaseError instead of silently falling back when the lookup errors` test in `lib/bookkeeping/__tests__/settlement-account.test.ts`, plus route-level regression tests asserting the 500 in both `match-supplier-invoice/route.ts` and its preview.

flojon:
@coderabbitai review

Issue #1016, opened 2026-07-14
create_transactions never binds a cash_account_id: manually-fed 193x accounts get no reconciliation, and matching falls back to 1930


reinthal:
## Summary

Transactions ingested outside the Enable Banking (PSD2) flow, via the MCP tool `gnubok_create_transactions` (CSV / Airtable / manual ingestion), are never bound to a `cash_accounts` row: `commitCreateTransaction` inserts them with `cash_account_id = null`. As a result, ledger accounts fed this way (e.g. `1935` Wise SEK) behave as if they had no kassakonto, and every consumer of the `transaction -> cash_account -> ledger_account` resolution chain degrades:

1. **Reconciliation errors out.** `gnubok_get_reconciliation_status(account_number="1935")` returns a validation error `"Okänt kassakonto för det här företaget"` because no `cash_accounts` row exists for `1935`.
2. **No unbooked-transaction detection** for these accounts: reconciliation iterates per cash account, and a manually-fed 193x account has no cash account to iterate over.
3. **The "Matcha mot befintlig verifikation" dialog falls back to 1930.** A transaction with `cash_account_id = null` cannot be resolved to its economic account, so the candidate verifikat list is filtered on `1930` regardless of which account the transaction belongs to ("Inga omatchade verifikationer på 1930 i perioden").

This is the remaining leg of the #985 / #986 / #987 root cause. Those PRs fixed the `transaction -> cash_account -> ledger_account` resolution chain for *forward booking*; this issue is about the *creation* side: manually-fed accounts never get the `cash_account_id` link in the first place, so the fixed chain has nothing to resolve.

### Correction to the original framing

An earlier internal framing assumed `cash_accounts` rows *cannot* exist without a bank connection. That is **not** the case, and no schema change is required:

- `cash_accounts.bank_connection_id` is **nullable by design** (`supabase/migrations/20260519110000_cash_accounts.sql`, lines 29-31), and `source` already permits `'manual'`.
- A manual `1930` row (`source = 'manual'`, `bank_connection_id = null`) is **seeded for every company** by `supabase/migrations/20260519154732_seed_default_cash_account.sql` and baked into `create_company_with_owner()`.
- Per #916, manual (null-connection) rows are already first-class: `upsertFromPsd2` promotes a manual holder in place on reconnect.

[...]

PR #1018, opened 2026-07-14
feat(import): add Wise (TransferWise) CSV import format

reinthal:
## What

Adds **Wise (TransferWise) CSV import** as a bank-file format plugin. Wise exports a single multi-currency transaction history (one row per balance movement); this makes it importable through the existing bank-file wizard (upload → preview → confirm → execute) with no new UI.

Motivated by users running Wise Business balances booked on BAS `1935`–`1938` with no Enable Banking feed (same cohort as #1016).

## How it maps a Wise row

- **Delimiter/quoting**: comma-delimited, `.` decimal, quoted fields (dates contain a space) → uses the existing quote-aware `parseCSVLine`.
- **Sign**: `Direction` IN → income, OUT → expense. Booked on the side that moved: target for IN, source for OUT.
- **Currency**: preserved **native** (USD stays USD, SEK stays SEK). SEK conversion is left to the existing downstream FX/booking pipeline (Riksbanken) — the parser never converts. Booking in the account's own currency via the Wise API rate is a deliberate follow-up.
- **Fees**: a non-zero `Source fee` / `Target fee` becomes its **own** negative `"Wise avgift"` transaction so the fee books separately (e.g. 6570) and the balance ties out — nothing folded or dropped.
- **Status**: only `COMPLETED` rows import; pending/cancelled/refunded are skipped.
- **Dedup**: `external_id` keys on the stable Wise ID (`TRANSFER-…`, `PLAN_ORDER-…`, plus a `-fee`/`-tgtfee` suffix for fee rows) via a new `wise` branch in `generateExternalId`, so re-importing the same statement dedups exactly instead of relying on the row hash.

## Changes

- `lib/import/bank-file/formats/wise.ts` (new) — detect + parse.
- `lib/import/bank-file/parser.ts` — register in `FORMATS`, `wise` branch in `generateExternalId`.
- `lib/import/bank-file/types.ts` — `'wise'` in `BankFileFormatId`.
- `components/import/BankFileUploadStep.tsx` — manual-format picker entry.
- `app/api/v1/companies/[companyId]/imports/bank/route.ts` — `wise` in the `format` enum + doc string.

## Scope

Format plugin only (per the feature discussion). Not included, as follow-ups:

- Per-currency auto-routing of Wise rows to manual kassakonton (would build on #1016's `ensureManualCashAccount`).
- Wise-API exchange-rate lookup / booking in the account's default currency.

Multi-currency note: the statement can mix currencies, so `stats.total_income`/`total_expenses` sum across currencies (display-only, same naive behavior as the existing SEK-only formats). Each transaction carries its own currency and books correctly.

## Tests

`lib/import/bank-file/__tests__/parser.test.ts` (Wise block, using the real sample export): auto-detection, IN/OUT signing + moved-side currency, fee row splitting (incl. no row for a 0.00 fee), stable `external_id` for main and fee rows, COMPLETED-only filtering.

197 parser tests green; 1069 tests green across `lib/import` + `app/api/import` + `app/api/v1/companies`. Lint clean on touched files.

## Summary by CodeRabbit

* **New Features**
  * Added support for importing Wise (TransferWise) CSV bank statements.
  * Wise files can be auto-detected or selected manually during upload.
  * Correctly interprets transaction direction for booking and preserves per-row currencies.
  * Non-zero Wise fees are imported as separate “Wise avgift” transaction rows, with improved duplicate detection.

* **Documentation**
  * Updated bank import endpoint guidance and input validation to include the Wise format.

* **Tests**
  * Added Jest coverage for Wise parsing, fee handling, and stricter rejection/skip rules for invalid or non-completed rows.


reinthal:
### Known limitations (scoped) — tracked in #1019

This PR intentionally covers only the **Transaction history** export (`Direction` + `(after fees)` columns, `IN`/`OUT`, `COMPLETED` rows, inline fees). Not covered here, filed as follow-up #1019:

- **Per-currency Balance statement CSV** (`statement_<id>_<CUR>_...csv`, signed `Amount`, no `Direction`) — not detected (does not misfire; falls through to unknown).
- **Non-IN/OUT `Direction`** (e.g. `NEUTRAL` conversions/cashback) — currently bucketed as income; needs explicit handling.
- **Cross-currency conversions** (two-sided) — single-row/single-currency model books only one leg.
- **Statuses other than `COMPLETED`** (`REFUNDED`/chargeback) — skipped, not booked.

Nailing these correctly needs a richer real export (conversion + card + refund) — see #1019.

reinthal:
Addressed the CodeRabbit findings in d97b20d0:

- **Strict amount parse** — `parseWiseAmount` now rejects non-numeric/thousands-separated values (`"12abc"`, `"1,234"`) via a `/^-?\d+(\.\d+)?$/` guard instead of `parseFloat`; a bad amount skips the row with a warning.
- **Exact `COMPLETED`** — a blank/missing status no longer bypasses the filter (`status !== 'COMPLETED'` skips).
- **Unknown `Direction`** — deliberately **fails hard** (throws → `BANK_FILE_PARSE_FAILED`) rather than skipping/guessing, per maintainer instruction. A blank or `NEUTRAL` (conversion/cashback) row aborts the whole import so nothing is mis-signed. Proper conversion handling stays tracked in #1019.
- **No invented currencies** — a missing movement currency skips the row with a warning (no `SEK` default); a fee with no currency of its own is dropped with a warning instead of inheriting the movement currency.

New tests cover all five. Note: the hard-fail on `Direction` is intentional and stricter than the suggested skip-with-warning.

PR #1285, opened 2026-07-29
fix(invoices): stamp paid_at with the actual payment date, not now()

flojon:
## Summary
- Supplier-invoices overview status chip ("Betald {date}") was showing the moment an invoice was clicked paid, not the payment date actually selected/matched, because `paid_at` was written as `new Date().toISOString()` instead of the payment date already computed for the journal entry on the same request.
- Fixed all 8 call sites with the same root cause: dashboard supplier-invoice mark-paid, dashboard/v1 customer-invoice `settleInvoicePayment`, dashboard/v1 bank-transaction match (invoice + supplier invoice), transaction-to-journal-entry linking, and the agent/MCP pending-operations commit path (mark-invoice-paid + match-transaction-invoice).
- Manual mark-paid flows now use `paymentDate` (the value the user picked in the dialog); bank-match flows use `transaction.date` (the bank's own settlement date) — both already flowed into the journal entry's `entry_date` on the same request, so this only corrects `paid_at` to match what was already being booked.

## Test plan
- [x] `npm test` on all touched directories: 89 files / 1256 tests pass
- [x] Added regression tests asserting `paid_at`/backdated `payment_date` on: supplier-invoices mark-paid, customer-invoices mark-paid, match-invoice, match-supplier-invoice, link-journal-entry, and both agent/MCP commit paths
- [x] Verified each new test fails against the pre-fix `now()` behavior (checked by temporarily reverting one call site)
- [x] `npx eslint` clean on all touched files (pre-existing unrelated warnings only)
- [ ] No new tests added for the two v1 API routes (`match-invoice`/`match-supplier-invoice` under `app/api/v1/...`): they had no existing test scaffolding and mirror the already-tested dashboard routes exactly (see DECISIONS.md entry)

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated invoice and supplier-invoice `paid_at` to be derived from the actual payment/transaction date (including backdated values) instead of the request/server processing time.
  * Applied consistently across mark-paid flows, invoice matching, supplier-invoice matching, journal linking, and pending-operation commits.
  * `paid_at` is now anchored to midnight UTC for the selected date for consistent results across timezones.
* **Tests**
  * Expanded unit and integration coverage to assert `paid_at` in both HTTP responses and queued DB updates, including v1 route scenarios.
* **Documentation**
  * Updated the decision log with the revised `paid_at` behavior and timezone notes.


flojon:
Good catch on the date-vs-timestamp mismatch — fixed in 623f489f.

`paid_at` is `timestamptz` while `payment_date`/`transaction.date` are plain `date` columns, so writing the bare `yyyy-MM-dd` string left Postgres to interpret it through the session's `TimeZone` GUC — the same class of ambiguity this PR set out to remove, just relocated to the DB layer. All 9 write sites now anchor to `` `${date}T00:00:00Z` ``, matching the existing `tax_paid_at` precedent in `extensions/general/skatteverket/lib/agi-tax-settlement.ts:219`.

Verified: `parseISO('...T00:00:00Z')` formatted with `date-fns` renders the correct date for `TZ=UTC` and `TZ=Europe/Stockholm` (hosted default and the only production audience). A negative-UTC-offset host (e.g. self-hosted Docker run in a US timezone) could still roll the display back a day — that's a pre-existing, systemic tradeoff the codebase already accepted for `tax_paid_at`, not something newly introduced here, and fixing it repo-wide would mean forcing a display timezone in `lib/utils.ts` globally, well beyond this bug's scope. Noted in `DECISIONS.md`.

On the untested v1 routes: agreed that's a gap given `paid_at` is compliance-relevant, but adding the parity tests would mean building new test scaffolding from scratch for those two routes (no existing coverage to extend) — happy to do that in a follow-up if you'd rather not merge without it, just flagging it's a larger addition than the rest of this diff.

All 1256 tests across the touched directories still pass; re-ran them after this change.

flojon:
Addressed all 3 in 4b9ab653:

- **Event projection coverage**: extended the `invoice.paid` event assertion in `mark-invoice-paid.test.ts` to also require `paid_at: '2026-03-30T00:00:00Z'` on the payload, alongside the existing DB-update assertion.
- **match-invoice test asserting response only**: now also asserts the persisted `invoices.update` call's `paid_at`, not just the response body.
- **Untested v1 routes**: added `app/api/v1/companies/[companyId]/transactions/[id]/match-invoice/__tests__/route.test.ts` and the `match-supplier-invoice` sibling — 401 (no bearer token), 400 (`VALIDATION_ERROR` on missing id), 404 (not-found), and the happy path asserting `paid_at` equals `transaction.date` in both the response and the persisted DB row.

All 1285 tests across the touched directories pass (92 files), including the 8 new ones. Also had to `npm install` in this environment to pick up `stripe`/`swr`/`posthog-js` — already-declared `package.json` deps that weren't installed, unrelated to this PR's diff, no `package-lock.json` changes.

jakobwennberg-oss:
Closing: this fix landed on main via #1332 (merged 2026-08-02), which ported this PR's payment-date correction onto current main, with credit in its body, and extended it to two more v1 routes plus the DB-side matching RPCs. The one deliberate change from this PR's approach: paid_at anchors at UTC noon instead of UTC midnight, which avoids the negative-offset display regression this PR's own DECISIONS entry documented as residual risk.

Every call site fixed here is fixed on main today, so a rebase would only reintroduce the midnight anchor. Thanks for tracking the root cause across all eight call sites, and apologies that this sat open without a pointer to #1332 for two weeks.

PR #1288, opened 2026-07-29
Fix crash clicking a "Tidigare motparter" suggestion

flojon:
## Summary
- Clicking a "Tidigare motparter" (previous counterparty) suggestion in the Bokför transaktion dialog navigated to the generic error page ("Något gick fel").
- Root cause: `handleOpenTemplateReview` (app/(dashboard)/transactions/page.tsx) built a fake `BookingTemplate` for the counterparty path with only `id`/`name_sv`, cast with `as BookingTemplate` to bypass the type checker. This left `debit_account`/`credit_account` undefined, so `QuickReviewDialog`'s `defaultAccount` prop resolved to `undefined`, and the unconditional `accountOverride.startsWith('2')` check threw a `TypeError` during render — caught by the dashboard error boundary.
- Fix: populate the fake template with `debit_account`/`credit_account` already returned by the suggest-categories API, and guard the `startsWith` call defensively.

## Test plan
- [x] `npm run check:lint`
- [x] `npm run check:guards`
- [x] `npx vitest run --project unit lib/bookkeeping/__tests__/counterparty-templates.test.ts components/transactions/__tests__` (155 passed)
- [x] `npx tsc --noEmit` clean on touched files
- [ ] Manual click-through on a transaction with previous counterparty suggestions (not run in this environment)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved transaction template reviews by preserving selected debit and credit account details.
  * Prevented errors when an account override is not selected during quick review.
  * Maintained correct liability and equity account handling when account information is available.

### DIFF
--- a/app/(dashboard)/transactions/page.tsx
+++ b/app/(dashboard)/transactions/page.tsx
@@ -2163,7 +2163,12 @@ export default function TransactionsPage() {
         transaction,
         category: transaction.amount < 0 ? 'expense_other' : 'income_services',
         label: cpSuggestion.name_sv,
-        template: { id: templateId, name_sv: cpSuggestion.name_sv } as BookingTemplate,
+        template: {
+          id: templateId,
+          name_sv: cpSuggestion.name_sv,
+          debit_account: cpSuggestion.debit_account,
+          credit_account: cpSuggestion.credit_account,
+        } as BookingTemplate,
         templateId: undefined,
         linePattern: cpSuggestion.line_pattern ?? null,
         defaultDimensions: cpSuggestion.default_dimensions ?? null,
--- a/components/transactions/QuickReviewDialog.tsx
+++ b/components/transactions/QuickReviewDialog.tsx
@@ -202,7 +202,7 @@ export default function QuickReviewDialog({
   const isIncome = tx.amount > 0
   const isCounterpartyTemplate = !!(counterpartyLinePattern && counterpartyLinePattern.length > 0)
   const isTemplateBooking = !!templateId || isCounterpartyTemplate
-  const isLiabilityAccount = accountOverride.startsWith('2')
+  const isLiabilityAccount = !!accountOverride?.startsWith('2')
   // For non-SEK transactions, the verifikation and the headline must show
   // the SEK-converted total: the mall/category booking always posts in SEK.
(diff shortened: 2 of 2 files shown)


flojon:
Addressing the two notes from the Swedish Accounting Compliance Review:

**1. Unvalidated BAS accounts entering the booking template**

Traced the actual booking path for this flow (`handleQuickReviewConfirm` in `app/(dashboard)/transactions/page.tsx`, counterparty branch, ~line 2230): when a "Tidigare motparter" suggestion is confirmed, the client sends only `{ is_business: true, counterparty_template_id, dimensions? }` to `POST /api/transactions/{id}/categorize` — `category`, `vat_treatment`, and `account_override` (all derived from the fake `template` object I patched) are **not** sent for this branch. Server-side, `categorize/route.ts` re-fetches the authoritative `categorization_templates` row by `counterparty_template_id` (line ~299) and rebuilds `mappingResult` from that DB row via `buildMappingResultFromCounterpartyTemplate`, then runs the existing pre-validation against the chart of accounts (line ~403, returns `ACCOUNTS_NOT_IN_CHART` if a learned account was deactivated). So the `debit_account`/`credit_account` I added to the fake template are UI-display-only (populate `defaultAccount` / the account picker) and never reach the booking engine — the real accounts are always re-resolved and re-validated server-side from the DB, unaffected by this diff.

**2. `isLiabilityAccount` tolerating `undefined`**

`isLiabilityAccount` only gates local UI (whether to show the liability-account variant of the account/VAT picker) in this dialog. For the counterparty path specifically, `vat_treatment` isn't even transmitted (per #1), so there's no path where this flag changes what gets posted to the ledger — it's presentational. Previously an `undefined` `accountOverride` threw synchronously on every counterparty-suggestion click (the crash this PR fixes); the `?.` just lets the component render instead of crashing, defaulting to "not a liability account" for display purposes only, same as it would for any other still-loading/unset state.

No changes needed based on these notes — happy to add a code comment pointing at the re-validation in `categorize/route.ts` if that'd help future readers, let me know.

mattssonn:
Thanks for the contribution. This fix has since been incorporated in the more complete, type-safe implementation merged via #1291, so I’m closing this pull request as superseded.

PR #1294, opened 2026-07-30
fix(transactions): block invalid invoice match targets

### LINKED ISSUE #1260: [Transactions] Matching a transaction against an already-paid invoice should give a clearer error
## Problem
`InvoiceMatchDialog` computes its amount-diff purely from `remaining_amount`/`total`, with no check on the target invoice's `status`. When the target invoice turns out to be already fully paid (`status=paid`, `remaining_amount=0`) — most commonly reached via the stale-pointer bug in the companion issue (#1259), but possible any time a match candidate is no longer open — the dialog shows the same generic "Beloppen skiljer sig" / partial-payment warning it would show for a genuinely partial payment. That's misleading: the real problem isn't a partial payment, it's that this invoice isn't a valid match target at all anymore. A user reading "fakturan blir delbetald" has no way to tell the difference from the message alone.

## Solution
Left to the maintainers' judgment on exact UX (distinct blocking message vs. distinct-but-overridable message, and whether to extend beyond `status=paid` to other non-open statuses like credited/cancelled) — flagging the gap rather than prescribing the fix. Should cover both the supplier-invoice and customer-invoice branches of the dialog, since both share the same amount-diff logic.

## Why
Clearer error messaging prevents user confusion and reduces the chance someone force-confirms a match against an invoice that shouldn't be touched, expecting a "delbetald" outcome rather than the real (invalid-target) situation.

## Where

**Files:**
- `components/transactions/InvoiceMatchDialog.tsx:507-576` — the amount-comparison block (`amountsMatch` / `isOreRounding` / mismatch branches) has no case for "target invoice status is not open"; that's where a distinct message/state would need to be added
- `components/transactions/InvoiceMatchDialog.tsx:454-498` — the invoice-detail display blocks for both the customer (`isCustomerInvoice`) and supplier (`isSupplierInvoice`) branches, which would need the same treatment
- `messages/sv.json:2587-2590` (`amounts_differ`, `amount_diff`, `partial_payment_note`) and the matching `messages/en.json` keys — new string(s) would go here per the i18n conventions in `.claude/rules/i18n.md`

**Related files:**
- `components/transactions/MatchAllocationDialog.tsx` — the batch-allocation dialog shares the same underlying data (`remaining_amount` filtering) and may want the same treatment for consistency

---
*Filed via /create-ticket*

mattssonn:
## Summary

- classify invoice match targets as matchable, settled, or not open
- block stale paid and non-open customer or supplier invoice targets with clear Swedish and English copy
- suppress stale duplicate warnings, voucher previews, and confirm outcomes for invalid targets
- preserve existing partial-payment behavior and add focused regression coverage

## Root cause

`InvoiceMatchDialog` treated every non-matchable target as a normal amount difference. A paid target with no remaining balance therefore looked like a partial payment, while cancelled, credited, disputed, reversed, or otherwise non-open targets were later mislabeled as fully paid.

## Validation

- `npm test -- lib/invoices/__tests__/matchable-statuses.test.ts lib/worklist/__tests__/categories.test.ts` (36 tests passed)
- changed-file ESLint passed
- full `npm run lint` completed with 0 errors (257 existing warnings)
- Swedish and English namespace parity and `git diff --check` passed
- full unit run passed 940 files and 11,939 tests; one unrelated existing guard suite, `scripts/checks/__tests__/format-currency-sek-label.test.ts`, failed to parse locally before running tests

Fixes #1260

## Summary by CodeRabbit

* **New Features**
  * Improved invoice matching to distinguish between matchable, fully settled, and no-longer-open invoices.
  * Added clearer, localized messages explaining why an invoice cannot be matched.
* **Bug Fixes**
  * Blocked invoices no longer display voucher previews, confirmation panels, or enabled confirmation actions.
  * Prevented unavailable match routes from exposing misleading payment or bookkeeping options.
* **Tests**
  * Added coverage for invoice target states, including partially paid, settled, unavailable, and invalid invoices.


mattssonn:
@coderabbitai review

PR #1295, opened 2026-07-30
fix(mcp): scope VAT close reconciliation account

### LINKED ISSUE #1290: [Bank Recon] vat_close_check reports phantom reconciliation diff (unscoped cash account)
## Problem
`gnubok_vat_close_check` can report a nonzero bank reconciliation difference in its `bank_unreconciled` blocker (e.g. "-5599.97 kr") with `count: 0` unmatched transactions and `count: 0` unmatched GL lines on account 1930 — i.e. the blocker fires with no actual unmatched items to point to. Calling `gnubok_get_reconciliation_status` directly for the same period/company (both a narrow date-range call and a cumulative-to-date call) consistently returns `is_reconciled: true` with `difference: 0`, never reproducing the figure the close-check reported.

## Root cause
Both paths call the same underlying `getReconciliationStatus` (`lib/reconciliation/bank-reconciliation.ts:477`), so there's no duplicated logic — the bug is in how the two call sites populate its trailing parameters:

- `gnubok_get_reconciliation_status` (`extensions/general/mcp-server/server.ts:8977-8985`) looks up the `cash_accounts` row for the target account (default `1930`) and explicitly passes `accountNumber`, `currency`, `cashAccountId`, and `includeUnassigned = is_primary`.
- `computeVatCloseCheck` (`extensions/general/mcp-server/server.ts:1780`) calls `getReconciliationStatus(supabase, companyId, start, end)` with only 4 positional args — `cashAccountId` is left `undefined`.

In `scopeTransactionsToAccount` (`lib/reconciliation/bank-reconciliation.ts:227-251`), when `cashAccountId` is falsy the function falls through to `return query.eq('currency', currency)` (line 250) — i.e. **no `cash_account_id` scoping at all**. It sums every SEK bank transaction across *all* of the company's cash accounts (e.g. checking + savings), producing `bankTotal` (`bank-reconciliation.ts:621`). Meanwhile the GL side stays correctly scoped to `account_number = '1930'` only (`bank-reconciliation.ts:572`). The resulting `difference = bankTotal - glPeriodMovement` (`bank-reconciliation.ts:668`) picks up the other cash account's transaction volume with nothing on the GL side to offset it. Those extra transactions are already matched to their own account's journal entries, so they never show up as unmatched — explaining `count: 0` alongside a nonzero `difference`.

## Solution
In `computeVatCloseCheck` (`server.ts:1780`), before calling `getReconciliationStatus`, resolve the primary/default `cash_accounts` row for `'1930'` the same way the `gnubok_get_reconciliation_status` handler does, and pass `bankAccount='1930'`, the resolved `currency`, `cashAccountId`, and `includeUnassigned = Boolean(cashAccount.is_primary)` — instead of relying on the 4-arg call's implicit defaults.

## Why
This produces a false-positive VAT-close blocker with no actionable unmatched items to investigate, for any company with more than one SEK cash account. It's confusing and erodes trust in the close-check output — a user chasing "0 unmatched transactions, 0 unmatched GL lines" but a nonzero diff has no way to resolve it through the tool itself, since the standalone reconciliation tool (correctly scoped) shows everything is fine.

## Where

**Files:**
- `extensions/general/mcp-server/server.ts:1780`: `computeVatCloseCheck`'s call to `getReconciliationStatus` — missing `cashAccountId`/`currency`/`includeUnassigned` args
- `extensions/general/mcp-server/server.ts:8977-8985`: the correctly-scoped reference implementation in the `gnubok_get_reconciliation_status` tool handler — mirror this lookup
- `lib/reconciliation/bank-reconciliation.ts:227-251` (`scopeTransactionsToAccount`): the fallback-to-unscoped behavior when `cashAccountId` is falsy — worth considering whether this should throw/warn instead of silently including all cash accounts
- `lib/reconciliation/bank-reconciliation.ts:477` (`getReconciliationStatus`), `:572` (GL scoping), `:621` (`bankTotal`), `:668` (`difference` calc)
- `extensions/general/mcp-server/server.ts:1825-1829`: the `bank_unreconciled` blocker construction that surfaces the bad figure

---
*Found via investigation of a real VAT close-check false positive.*

mattssonn:
## Summary

- resolve the `1930` cash-account identity before the VAT close check runs bank reconciliation
- share the established MCP reconciliation scope resolution so currency, cash-account ID, and unassigned-row behavior cannot drift between handlers
- add regression coverage for primary, non-primary, same-currency, and legacy fallback cases

## Root cause

`computeVatCloseCheck` called `getReconciliationStatus` with only four arguments. The bank side therefore fell back to every transaction in the account currency while the GL side remained scoped to `1930`, allowing another SEK cash account to create a phantom difference with no unmatched rows.

Closes #1290

## Validation

- affected reconciliation and VAT-close suite: 5 files, 85 tests passed
- `npm run lint`: passed with 0 errors and the existing warning baseline
- full unit suite: 941 files and 11,937 tests passed; one unchanged guard test failed before collection with a bare Windows-local `SyntaxError` (`scripts/checks/__tests__/format-currency-sek-label.test.ts`, byte-identical to `origin/main`)

### DIFF
--- a/extensions/general/mcp-server/__tests__/vat-close-check-reconciliation-scope.test.ts
+++ b/extensions/general/mcp-server/__tests__/vat-close-check-reconciliation-scope.test.ts
@@ -1,0 +1,151 @@
+/**
+ * gnubok_vat_close_check: cash-account scoping for bank reconciliation.
+ *
+ * getReconciliationStatus only isolates same-currency bank feeds when its
+ * cashAccountId is populated. These tests pin the VAT close check to the same
+ * account resolution used by the standalone reconciliation tool so another
+ * cash account cannot inflate 1930's bank total.
+ */
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+
+vi.mock('@/lib/reconciliation/bank-reconciliation', () => ({
+  getReconciliationStatus: vi.fn(async () => ({
+    is_reconciled: true,
+    difference: 0,
+    unmatched_transaction_count: 0,
+    unmatched_gl_line_count: 0,
+  })),
+}))
+
+import { getReconciliationStatus } from '@/lib/reconciliation/bank-reconciliation'
+import { computeVatCloseCheck } from '../server'
+
+const COMPANY_ID = 'aaaaaaaa-aaaa-4aaa-8aaa-aaaaaaaaaaaa'
+const PERIOD = { period_type: 'monthly', year: 2026, period: 1 }
+const getReconciliationStatusMock = vi.mocked(getReconciliationStatus)
+
+interface CashAccountFixture {
+  id: string
+  currency: string
+  is_primary: boolean
+}
+
+function mockSupabase(cashAccount: CashAccountFixture | null) {
+  const cashAccountFilters: Array<[string, unknown]> = []
+
+  const makeChain = (
+    rows: unknown[],
+    maybeSingleData: unknown = null,
+    eqCalls?: Array<[string, unknown]>,
+  ): Record<string, unknown> => {
+    const chain: Record<string, unknown> = {}
+    const settled = { data: rows, error: null, count: rows.length }
+    chain.range = () => settled
+    chain.single = async () => ({ data: null, error: null })
+    chain.maybeSingle = async () => ({ data: maybeSingleData, error: null })
+    chain.then = (resolve: (value: unknown) => void) => resolve(settled)
+    for (const method of [
+      'order', 'lte', 'gte', 'neq', 'in', 'is', 'select',
+      'limit', 'contains', 'filter', 'not', 'or',
+    ]) {
+      chain[method] = () => chain
+    }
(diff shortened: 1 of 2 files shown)


mattssonn:
Addressed the compliance finding in `da66c6a8`: cash-account lookup errors now fail closed instead of falling back to currency-only reconciliation, with a regression test proving `getReconciliationStatus` is not called after a lookup error.

Revalidated locally: affected suite 5 files / 86 tests passed, changed-file ESLint 0 errors, `git diff --check` clean, and `npm run check:guards` passed.

mattssonn:
Addressed the remaining low-severity compliance feedback in `7c8e807c`: the raw Supabase error is now recorded only through the structured server logger, while MCP callers receive a sanitized Swedish error. The fail-closed regression test was updated accordingly.

Revalidated: affected suite 5 files / 86 tests passed, changed-file ESLint 0 errors, diff check clean, and guard ratchet passed.

PR #1296, opened 2026-07-30
fix(vat): recover ruta 05 for null-rate custom accounts

### LINKED ISSUE #1289: [VAT] ruta05 silently drops custom 30xx accounts missing default_vat_rate
## Problem
`gnubok_vat_close_check` raises a false `OUTPUT_VAT_WITHOUT_SALES_BASE` error (ERROR severity, blocks VAT-close readiness) when a company's sales are booked to a custom class-3 revenue account like 3011 ("Försäljning tjänster inom Sverige, 25 % moms") instead of the hardcoded 3000-3003 range. Verified against a live company: a sale was booked as 9,725 kr to account 3011 and 2,431.25 kr output VAT to 2611 (an exact 25% match — the bookkeeping is correct), yet `ruta05` computed as 0 while `ruta10` correctly showed 2,431.25.

Root cause: `chart_of_accounts.default_vat_rate` was NULL for account 3011. `rutorFromTotals()` (`lib/reports/vat-declaration.ts:407-429`) only includes a fixed whitelist (3000/3001/3002/3003, `vat-declaration.ts:92-95`) plus whatever `fetchDynamicRuta05Accounts()` (`lib/reports/vat-revenue-accounts.ts:96-137`) finds — and that dynamic fallback requires `default_vat_rate` to already be 0.25/0.12/0.06 (`vat-revenue-accounts.ts:120`). An account with NULL `default_vat_rate` is excluded from both paths entirely — never even queried for ruta05 — despite `ruta10` being computed correctly from a different code path (2611 directly).

The upstream cause: `gnubok_create_account` (`extensions/general/mcp-server/server.ts:6049-6062`) only sets `default_vat_rate` from an explicit `args.default_vat_rate` — it never derives it from the BAS 2026 catalog, even though `BASReferenceAccount` (`lib/bookkeeping/bas-reference.ts:16-25`) doesn't carry a VAT-rate field at all, and many BAS account numbers unambiguously encode the rate in their own name/number convention (x011 = 25%, x012 = 12%, x013 = 6% for sales accounts). A user creating account 3011 without also passing `default_vat_rate=0.25` — easy to skip, since the rate is already in the account's name — silently produces a broken VAT declaration months later. Code comments in `vat-revenue-accounts.ts:1-24` reference this exact failure mode as a prior issue (#1261) that was only partially closed.

The check's own error hint text ("Kontrollera att intäktskonton (3001/3002/3003) är bokförda...") is also misleading, since it implies only those three accounts matter.

## Solution
1. **Backfill**: identify and fix existing `chart_of_accounts` rows where a class-3 (revenue) account has NULL `default_vat_rate` but a resolvable rate (either from BAS convention on the account number, or from a majority-vote of the VAT actually booked against it historically).
2. **Code fix**: `gnubok_create_account` should derive `default_vat_rate` from the BAS 2026 catalog/account-number convention when the caller doesn't supply it explicitly for a known BAS revenue account number, instead of leaving it NULL.
3. **Guardrail**: add validation (DB constraint or app-level check) preventing a class-3, VAT-liable-range account from being saved with NULL `default_vat_rate` going forward, closing this whole class of bug rather than just this one call site.
4. **Message fix**: update the `OUTPUT_VAT_WITHOUT_SALES_BASE` hint text to no longer imply only 3001/3002/3003 matter.

## Why
This silently produces an incorrect Skatteverket VAT declaration readiness check — the numbers are actually correct, but the tool blocks filing with a false ERROR, and the misleading hint text sends users chasing the wrong accounts. Any company with a custom 30xx/31xx revenue account (common — Accounted seeds no varugrupp accounts per `vat-declaration.ts:518`) is at risk. This is a compliance-adjacent workflow (VAT close), so a false blocker has real business impact even though it isn't a data-corruption bug.

## Where

**Files:**
- `lib/reports/vat-revenue-accounts.ts:96-137` (`fetchDynamicRuta05Accounts`): currently requires `default_vat_rate` already set; add BAS-convention fallback resolution
- `lib/reports/vat-declaration.ts:92-95, 407-429, 518` (`ACCOUNT_RUTA`, `rutorFromTotals`): whitelist + comment referencing the incomplete #1261 fix
- `lib/reports/vat-declaration-checks.ts:260-274`: `OUTPUT_VAT_WITHOUT_SALES_BASE` check and its hint text
- `extensions/general/mcp-server/server.ts:6049-6062` (`gnubok_create_account`): derive `default_vat_rate` from BAS catalog/number convention instead of leaving it undefined when not passed
- `lib/bookkeeping/bas-reference.ts:16-25` (`BASReferenceAccount`): consider adding a VAT-rate field or convention-based lookup helper

**Related files:**
- `lib/reports/__tests__/vat-declaration.test.ts:960-972`: existing test already covers 3011 with `default_vat_rate: 0.25` set explicitly — add a case for NULL/unset rate to catch regressions
- `lib/import/account-sync.ts:178`, `lib/import/sie-import.ts:1590`, `lib/import/opening-balance/parser.ts:151`, `lib/import/opening-balance/execute-helpers.ts:93`: other `getBASReference` call sites that create/sync accounts — check whether they have the same gap
- `app/api/bookkeeping/accounts/route.ts:106`: the REST account-creation route has the same `body.default_vat_rate ?? null` pattern

---
*Found via investigation of a real VAT close-check false positive.*

mattssonn:
## Summary

- include safely resolvable custom 30x1, 30x2, and 30x3 revenue accounts in ruta 05 when `default_vat_rate` is null and the account name explicitly confirms the matching 25, 12, or 6 percent VAT rate
- preserve every explicit VAT setting, including `0`, and avoid any historical-data backfill or account mutation
- update the close-check remediation text and add focused coverage across the core report, ruta sources API, and MCP completeness check

## Why this is safe

The fallback is report-time only. It requires both the existing BAS account-number convention and an explicit matching VAT rate in the account name. Ambiguous custom accounts remain excluded and continue to produce the close blocker rather than being silently guessed.

## Validation

- 165 focused regression tests passed across 6 suites
- all `lib/reports` tests passed: 760 passed, 2 skipped
- `npm run lint` passed with 0 errors and existing warnings only
- `npm run check:guards` passed
- full unit run reached 11,935 passed and 4 skipped; the only failure is the unrelated pre-existing syntax error in `scripts/checks/__tests__/format-currency-sek-label.test.ts`

Closes #1289

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved VAT declaration handling for ruta 05 accounts with missing VAT rates.
  * VAT rates can now be resolved from recognized account-number patterns and matching account labels for 25%, 12%, or 6% VAT.
  * Explicitly configured VAT rates continue to take precedence.
  * Reduced incorrect “output VAT without sales base” warnings for valid domestic sales mappings.
  * Updated guidance to help identify missing “Standard moms” account mappings.


mattssonn:
Follow-up on the advisory compliance questions:

- The `^30\d([123])$` scope is intentional. The issue is specifically about custom subaccounts such as 3011, and its documented convention is 30x1/30x2/30x3. Restricting the fallback to 3001/3002/3003 would leave the reported defect unfixed. A number alone is never accepted: the account label must contain exactly one matching explicit VAT rate, and any configured `default_vat_rate`, including 0, remains authoritative.
- Report-time inference is deliberate because no safe company-independent value can be persisted for a custom account. The calculation already uses mutable current chart metadata for explicitly configured dynamic accounts. This fallback follows the same report model while avoiding writes to account configuration, posted entries, or historical data. Renaming the account can change a newly generated report just as changing its explicit Standard moms can today; this patch does not mutate or backfill either field.

Given those constraints, the advisory points do not call for a code change in this PR.

PR #1322, opened 2026-07-31
fix: return has_more from paginated MCP list tools

bjornbergenheim:
paginatedSchema() marks has_more as required, but two handlers never returned it, so every call failed MCP output validation with "data must have required property 'has_more'". Both tools were unusable rather than merely returning an incomplete envelope.

Fix the handlers rather than relaxing the schema: the other four paginatedSchema tools already honour the contract.

Neither tool accepts an offset input, so there is no next_offset to hand back. has_more only tells the caller to narrow the filter, which is why the schema requires has_more but leaves next_offset optional.

Add a table-driven test pinning the envelope for both tools.

# What and why

`paginatedSchema()` in `extensions/general/mcp-server/server.ts` marks
`has_more` as required, but two tool handlers never returned it:
 
- `gnubok_list_invoices`
- `gnubok_list_recurring_schedules`
Every call therefore failed MCP output validation before the result reached the
client:
 
```
MCP error -32602: Structured content does not match the tool's output schema:
data must have required property 'has_more'
```
 
Both tools were effectively dead over MCP, not merely returning an incomplete
envelope. Found on a self-hosted instance while trying to list invoices after
sending the first one.
 
I fixed the two handlers rather than relaxing the schema, since the other four
`paginatedSchema` tools already return `has_more`. Neither of these two accepts
an `offset` input, so there is no `next_offset` to hand back: `has_more` only
signals that the caller should narrow the filter. That matches the schema, which
requires `has_more` but leaves `next_offset` optional.
 
Adds `extensions/general/mcp-server/__tests__/paginated-list-has-more.test.ts`,
a table-driven test covering both tools: the schema contract, `has_more` false
when the page holds every match, `has_more` true when matches exceed the page,
the empty result, the null-count fallback, and the database error path.
 
Also appends the customary `DECISIONS.md` line per `CLAUDE.md`.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Invoice and recurring-schedule list results now include accurate pagination status through the `has_more` field.
  * Pagination totals and counts are handled consistently, including empty results and unavailable total counts.
  * Database errors continue to be surfaced correctly.

* **Tests**
  * Added coverage for full, partial, empty, and error scenarios across both list tools.


bjornbergenheim:
both coderabit comments fixed, history rebased on current main and merged to one commit.

mattssonn:
Closing as superseded by #1327, merged in commit 8a2498a9873d734023ca3599e630bcef9f957e6d. Current main contains this PR's required has_more behavior and test coverage, and extends it with real offset/next_offset continuation, deterministic tie-break ordering, and lookahead handling when the exact count is unavailable. The thread-aware review audit is clean, and there are no remaining changes in this PR that are absent from main.

PR #1327, opened 2026-08-01
fix(mcp): add true continuation to invoice list tools

mattssonn:
## Summary

- add optional nonnegative integer `offset` continuation to `gnubok_list_invoices` and `gnubok_list_recurring_schedules`
- return the full pagination envelope with `has_more` and `next_offset`
- use stable descending `id` tie-breakers so equal invoice dates or creation timestamps do not reshuffle between page requests
- use a one-row lookahead to preserve continuation when Supabase omits an exact count
- add shared contract, execution, edge-case, and input-normalization tests for both tools

## Why

Both tools advertised the shared paginated output schema but did not return its required `has_more` field. More importantly, they accepted only `limit`, so rows beyond the first page were unreachable. The additive `offset` input makes page continuation possible while preserving the existing default first-page behavior and response fields.

## Compatibility

- existing callers that omit `offset` still receive the first 50 rows
- existing inputs and response fields are preserved
- `offset`, `has_more`, and `next_offset` are additive
- offsets are schema-validated as nonnegative integers and defensively normalized for direct execution
- returned page size remains capped at the requested limit even when one lookahead row is fetched internally
- the existing gnubok wire-format names and accounted aliases continue to use the same tool definitions

## Supersedes

Supersedes #1322, the older narrow `has_more`-only PR. This PR includes that required-schema repair and adds true page continuation, so the two PRs should not both be merged unchanged.

## Verification

- focused continuation tests: 14/14 passed
- MCP server tests: 751 passed; one unrelated timeout passed 4/4 when rerun in isolation
- targeted lint after the final review fixes: passed with 0 errors and existing warnings
- `npm run lint` before the final review fixes: passed with 0 errors and existing repository warnings
- `git diff --check`: passed
- full unit run: 12,091 tests passed and 4 skipped; one unrelated pre-existing suite, `scripts/checks/__tests__/format-currency-sek-label.test.ts`, still fails during import with `SyntaxError: Invalid or unexpected token`

No migrations, API routes, UI strings, or database operations are included.

Ready for final checks and authorized merge.

### DIFF
--- a/DECISIONS.md
+++ b/DECISIONS.md
@@ -729,3 +729,5 @@ One line per decision: `[YYYY-MM-DD] <decision>: <why>`. Appended by agents and 
 
 [2026-08-01] Cloud Backup OAuth redirect URIs resolve from NEXT_PUBLIC_APP_URL, with request origin only as a self-hosted fallback: Google and Dropbox require pre-registered callbacks, so deriving them from an old alias or preview host can reject the flow before consent; one resolver keeps authorization, exchange, revoke, and sync origins consistent.
 [2026-08-01] Do not claim a fixed Supabase Docker revision: Accounted does not maintain a tested Supabase stack pin, so the self-hosting guide requires one complete immutable upstream tag or commit instead of naming an unverified version.
+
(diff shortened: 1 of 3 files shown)


mattssonn:
Follow-through note on the low-severity offset-cap advisory: I am not adding a fixed maximum in this focused PR. The existing paginated MCP tools use the same nonnegative, uncapped offset contract, and a fixed cap would make rows beyond that cap unreachable, recreating the continuation bug this PR fixes. Per-request `limit` remains capped at 100 and MCP API-key rate limiting still applies. If deeper offset cost needs additional controls, it should be designed consistently across every paginated MCP tool, preferably as a separate cursor or pagination policy change rather than a unique limit on these two tools.

PR #1339, opened 2026-08-02
feat(invoice-inbox): surface WhatsApp chat context in booking flows

jakobwennberg-oss:
PR5a of the WhatsApp intake track. SIBLING of the conversation-layer PR: both are based on #1338's branch (`feat/whatsapp-inbox-extension`), file-scoped to avoid overlap (this PR touches nothing under `extensions/general/whatsapp-inbox/**`, `vercel.json`, or `docker/crontab*`). Retarget rules apply: if the base branch merges or is retargeted, retarget this PR the same way before merging.

## What it does

The WhatsApp bot writes verified human answers to `invoice_inbox_items.channel_context` (type `InboxChannelContext`). This PR makes the in-app booking flows READ that context:

- **Core renderer** `lib/documents/channel-context-notes.ts`: `renderChannelContextNotes()` renders the context as one deterministic compact Swedish line for the notes path, e.g. `Representation: Anna Berg (Volvo), Jakob W · Syfte: uppföljning av avtal`. Capped at 220 chars (the verifikat description caps at 500 in `transaction-entries.ts`) by dropping WHOLE participant names (`… och N till`), never mid-name. Precedence: representation, then user_note; caption only when nothing else exists. Core lib, no `@/extensions` imports.
- **FieldsRail "Från WhatsApp" block** (`InvoiceInboxWorkspace.tsx`): caption (Meddelande), deltagare, syfte, anteckning rows, mirroring the email metadata block; plus an ochre `AttnLine` when `pending_question.status === 'moved_to_app'` (chat question expired unanswered, complete it here).
- **Notes threading into booking**:
  - *Book-direct*: server defaults `notes` to the rendered string when the request sends none (caller-supplied value always wins); `BookDirectlyDialog` prefills its notes input with the same string so the user reviews/edits what will land.
  - *Convert*: the supplier invoice `notes` column gets the same default (schema already has `notes`).
  - *Bulk-book* (`bulkBookMatchedInboxItems`): per-item rendered context is JOINED with the shared batch note (`' · '`), because bulk has no per-item notes review and caller-wins would drop the only copy of the representation documentation (see DECISIONS.md).
- **Inbox list**: whatsapp rows get a `MessageCircle` icon and a quiet `Fråga obesvarad` badge for moved_to_app items. No worklist change: unresolved whatsapp items are already counted by `countInboxDocuments` (no double-count).
- The list route select now includes `channel_context`; the single-item route already used `select('*')`.
- sv/en strings for all 7 new `inbox_workspace` keys.

## Verification

- `npx vitest run extensions/general/invoice-inbox lib/documents lib/transactions`: 28 files, 421 tests passed
- Full `npm test`: 963 files passed (2 skipped), 12232 tests passed (4 skipped)
- `npm run lint`: 0 errors
- `npx tsc --noEmit`: exactly 405 errors, all pre-existing, none in touched files
- Renderer suite mutation-checked: truncation deliberately broken, 2 tests failed, restored
- sv/en key parity script-verified for every new key

New tests: renderer suite (representation full, whole-name overflow truncation, user_note only, caption only/fallback, empty to null, self participant, 220 cap), book-direct server default + caller-wins + no-context, convert notes default + caller-wins, bulk-book threading (rendered alone, joined with shared note, passthrough without context).


jakobwennberg-oss:
Three adversarially-verified review findings fixed on this branch (2 commits: `014682fe`, `3eed9494`).

**1. [high] Cleared notes were silently re-applied server-side.** The dialog prefills the rendered chat note; it sent `notes.trim() || undefined`, and both routes defaulted from `channel_context` on any falsy value, so a user who read the prefill, disagreed and deleted it got it written back onto a posted verifikat, removable only through a formal rättelse. Both comments claiming "an edited value always wins" were false for exactly that edit.

Fix: **presence of the field decides, not truthiness.** `BookDirectlyDialog` now always submits `notes` (empty string included); book-direct and convert default from `channel_context` only when the field is *absent* from the request (MCP, older clients). `BookInboxItemDirectlySchema.notes` stays `z.string().max(2000).optional()` on purpose, with a comment at the schema explaining that `.default('')` or a `min(1)` would collapse "cleared" and "no opinion" back into one value. Both route comments now describe the real rule. Convert got the same treatment; note that its UI (`NewSupplierInvoiceForm`) still neither prefills nor previews the chat context, so nothing regressed there: it sends `notes: data.notes || undefined`, i.e. absent when empty, and the default keeps working for it.

**2. [medium] Unreviewed photo caption burned into verifikat text.** `renderChannelContextNotes` fell back to the raw caption, and `bulkBookMatchedInboxItems` appended the result per item with no per-item notes field (the MCP staged-approval preview carries no per-item PII either), so chat text nobody had seen reached an immutable description.

Fix: the renderer takes `{ includeCaption }` and leaves the caption out **by default**. Representation answers and `user_note` are replies to a question the bot asked; the caption is not. Only the Bokför direkt prefill opts in, where the user reads the string in an editable field (and, per fix 1, can now really delete it). Bulk-book and both server-side defaults thread the answered parts only.

**3. [low] Full participant list was unreachable in the archive.** The verifikat line caps at 220 chars and drops whole names (`… och N till`), while the complete deltagare list lives only in `invoice_inbox_items.channel_context`, and that table sat in `ARCHIVE_EXCLUDED_TABLES` under a rationale that predates the column. A company leaving with the full-archive export as its BFL 7-year record kept incomplete representation documentation.

Fix: the archive route, as suggested. `invoice_inbox_items` is now dumped as a **column projection** via the new `MasterDataTableSpec.columns`: underlag provenance (document, matched transaction, created verifikat / leverantörsfaktura) plus `channel_context`, so each answer is tied to what was booked from it, while inbox workflow state (email bodies, OCR output, error messages) stays out of the archive. Documents themselves remain in `dokument/`. Documented in DECISIONS.md, along with the two decisions above (the first supersedes the 2026-08-02 "server default fills empty/absent notes" line).

**Verification:** `npx vitest run extensions/general/invoice-inbox lib/documents lib/transactions lib/reports` green (1207 tests); full `npm test` green (12242 passed, 963 files); `npm run lint` 0 errors; `npx tsc --noEmit` 405 errors, unchanged from the pre-change baseline; `npm run check:guards` passed. Each fix ships a test that fails without it, and the book-direct cleared-notes tests were mutation-checked (restoring the truthiness fallback fails them, restoring the fix passes). No new UI strings, so sv/en parity is untouched.

PR #1364, opened 2026-08-03
feat(invariants): shared format contracts + upgrade-path CI

jakobwennberg-oss:
Two independent hardening changes that each close a failure class our history says recurs. No user-visible behaviour change except one improved validation message.

## 1. `lib/invariants/`: one home for shared format rules

The same format rules were written out independently across the codebase, and where they disagreed the disagreement was invisible until something failed at a deadline.

The worst case, now fixed: four Skatteverket- and Bolagsverket-bound export paths each had their own idea of a valid organisationsnummer.

| Path | Old rule | Failure |
|---|---|---|
| `lib/skatteverket/format.ts` | strip `-` only | threw on any input containing a space |
| `lib/salary/ku/ku10-generator.ts` | `replace('-', '')` | first hyphen only, spaces survived |
| `lib/salary/agi/xml-generator.ts` | strip non-digits | stray letters passed the length check |
| `lib/bokslut/ixbrl/validate/rules.ts` | `/^\d{6}-?\d{4}$/` | rejected the 12-digit form, no check digit |

A company stored with a space, or in 12-digit form, could file AGI all year and then fail at the årsredovisning deadline with a message that did not say why. There is now a cross-path test that feeds the same org number through all four and fails if any one drifts.

`lib/invariants/` owns account number, ISO date, four-digit fiscal year and org number, each with the rationale recorded next to the rule. `normalizeOrgNumber` moved here from `lib/company-lookup/` and `isSaneDateString` from `lib/utils.ts`; both old paths re-export, so no caller changed. `lib/api/schemas.ts` builds its primitives on the module, so ~100 schemas inherit any future correction.

### Three judgment calls worth reviewing

**The årsredovisning check-digit verdict is `warn`, not `error`.** A wrong Luhn digit is almost certainly a typo worth surfacing, but whether every org number Bolagsverket accepts satisfies Luhn is a Swedish domain question not covered by the `swedish-payroll` skill. An `error` there blocks Skicka in, and we do not block a statutory filing on an unverified assumption.

**KU10's 12-digit passthrough is pinned by a test, not changed.** KU10 emits 12 digits when the org number is stored in 12-digit form. That is pre-existing (the old `replace('-','')` did the same). Whether the KU10 schema wants 10 or 12 is an open domain question, so it is pinned rather than silently altered. **Needs a domain answer.**

**ROT/RUT `brf_org_number` schemas are untouched.** They use a different regex for a documented reason (Skatteverket's schema wants the 12-digit sekelsiffra form there). Genuinely a different contract, so it stays local per the module's own scope rule.

Guard 8 (`hand-rolled-invariant`) tracks the remaining 114 inline copies as a ratchet that may only go down, same mechanism as the existing `roundOre` guard. Verified that it fails on a new copy.

## 2. `pg-upgrade` CI job: migrations tested against real data

`pg-real` applies all 548 migrations to an **empty** database. Empty means zero rows, so a migration that adds a NOT NULL, adds a CHECK, creates a unique index or backfills passes trivially and can still fail on production. CI proved a fresh install works; nothing proved an existing install upgrades.

The new job applies the schema at the merge base, seeds a small real company (three posted verifikat, balanced lines, one öre-level amount), applies **only** the migrations the PR adds, then asserts the data survived: entries still posted, lines intact, ledger still balances, öre unchanged, voucher numbers sequential. A PR with no migration no-ops.

Verified locally against `supabase/postgres:15.8.1.060` rather than assumed, with three deliberately bad migrations:

| Bad migration | Empty DB (current job) | Seeded DB (new job) |
|---|---|---|
| Rescale money on posted lines | would pass | ERROR (immutability trigger) |
| CHECK violating the öre row | **exit 0, ships** | exit 3, fails |
| NOT NULL on a populated column | **exit 0, ships** | exit 3, fails |

Base migrations are read out of the merge-base git tree, not the working tree, so a PR that edits an already-shipped migration still gets the original applied and the edit surfaces as a failure here.

## Verification

- `npm test`: 12,180 passed, 2 skipped
- `npm run lint`: 0 errors
- `tsc`: 505 pre-existing errors, all in test files, none in anything this PR touches

This PR adds no migration, so `pg-upgrade` should no-op on itself and `check-pg-test-coverage` has nothing to gate.

## Summary by CodeRabbit

- **Improvements**
  - Standardized validation for organization numbers, account numbers, fiscal years, and dates across settings and exports.
  - Organization numbers now support consistent formatting and clearer invalid check-digit handling.
  - Improved normalization in salary and tax-reporting files.

- **Bug Fixes**
  - Malformed organization-number text is no longer silently accepted.
  - Database upgrades preserve existing records, balances, numbering, and financial precision.

- **Tests**
  - Expanded coverage for validation, export formats, and database migration upgrades.


jakobwennberg-oss:
## Response to the Swedish compliance review

All four points checked against the tree rather than reasoned about. One of them corrects the finding's framing; one produced a change.

### 1. iXBRL check-digit severity: the framing is off, this is not a downgrade

The finding reads it as "downgraded from error to warn", which implies an error-level check-digit gate existed. It did not. On `main`, `lib/bokslut/ixbrl/validate/rules.ts` tested `/^\d{6}-?\d{4}$/` only, and `grep -ci luhn` on that file returns **0**. A Luhn-invalid but shape-valid org number such as `556999-9999` produced **no issue at all**, and would have gone to Bolagsverket unflagged.

Versus `main`, this PR **adds** a check-digit warning where there was none. The net change is strictly more protective, not less. What is `error` on both sides is the shape rule, and that is unchanged.

Two things did get stricter versus `main`: the 12-digit form is no longer falsely rejected, and the same rule now applies across all four export paths.

The recommendation itself stands and is already recorded as an open question in `DECISIONS.md`: whether Bolagsverket's accepted set is fully Luhn-conformant needs a primary source before `error` is justified. It is not treated as settled.

### 2. KU10 12-digit passthrough: agreed, and deliberately pinned

Same conclusion the PR reached. Not covered by the `swedish-payroll` skill (checked: no reference file addresses the KU10 identity digit format), so it is pinned by a test rather than changed on a guess. Pre-existing behaviour, unchanged by this PR: the old `replace('-', '')` did the same. Flagged for a primary Skatteverket source before anyone touches it.

### 3. `seed.sql` raw posted-entry inserts: change made

Fair point, and the risk is that it reads like a sanctioned pattern to whoever finds it next. `57e20413` puts the reasoning in the file: CI-only, safe because the database is thrown away with the job (no voucher sequence to keep gapless, no retention obligation under BFL 5 kap 6-7 §), and an explicit instruction not to reuse it for a seed script, migration or repair script, with a pointer back to Hard Rule 2.

It is not reachable from any production path: nothing outside `.github/workflows/test-pg-real.yml` references `tests/pg/upgrade/`.

### 4. Intake enforcement behind the permissive exporters: verified, holds

The concern is right to raise, and the answer is that the identifier feeding all four exporters is enforced.

The company's own org number, which is what KU10, AGI, SRU redovisare and iXBRL all read, is written through `lib/company/actions.ts:113`, which calls `normalizeOrgNumber` and refuses malformed input rather than storing it. `app/api/company/check-org-number/route.ts:43` normalizes on the same rule.

Customer and supplier org numbers are separately unvalidated (`org_number: z.string().optional()` at `lib/api/schemas.ts:834, 865, 902, 1691`). Those are third-party identifiers on invoices and ROT/RUT, not the filing identity, so they do not reach these exporters. Tightening them is a real follow-up but out of scope here, and it would want the ROT/RUT `brf_org_number` rule considered alongside it since that one is deliberately different.

jakobwennberg-oss:
Confirming the reachability question on the re-run finding, verified rather than asserted.

`tests/pg/upgrade/seed.sql` has exactly two references in the entire repository, both inside the CI job that owns it:

```
.github/workflows/test-pg-real.yml:119   psql ... -f tests/pg/upgrade/seed.sql
.github/workflows/test-pg-real.yml:132   psql ... -f tests/pg/upgrade/assert.sql
```

- No npm script references it (`grep upgrade package.json` is empty), so there is no `npm run` path to it.
- Vitest cannot pick it up: both projects glob `**/*.test.ts` and `**/*.pg.test.ts`, and these are `.sql`.
- It is not a migration and lives outside `supabase/migrations/`, so no migration runner reaches it.
- The job that runs it points `DATABASE_URL` at the ephemeral `supabase/postgres` service container, which is destroyed with the job.

The only way to execute it against a real database is for a human to type the `psql` invocation by hand at a production connection string, which the header now explicitly tells them not to do.

jakobwennberg-oss:
## Correction to my earlier reply, and the precise answer

I said this PR was "strictly more protective, not less" versus `main`. That was too broad, and the bot's sharper re-run wording is right that one input class does become non-blocking. Correcting the record with the actual behaviour, tested rather than reasoned about:

| Input | `main` (`/^\d{6}-?\d{4}$/`) | This PR | Net |
|---|---|---|---|
| 10-digit, good check digit | pass | pass | unchanged |
| 10-digit, **bad** check digit | **pass, no issue at all** | warn | **more protective** |
| Hyphenated, good | pass | pass | unchanged |
| 12-digit, good check digit | **ERROR, blocks Skicka in** | pass | **bug fixed** |
| Spaced, good | **ERROR, blocks Skicka in** | pass | **bug fixed** |
| 12-digit, **bad** check digit | **ERROR, blocks** | warn | **less blocking** |

So: the last row is real. A 12-digit org number with a bad check digit used to block and now warns.

### Why it still stays a warning

The old block on that row was incidental, not a check-digit gate. `main` blocked **every** 12-digit org number regardless of its check digit, including valid ones, which is the false-rejection bug this PR fixes. Removing a wrong rule necessarily removes the blocking it did by accident. Preserving it only for the bad-check-digit subset would mean making severity depend on whether the user typed 10 or 12 digits, which is not a principled rule and would be a strange thing to explain to a user.

Two things bound the residual risk:

1. **The class is largely unreachable through the app.** A company's org number is written via `lib/company/actions.ts:113`, which runs `normalizeOrgNumber` (Luhn-checked) and refuses malformed input rather than storing it. A stored bad check digit in any format essentially requires a legacy row or a direct database write.
2. **It is no longer silent.** For the far more common 10-digit case, `main` gave the user nothing at all; they would have learned about it from Bolagsverket. Now they get a message naming the check digit.

The recommendation is unchanged and stands as an open item in `DECISIONS.md`: confirming against a Bolagsverket primary source whether every accepted org number satisfies mod-10 is what would justify promoting this to `error` across all formats. That is the right fix, and it is a domain question rather than a code one.

PR #1374, opened 2026-08-03
fix(salary): book net deductions as settlement lines so salary entries balance

### LINKED ISSUE #316: [payroll] Net deductions never booked — produces unbalanced journal entries
**Severity**: critical
**File**: `lib/salary/salary-entries.ts:147-154`

### Description
In `createSalaryEntry`, the line-item loop skips every item where `is_net_deduction` is true (line 148). The `baseRemainder` logic then books the full `gross_salary` to 7xxx. But the credit side only books 2710 (tax) + 1930 (net pay). For an employee with gross 40,000, tax 12,000, net_deduction_union 500, net_salary 27,500: debit = 40,000, credit = 12,000 + 27,500 = 39,500 — off by 500. The missing 500 should credit a payable (e.g. 2890 Övriga kortfristiga skulder) for the union. The `check_journal_entry_balance()` trigger will reject this, so either the entry fails to post or the test never covered this scenario. No tests exist for `salary-entries.ts`, so this has not been caught.

### Suggested fix
Build a second mapping for net deductions to credit-side liability accounts (2890/2899 for union, 1680 for advances, 2690 for benefit payments) and push corresponding credit lines. Add tests for `createSalaryRunEntries` covering net deductions. Pair with the related account-mapping issue (`lib/salary/account-mapping.ts:38-41` currently maps net-deduction items to 7210 instead of liability accounts).

---

mattssonn:
## Problem

`createSalaryEntry` skipped every line item with `is_net_deduction`, so the debit side booked full gross salary while the credit side only booked 2710 (tax) and 1930 (net pay). Any run with a net deduction (union fee, salary advance repayment, benefit co-payment) produced an entry unbalanced by exactly the deducted amount, which `check_journal_entry_balance()` rejects: the salary run could not be booked at all.

## Fix

- Net deduction items now book as aggregated settlement lines on the credit side (negative payslip amount = credit; a positive correction books as a debit repayment). Like the 2710/1930 legs they stay undimensioned.
- Default account mapping in `account-mapping.ts` moves off 7210: `net_deduction_advance` -> 1613 (credits back the advance receivable), `net_deduction_union` -> 2794 Fackföreningsavgifter, `net_deduction_benefit_payment` -> 7385 (reduces the benefit cost), `net_deduction_other` -> 2799 Övriga löneavdrag. An explicit `account_number` on the line still overrides.
- Keeping the mapping in `LINE_ITEM_ACCOUNTS` (rather than a parallel table) means payslip line creation, the booking preview route and the voucher builder all resolve the same account.
- The salary run preview route mirrors the new lines so the preview matches what gets booked.

## Tests

Two new tests in `lib/salary/__tests__/salary-entries.test.ts`: union-fee deduction keeps the entry balanced on 2794, and a five-deduction run covering all four default mappings plus an explicit 2890 override. Full `lib/salary` + `app/api/salary` suites pass (679 tests), lint 0 errors, check:guards green.

Fixes #316


mattssonn:
Compliance review findings addressed in 4b8f4774:

**Finding 1 (account verification):** all three defaults verified against the repo's BAS 2026 reference data (lib/bookkeeping/bas-data/), which is the same catalog ensureSalaryAccountsExist auto-creates from:
- 1613 = Övriga förskott (asset, group 16, SRU 7212): crediting the advance receivable on repayment is the standard treatment for löneförskott.
- 2794 = Fackföreningsavgifter (liability, group 27, SRU 7231): exact-purpose account for withheld union fees.
- 2799 = Övriga löneavdrag (liability, group 27, SRU 7231): exact-purpose account for other net deductions.

Balance-sheet classification and SRU codes come from the same reference rows, so balansräkning and INK2R exports classify these correctly.

**Finding 2 (7385 heterogeneous benefits):** label corrected to the BAS name Kostnader för fri bil. The mapping keeps 7385 as the default because there is a single net_deduction_benefit_payment item type with no benefit-kind field; fri bil is the dominant co-payment case and other benefit kinds set account_number per line (7381/7382/7388/7389/7699). Recorded in DECISIONS.md; a per-benefit split needs a schema change and its own issue.

**Finding 3 (repayment direction):** added a test booking a positive net deduction as a debit repayment on 2794 with the entry balanced.

@coderabbitai review

PR #1382, opened 2026-08-03
fix(salary): book lonevaxling pension provision from frozen run snapshot

### LINKED ISSUE #317: [payroll] Löneväxling pension entry never booked
**Severity**: critical
**File**: `app/api/salary/runs/[id]/book/route.ts:68-86` and `lib/salary/salary-entries.ts:104-114`

### Description
`createSalaryRunEntries` creates a pension entry only when `pension_contribution > 0` on the employees it receives (line 110). But the book route never populates `pension_contribution` or `pension_slp` — those fields are completely absent from the `.map()` at route.ts:68-84. As a result, even if a user configures a `gross_deduction_pension` line item, the 1.058 multiplier is calculated by `calculateLoneVaxling` (never called from the book flow anyway) but no 7410/2740/7533/2514 lines are ever emitted. The pension deduction is silently lost from the books — 7218/7210 side shows reduced salary, but the pension obligation to the försäkringsbolag is never recorded.

### Suggested fix
In `book/route.ts`, for each employee with `gross_deduction_pension` line items, compute `pension_contribution = abs(gross_deduction_total) * 1.058` and `pension_slp = pension_contribution * config.slpRate`, then pass to `createSalaryRunEntries`. Or compute these during `calculate` and store on `salary_run_employees` for re-use at booking.

---

mattssonn:
## Summary
- The salary booking flow never populated `pension_contribution` / `pension_slp`, so an employee with a `gross_deduction_pension` line item got a reduced salary entry while the pension obligation itself (7410/2740) and the SLP provision (7533/2514) were never posted. The deduction was silently lost from the books.
- Fix derives both amounts at the shared `createSalaryRunEntries` boundary from the run's frozen `calculation_params.slpRate` snapshot, so all three booking paths (dashboard route via `bookPaidSalaryRun`, MCP `book_salary_run` pending operation, v1 REST route) emit the pension verifikat consistently.
- The 1.058 löneväxling factor and öre rounding are shared with the planning calculator via a new `calculateLoneVaxlingPensionProvision` helper (uses `roundOre`), so planning preview and posted amounts can never drift.
- Explicit `pension_contribution`/`pension_slp` values passed by callers are preserved; derivation only fills the gap when they are absent. A pension deduction on a run without a rate snapshot fails loudly instead of booking without SLP.
- v1 dry-run preview now lists 'pension provision and SLP' when a pension deduction is present.

## Testing
- `npx vitest run lib/salary app/api/v1/companies/[companyId]/salary-runs` (574 tests green), including new tests for derivation from line items, snapshot passthrough in `bookPaidSalaryRun` and the v1 book route.
- `npm run lint` (0 errors), `npm run check:guards` green.

Fixes #317


mattssonn:
Verified both advisory compliance findings before merge:

1. Missing-snapshot hard fail: calculation_params (including slpRate) has been written by the calculate step since the original salary module (#245, migration 20260414120000_salary_module.sql), and booking requires the calculated-review-approved-paid lifecycle. Every bookable run therefore has the snapshot; the throw guards a corrupt-state invariant, it cannot hit legacy runs.

2. Frozen rate vs statutory rate at booking date: deliberate and documented in DECISIONS.md 2026-08-03. A run held across an SLP rate change needs recalculation anyway (avgifter and tax tables change too), and recalculating refreshes the snapshot.

No code change needed for either.

PR #1396, opened 2026-08-03
fix(packs): representation template named itself avdragsgill while booking 6072

jakobwennberg-oss:
Raised by the Swedish compliance bot on #1395, chased to ground.

## The defect

The template called itself **"Representation (avdragsgill, 25% moms)"** and labelled its cost line *"Representation avdragsgill"*, while booking to **`6072`**, which BAS 2026 defines as **"Representation, ej avdragsgill"**.

The `swedish-vat` skill lists that exact confusion under *Representation errors*:

> Not separating avdragsgill/ej avdragsgill on correct accounts (6071 vs 6072).

The account was right and the words were wrong. Meal representation stopped being income-tax deductible in 2017; only enklare förtäring (max 60 kr/person) survives, on `6071`.

Worse, our own static `representation_external` in `booking-templates.ts` uses `6071` at `12%`, so the two template systems contradicted each other and both are reachable by users.

## Correcting the record

The compliance bot suggested the seeded `ratio: 0.8` encoded a phantom "80% deductible" rule, and I repeated that before checking. **It did not.**

```
1 / 1.25 = 0.8
```

It was the net-of-VAT fraction pairing with the 25% rate: cost 80 + VAT 20 = 100. The arithmetic balanced and was never wrong. Only the naming and the missing cap were. The ratio is now commented in the file so the next reader doesn't repeat the misreading.

## The fix

| | Before | After |
|---|---|---|
| Name | Representation (avdragsgill, 25% moms) | Representation, måltid (ej avdragsgill) |
| Cost line | `6072` labelled *"avdragsgill"* | `6072` labelled *"ej avdragsgill"* |
| VAT | 25% | 12% (restaurang/catering, matching our static template) |
| Cost ratio | 0.8 (= 1/1.25) | 0.892857 (= 1/1.12) |
| Cap guidance | none | `legal_note` |

The `legal_note` carries the **300 kr per person** VAT cap, which this format **cannot compute**: a template sees a total and never a participant count. It says so plainly and tells the user to adjust the VAT line, rather than pretending the number is final.

## Slug deliberately unchanged

It still reads `representation-avdragsgill-25-moms`. A slug is an identifier, not a label (`packs/README.md`). Renaming it would retire this template and insert a new one, breaking every company's `booking_template_usage` history. The name is what users read, and that is now correct. A comment in the file explains why.

## The drift guard did its job, and exposed its own flaw

Renaming the template failed `port-is-lossless.test.ts`, correctly. That surfaced that the test joined packs to seeded rows **by name** — the same fragility that made `pack_slug` the sync's key in #1390. `INTENTIONAL_DIVERGENCES` now records each entry's `seededName` explicitly, so a rename stays checkable.

## Verification

- `npm test`: 12,338 passed, 2 skipped
- `npm run lint`: 0 errors
- `npm run validate:packs`: 26 valid, 0 quarantined

## Open question, not resolved here

The 12% rate is the restaurang/catering rate and matches our existing static template. The `swedish-vat` skill notes food VAT dropped 12% → 6% in April 2026, but does not settle whether that applies to restaurant meals or only to livsmedel. I did not guess: 12% keeps the two template systems consistent, and the `legal_note` tells the user the rate follows the underlag. **Worth a domain check.**

## Summary by CodeRabbit

- **Updates**
  - Updated the representation template to clearly describe non-deductible meal representation.
  - Refined the associated business-line label and calculation for 12% VAT.
  - Updated the VAT line to apply the 12% rate with clearer wording.
  - The template’s existing identifier and ordering remain unchanged.

### DIFF
--- a/lib/packs/__tests__/port-is-lossless.test.ts
+++ b/lib/packs/__tests__/port-is-lossless.test.ts
@@ -23,18 +23,42 @@ import seeded from './fixtures/seeded-system-templates.json'
  * Every entry is a defect found by the pack validator and fixed with a domain
  * source cited in the commit.
  */
-const INTENTIONAL_DIVERGENCES: Record<string, string> = {
-  loneutbetalning:
-    'Seeded version debited 2710 @0.3 + 2920 @0.12 + 7010 @1.0 against a single 1.0 credit, ' +
+interface Divergence {
+  /**
+   * The name this template carried in the seed. Needed because a fix may rename
+   * it, and the seeded fixture predates pack_slug so name is the only join key
+   * available there. Exactly the fragility that made pack_slug the sync's key.
+   */
+  seededName: string
+  reason: string
+}
+
+const INTENTIONAL_DIVERGENCES: Record<string, Divergence> = {
+  loneutbetalning: {
+    seededName: 'Löneutbetalning',
+    reason:
+      'Seeded version debited 2710 @0.3 + 2920 @0.12 + 7010 @1.0 against a single 1.0 credit, ' +
     'so it totalled 1.42x the amount and could never post. Rebuilt per the swedish-payroll ' +
     'skill: Debit 7010 gross, Credit 2710 tax, Credit 1930 net. The 2920 semesterlöneskuld ' +
     'line moved out because vacation accrual is its own verifikat (7290/2920).',
-  'periodiseringsfond-avsattning-ab':
-    'Seeded version used account 2113, i.e. the fund for tax year 2013 under the pre-2020 ' +
+  },
+  'periodiseringsfond-avsattning-ab': {
+    seededName: 'Periodiseringsfond avsättning (AB)',
+    reason:
+      'Seeded version used account 2113, i.e. the fund for tax year 2013 under the pre-2020 ' +
     'year-tagged block. Those funds had to be reversed years ago and the account is not in ' +
     'BAS 2026, so the template could not resolve. Now uses 2110 Periodiseringsfonder, which ' +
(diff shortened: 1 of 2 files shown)


jakobwennberg-oss:
## Open question closed: 12% is correct

I flagged the rate as needing a domain check. The compliance re-review pointed at the source, and I verified it:

> From 1 April 2026, livsmedel drops from 12% to 6% (tillfälligt, Prop. 2025/26:55). **Restaurang/servering stays at 12%.**
> — `swedish-accounting-compliance/SKILL.md:68`

and further:

> Take-away/avhämtning = 6%, servering/förtäring på plats = 12%
> — `references/skatteverket.md:34`

Representation meals are servering, so 12% is right and the April 2026 change does not reach them. No further action on the rate.

## On the 300 kr cap being documented rather than enforced

Agreed, and it is why this is a `legal_note` rather than a computed line: the pack format sees a total and never a participant count, so it cannot compute the cap. That is a real format limitation, not an oversight.

The suggestion of a `participant_count` input that computes the capped base is the right shape for a fix, but it is a schema change plus a UI change, not something to fold into a correction PR. Worth its own issue.

What this PR does change: before it, the template was silently wrong in two ways (mislabelled, and 25% on a 12% supply). After it, the remaining gap is explicit and in front of the user.

## On historical postings

This is the finding worth acting on and I cannot settle it from here: it needs `booking_template_usage` on prod to say whether this template was ever actually applied, and to how many verifikationer.

Flagging rather than assuming. If it was used, entries booked at 25% on a 12% supply over-deducted input VAT, and per BFL 5 kap 5 § that is a rättelse question for closed periods, not something a template fix reaches.

PR #1397, opened 2026-08-03
fix(bookkeeping): three static templates named accounts that break or mis-book

jakobwennberg-oss:
You asked me to analyse across the program. I cross-checked **all 86 booking templates** (26 packs + 60 static) against BAS 2026 on four axes: account existence, label-vs-account contradiction, the `ratio` ↔ `vat_rate` coupling, and disagreement between the two template systems.

**The pack catalogue is clean.** The static registry had three defects, none visible to any existing check.

## Two templates that could never work

`lib/bookkeeping/account-backfill.ts` seeds a missing account into a company's chart **only if it appears in `BAS_REFERENCE`**. An account that doesn't is unreachable, so the template fails with `AccountsNotInChartError` every time, for every company.

| Template | Was | Now | Why |
|---|---|---|---|
| `vehicle_parking` | `5614` | `5619` | The BAS 561x run skips 5614 entirely. 5619 is the sibling *"Övriga kostnader för personbilar och mc"*. |
| `it_cloud_hosting` | `5421` | `5420` | BAS has no 5421. Both sibling SaaS templates already use `5420 Programvaror`. |

## One that silently mis-booked, which is worse

| Template | Was | Now |
|---|---|---|
| `travel_hotel` | `5820` **Hyrbilskostnader** | `5830` **Kost och logi** |

The *Hotell* template filed hotel nights under **car hire**. It balanced, the account existed, nothing warned. This is the #1321 shape again: the name says one thing, the booking does another.

An existing test asserted `5820`, so the bug was pinned as expected output. That assertion now reads `5830` and carries the reason, so nobody "fixes" it back.

## The guard

`template-accounts-exist.test.ts` makes the 5614 class unable to return:

- every account any static template names must resolve in BAS 2026
- account numbers must be strings
- a VAT-bearing purchase may not debit equity (class 2) or revenue (class 3)

That last check is deliberately narrow. My first version required a cost account (4-7) and immediately flagged `equipment_capital`, which debits `1250 Inventarier` and reclaims VAT — **correct**, that is how capex is booked. Class 1 is legitimate; only a sign error lands in 2 or 3.

Verified the guard fails on `5614` and passes on `5619`, rather than trusting it.

The pack catalogue already had this check via `scripts/validate-packs.ts`. The static registry had nothing, which is how it drifted.

## What the audit did NOT find

No label-vs-account contradictions remain in the packs (the representation one was #1396), the `ratio` ↔ `vat_rate` coupling holds everywhere, and no two templates disagree on the VAT rate for a shared debit account.

## Verification

- `npm test`: 12,350 passed, 2 skipped
- `npm run lint`: 0 errors
- `npm run validate:packs`: 26 valid, 0 quarantined

## Related concern I want to flag, not bury

In #1388 I **added account `2012`** to the BAS reference, on the strength of the `swedish-year-end-closing` skill using it twice. This audit showed **three** in-repo sources omit it: the lib reference, `dev_docs/BASKONTOPLAN.md`, and `seed_chart_of_accounts()`. Only the skill says it exists.

I acted on one source against three. The consequence is bounded (the account is only seeded when something references it, and only the EF preliminär-F-skatt template does), but it deserves a primary-source check against bas.se rather than my judgement. Details in the PR discussion.

### DIFF
--- a/lib/bookkeeping/__tests__/booking-templates.test.ts
+++ b/lib/bookkeeping/__tests__/booking-templates.test.ts
@@ -426,7 +426,10 @@ describe('buildMappingResultFromTemplate', () => {
     const tx = makeTransaction({ amount: -1120 })
     const result = buildMappingResultFromTemplate(template, tx, 'enskild_firma')
 
-    expect(result.debit_account).toBe('5820')
+    // 5830 Kost och logi, not 5820 Hyrbilskostnader. This assertion used to
+    // read 5820, which pinned a real bug: the Hotell template posted hotel
+    // nights into car hire. It balanced, so nothing ever complained.
+    expect(result.debit_account).toBe('5830')
     expect(result.vat_lines).toHaveLength(1)
     expect(result.vat_lines[0].account_number).toBe('2641')
     expect(result.vat_lines[0].debit_amount).toBe(120) // 1120 * 0.12 / 1.12 = 120
--- a/lib/bookkeeping/__tests__/template-accounts-exist.test.ts
+++ b/lib/bookkeeping/__tests__/template-accounts-exist.test.ts
@@ -1,0 +1,74 @@
+import { describe, it, expect } from 'vitest'
+import { BOOKING_TEMPLATES } from '@/lib/bookkeeping/booking-templates'
+import { getBASReference } from '@/lib/bookkeeping/bas-reference'
+
+/**
+ * Every account a booking template names must exist in BAS 2026.
+ *
+ * This is not pedantry: `lib/bookkeeping/account-backfill.ts` seeds a missing
+ * account into a company's chart on demand ONLY if it appears in
+ * BAS_REFERENCE. An account that does not is unreachable, so the template
+ * fails with AccountsNotInChartError every single time it is used, for every
+ * company. It is a template that can never work.
+ *
+ * Three shipped templates were in exactly that state until this test existed:
+ *   vehicle_parking    5614  (the BAS 561x run skips 5614)
+ *   it_cloud_hosting   5421  (BAS has 5420 Programvaror, no 5421)
+ *   preliminar-f-skatt-ef  2012  (found separately by the pack validator)
+ *
+ * The pack catalogue gets this check from `scripts/validate-packs.ts`. The
+ * static registry had nothing, which is why it drifted.
+ */
+describe('BOOKING_TEMPLATES reference only real BAS accounts', () => {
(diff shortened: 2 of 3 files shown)


jakobwennberg-oss:
Three findings, answered in order.

## 1. Reverse charge IS booked two-sided (resolvable, not a concern)

The bot could not see this from the diff. `lib/bookkeeping/vat-entries.ts` generates the offsetting pair, never a net:

- EU / non-EU purchase: **Debit 2645** + **Credit 26x4** (`2614` at 25%)
- Domestic reverse charge: **Debit 2647** + **Credit 26x4** (ML 16 kap)

Both legs are always emitted, so `it_cloud_hosting`'s `vat_treatment: 'reverse_charge'` with `vat_rate: 0` on the template is correct: the template carries no VAT line because the engine adds the fiktiv-moms pair itself. No silent netting.

## 2. 5420 vs 6540: the bot is right, and I am flagging rather than acting

This is a fair challenge and my justification in the PR body was weak. "Sibling templates already use 5420" shows internal consistency, not correctness.

BAS 2026 does have a better-fitting account:

```
5420  Programvaror        (group 54, förbrukningsinventarier: purchased software)
6540  IT-tjänster         (övriga externa kostnader: IT services)
```

A recurring hosting or SaaS subscription is an IT service, so `6540` is arguably the right class and `5420` reads as purchased licences.

I am **not** changing it in this PR, deliberately. The same argument applies equally to `it_saas_subscription` and `it_saas_eu`, which both use 5420 today. Moving one and leaving two would create the inconsistency this audit exists to remove, and moving all three is a categorisation change to a template family that affects where costs land in existing customers' books. That is a decision, not a bug fix.

What this PR does is unbreak a template that could never post at all. The 5420 → 6540 question is a separate call for all three, worth its own issue.

## 3. Historical Hotell postings: agreed, and this is the important one

Correct, and it is the sharpest distinction in the finding: 5614 and 5421 never resolved, so **no booking ever succeeded** through those two and there is nothing to remediate. 5820 is a real account, so every Hotell booking **did** succeed and is sitting in customers' huvudböcker as car hire.

Fixing the template only changes future postings. Per BFL 5 kap 5 § the already-posted entries need a rättelsepost if material, not a silent reclassification, and this repo's sanctioned paths are `reverseEntry()`/`correctEntry()` or the inline rättelse RPCs.

I cannot size this from here: it needs a prod query for posted `journal_entry_lines` on 5820 whose entry originated from the Hotell template. That is the second historical-remediation item this audit has surfaced (the representation template on #1396 is the other), and both want the same decision from the founder before anything touches posted entries.

PR #1401, opened 2026-08-04
fix(reports): stop double-counting arets resultat on open years

### LINKED ISSUE #1333: Balance sheet double-counts arets resultat on an open fiscal year: 2099's posted balance and the synthetic 'Beraknat resultat' are both added to equity
## Summary

On an **open** (not year-end-closed) fiscal period where a resultatavslut has been posted into account **2099 (Årets resultat)**, `generateBalanceSheet` adds the year's result to equity **twice** — once as 2099's own posted balance and once as a synthetic "Beräknat resultat" row derived from class 3-8. This produces a false imbalance warning whose `differens` equals the result.

## Environment

- Report: `/rapporter` → **Balansräkning**
- Fiscal year: **open** (`is_closed = false`), with a resultatavslut manually posted to 2099
- Commit: `8a2498a`

## Observed (real data)

**Assets**
- 1790 = 145 951,00
- 1930 = 0,50
- **Total tillgångar = 145 951,50**

**Equity / liabilities (posted balances)**
- 2081 = -25 000,00
- 2093 = -150 000,00
- 2099 = +29 918,59 (årets resultat, förlust)
- 2893 = -870,09

Manually summing every class-2 account ties out and the ledger balances. But the report shows:

- **Eget kapital och skulder = 116 033**
- **Differens = 29 918** — exactly the 2099 balance

and raises a false imbalance warning.

## Root cause

`lib/reports/balance-sheet.ts`:

1. `generateTrialBalance(..., { closingEntry: 'include' })` (correctly) keeps the resultatavslut, so **2099 carries the result** and is summed into `total_equity_liabilities` via the `'20'` (Eget kapital) group.
2. Immediately afterward, the same function computes a **synthetic** result from class 3-8 and, when non-zero, pushes an extra "Beräknat resultat" row into equity:

```ts
const periodResult = Math.round(
  incomeExpenseRows.reduce((sum, r) => sum + (r.closing_credit - r.closing_debit), 0) * 100
) / 100

if (Math.abs(periodResult) > 0.005) {
  equityLiabilitySections.push({
    title: 'Årets resultat',
    rows: [{ account_number: '', account_name: 'Beräknat resultat', amount: periodResult }],
    subtotal: periodResult,
  })
}
```

There is **no reconciliation** between (1) and (2). The synthetic row was designed for the pre-closing case (per the comment: *"Before year-end closing, this result lives on class 3-8 accounts and must be included in equity"*), but it fires **unconditionally** whenever class 3-8 is non-zero — even when the result has already been posted to 2099 on an open period. When the result exists both on class 3-8 and on 2099 (i.e. a resultatavslut was posted while the period is still open, without zeroing class 3-8), it is counted **twice** in equity, so `total_equity_liabilities` is understated by exactly the result and `differens` equals the 2099 balance.

## Expected

Årets resultat should be included in equity **exactly once**, whether it currently lives on class 3-8 (pre-resultatavslut) or on 2099 (post-resultatavslut), regardless of whether the period is formally closed. No false imbalance warning.

## Suggested fix

Reconcile the two sources of "årets resultat" so it is counted once. For example, exclude any amount already booked to 2099 from the synthetic `periodResult` (or skip the synthetic "Beräknat resultat" row when a resultatavslut is already present for the period).

## Not a duplicate of #735

#735 concerns the year-end rollover 2099 → 2091 in the **next** year's opening balance. This is a **same-year** double-count on an **open** period, independent of the rollover.

jakobwennberg-oss:
## Summary

The balance sheet built its synthetic "Arets resultat" section purely from class 3-8 rows, while 2099's posted balance already landed in the class 2 Eget kapital group (the trial balance is generated with closingEntry 'include'). When a resultatavslut was posted to 2099 on an open period whose counter-line sat outside class 3-8 (class 0/9 or missing account class), the result was counted twice and the report raised a false imbalance whose differens equaled the 2099 balance.

Fix: the period-result filter now takes every row NOT in class 1-2, written as a negated range (`!(r.account_class >= 1 && r.account_class <= 2)`) so null/undefined account_class rows are included. The counter-line of a mangled resultatavslut then offsets inside the period result, so 2099 is never counted twice. A genuinely unbalanced derived IB (untransferred prior year) still produces a real differens and the existing imbalance diagnosis still fires (covered by existing tests).

Two regression tests model the production data from the issue: 2099 in class 2 carrying the year's result, class 3-8 net carrying the same result, and an offsetting class 9 (respectively class-null) counter-row; both assert total_equity_liabilities === total_assets, no doubled equity, and no imbalance diagnosis.

No migration, no i18n changes (report labels are deliberately Swedish-only), no API changes. lib/reports/balansrapport.ts and lib/reports/imbalance-diagnosis.ts are deliberately untouched.

## Verification

- npx vitest run lib/reports: 53 files passed, 1 skipped; 779 tests passed, 2 skipped (includes cross-surface-agreement.test.ts and imbalance-diagnosis.test.ts)
- npm run lint: 0 errors; 261 warnings, byte-identical to clean origin/main (pre-existing)
- npm test (full suite): 986 files passed, 2 skipped; 12390 tests passed, 4 skipped
- check:guards not run: no app/api changes

Fixes #1333

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected balance-sheet result calculations for closing entries involving class 9 or unclassified accounts.
  * Prevented result amounts from being counted twice.
  * Improved balance accuracy and eliminated incorrect imbalance diagnoses or unnecessary result sections.


jakobwennberg-oss:
DECISIONS.md entry added in a4fda989, as flagged by CodeRabbit and the internal review pass. Note for merge ordering: PR #1405 also appends a DECISIONS.md line, so whichever merges second may need a trivial both-lines conflict resolution.

PR #1402, opened 2026-08-04
fix(salary): report F-skatt compensation on FK131 only, not FK011 too

### LINKED ISSUE #315: [payroll] F-skatt compensation double-reported on AGI (FK011 + FK131)
**Severity**: critical
**File**: `app/api/salary/runs/[id]/agi/xml/route.ts:107-113` and `lib/salary/agi/xml-generator.ts:281-319`

### Description
The AGI xml route sets `fSkattPayment: emp?.f_skatt_status === 'f_skatt' ? sre.gross_salary : undefined`, but `generateAGIXml` unconditionally emits FK011 `KontantErsattningUlagAG` whenever `emp.grossSalary > 0` (line 282) AND also emits FK131 `KontantErsattningEjUlagSA` when `fSkattPayment > 0` (line 317). F-skatt employees therefore get the same gross amount reported both as (a) subject to arbetsgivaravgifter (FK011/020) and (b) not subject to sociala avgifter (FK131). Skatteverket treats FK011 and FK131 as mutually exclusive for the same payment stream.

### Suggested fix
In xml-generator, when `emp.fSkattPayment > 0`, emit only FK131 and skip FK011. Alternative: set `emp.grossSalary = 0` in the route when it's an F-skatt payment. Also exclude F-skatt amounts from `totalAvgifterBasis` in the totals.

---

jakobwennberg-oss:
## Summary

Issue #315: an F-skatt payee's cash compensation was double-reported on the AGI. `generateAgiDeclaration` passed `grossSalary` through unconditionally while also setting `fSkattPayment`, so the XML generator emitted both FK011 (KontantErsattningUlagAG, underlag for arbetsgivaravgifter) and FK131 (KontantErsattningEjUlagSA) for the same payment in the same individuppgift. Per Skatteverket's AGI spec (and the repo's own swedish-payroll reference: Ruta 011 = gross cash salary, Ruta 131 = payments to F-skatt holders not subject to social charges) these are mutually exclusive for the same payment stream.

Fix at the data layer: `grossSalary` is now zeroed for `f_skatt` payees, so FK011 is never emitted for that payment while FK131 keeps carrying it. A generator-side guard was deliberately avoided because an IU can legitimately carry both fields for genuinely mixed payments.

Not changed on purpose:
- FK487/avgifter totals were already correct (the calculation engine sets `avgifter_basis` 0 for f_skatt); a regression test now pins that.
- The empty-IU filter already keeps f_skatt rows via `fSkattPayment`, so no individuppgift is dropped (tested).
- FK570 numbering, `totalAvgifterBasis`, and `formatAmount` (whole-krona per SKV BELOPP spec) are untouched.

Both the internal dashboard route (`GET /api/salary/runs/{id}/agi/xml`) and the v1 public route (`POST .../salary-runs/{id}/generate-agi`) are thin wrappers over `generateAgiDeclaration`, so both surfaces are fixed by this one change. No route, migration, or i18n changes.

## Verification

- `npx vitest run lib/salary`: 32 files, 513 tests, all pass (includes new `agi-generate-declaration.test.ts` covering: f_skatt IU has FK131 only, no FK011/FK001; regular employee IU unchanged; FK487/497 totals exclude the f_skatt amount; f_skatt-only roster still produces an IU; plus a generator-level FK131 case in `agi-xml.test.ts`)
- `npm test` (full suite): 987 files passed, 2 skipped; 12394 tests passed, 4 skipped
- `npm run lint`: 0 errors (261 pre-existing warnings in untouched files; touched files lint clean)
- `check:guards` not applicable: no `app/api` changes

Fixes #315

## Summary by CodeRabbit

* **Tests**
  * Added coverage to verify that F-skatt payments are reported correctly in AGI declarations.
  * Confirmed F-skatt payments are excluded from gross salary, preliminary tax, and fee totals.
  * Verified regular employee reporting remains unchanged.
  * Confirmed declarations containing only F-skatt payments are treated as valid declarations rather than zero declarations.

* **Documentation**
  * Clarified AGI field mappings for F-skatt payments and regular employees.

### DIFF
--- a/lib/salary/__tests__/agi-generate-declaration.test.ts
+++ b/lib/salary/__tests__/agi-generate-declaration.test.ts
@@ -200,3 +200,48 @@ describe('generateAgiDeclaration: F-skatt payee (FK131 only, issue #315)', () =>
     expect(result.xml).not.toContain('faltkod="011"')
   })
 })
+
+describe('generateAgiDeclaration: avgifter overrides on an F-skatt row are ignored', () => {
+  // Regression for the CodeRabbit finding on #1402: computed avgifter are
+  // already 0 for f_skatt rows, but advanced-mode overrides used to coalesce
+  // past that (override ?? computed) at three aggregation points, restoring
+  // social charges for pay whose IU simultaneously asserts FK131.
+  const REGULAR_OVERRIDE_ROW = {
+    ...REGULAR_ROW,
+    avgifter_basis_override: 30000,
+    avgifter_amount_override: 9426,
+  }
+  const F_SKATT_OVERRIDE_ROW = {
+    ...F_SKATT_ROW,
+    avgifter_basis_override: 15000,
+    avgifter_amount_override: 4713,
+  }
+
+  it('excludes F-skatt overrides from FK487, totals and avgifterByCategory while regular overrides still apply', async () => {
+    const { supabase, enqueueMany } = createQueuedMockSupabase()
+    enqueueHappyPath(enqueueMany, [REGULAR_OVERRIDE_ROW, F_SKATT_OVERRIDE_ROW])
+
+    const result = await generateAgiDeclaration({ supabase: supabase as never, ...ARGS })
+
+    expect(result.ok).toBe(true)
+    if (!result.ok) return
+
+    // The regular employee's override flows through (the override mechanism
+    // itself must keep working); the F-skatt row's override is ignored.
+    expect(result.totals.totalAvgifterBasis).toBe(30000)
+    expect(result.totals.totalAvgifterAmount).toBe(9426)
+    expect(result.totals.avgifterByCategory).toEqual({
+      standard: { basis: 30000, amount: 9426 },
+    })
+    expect(result.xml).toContain(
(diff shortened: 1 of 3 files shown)


jakobwennberg-oss:
CodeRabbit's override finding verified real: computed avgifter_basis/avgifter_amount are already zero for f-skatt rows (calculation-engine sets them 0), but manual advanced-mode overrides coalesced past that (override ?? computed) at three aggregation points in generate-declaration.ts: the per-IU effectiveAvgifterBasis, the avgifterByCategory loop feeding FK487/FK061-series, and the totalAvgifterBasis reducer. A nonzero override on an f-skatt row would restore social charges for pay whose IU simultaneously asserts FK131.

Fixed in 786fadcb: an isFSkattRow predicate now forces 0 at all three points, with the invariant documented (F-skatt compensation never contributes to avgifter aggregates, overrides included). Regression test covers nonzero avgifter overrides on an f-skatt row (excluded from FK487, totals, and avgifterByCategory) while confirming a regular employee's override still flows through. Verified the test fails without the source fix. vitest lib/salary: 514 tests passing; lint clean on touched files.

PR #1403, opened 2026-08-04
fix(salary): recompute entitled_days on ledger rows and record pre-cutover taken days

### LINKED ISSUE #1347: Vacation ledger: entitled_days never recomputed on existing rows; opening balances cannot record taken days
**Source:** support mail 28/7, item 3 (semestersaldo migration). Two gaps remain after the accrual-per-employment-date work shipped.

**Verified in code:**
1. **Stale `entitled_days` on existing rows.** The recompute path in `lib/salary/vacation-ledger.ts` (line 159) carries the old value forward verbatim: `entitled_days: row.entitled_days`. It recomputes `accrued_days` and `taken_days` but never recalculates `entitled` from `employee.vacation_days_per_year` / employment date. Concrete blocker: an employee's existing ledger row keeps `entitled_days: 25` even after the entitlement logic was fixed, so a retest still shows the wrong number until the row is reset by hand.
2. **Opening balances only capture remaining days.** Ingaende saldo per employee exists via opening balances, but only as remaining days: there is no way to record an already-taken day as taken, so history migrated from another system loses the taken/remaining split.

**Ask:**
- Make the recompute path re-derive `entitled_days` for existing open rows (or provide an explicit recompute/reset action).
- Extend employee opening balances so taken days can be recorded as taken.

jakobwennberg-oss:
## Summary

- The vacation ledger sync (`lib/salary/vacation-ledger.ts`) carried `entitled_days` verbatim on existing open rows while re-deriving `accrued_days` and `taken_days`, so a stale entitled value survived every sync. The recompute loop now re-derives entitled exactly like the lazy-seed path: `computeEntitledDays()` (unchanged, locked semesterlagen logic) for normal years, with the opening-balance cutover still outranking recomputation for the year containing `cutover_date`.
- Opening balances could not record paid vacation days already taken in the cutover year under the previous payroll system. New additive migration `20260804120000_employee_opening_balances_taken_days.sql` adds `vacation_days_taken_this_year NUMERIC NOT NULL DEFAULT 0` with a 0..40 CHECK mirroring the existing style; the column rides the existing booked-run lock trigger.
- Field threaded through: `OpeningBalancesInput`/`ROW_COLUMNS`/upsert in `lib/salary/opening-balances.ts`, the shared Zod schema (`lib/api/schemas.ts`), the MCP tool schema + MERGEABLE_FIELDS (`extensions/general/mcp-server/server.ts`), the staged-operation executor (re-parses the shared schema, no change needed beyond it), the v1 REST response schema, and the employee editor form with new sv/en labels.
- Ledger semantics for the cutover year, on BOTH seed and recompute paths: `entitled = vacation_paid_days_remaining + vacation_days_taken_this_year`, `taken = booked-run taken + vacation_days_taken_this_year`. `vacation_paid_days_remaining` keeps meaning "remaining" (backward compatible, the v1 REST API is public), and the seeded taken days survive every subsequent sync.

Note: the migration is plain additive DDL and auto-applies on merge; nothing to run manually.

## Verification

- `npx vitest run lib/salary lib/pending-operations`: 58 files, 793 tests passed (includes 3 new ledger tests: stale entitled re-derived to the pro-rated value, cutover-year row keeps opening-derived values on recompute, seed folds pre-cutover taken days)
- `npx vitest run "app/api/v1/companies/[companyId]/employees" extensions/general/mcp-server lib/api`: 97 files, 1323 tests passed (includes new 400 out-of-range + happy-path assertions for the field)
- `npx vitest run lib/reports`: 53 files passed
- `npm test` (full suite): 986 files, 12393 tests passed
- `npm run lint`: 0 errors (261 pre-existing warnings)
- `npm run test:pg` NOT run (worktree must not touch a database; migration is plain additive DDL, no trigger/RPC/RLS change)

Fixes #1347

### DIFF
--- a/app/api/v1/companies/[companyId]/employees/[id]/opening-balances/__tests__/route.test.ts
+++ b/app/api/v1/companies/[companyId]/employees/[id]/opening-balances/__tests__/route.test.ts
@@ -88,6 +88,7 @@ const SAMPLE_ROW = {
   ytd_tax: 48000,
   ytd_net: 162000,
   vacation_paid_days_remaining: 12.5,
+  vacation_days_taken_this_year: 2,
   vacation_saved_days_by_year: { [`${CURRENT_YEAR - 1}`]: 5 },
   opening_semester_liability: 42000,
   opening_semester_liability_avgifter: 13196.4,
@@ -102,6 +103,7 @@ const VALID_BODY = {
   ytd_tax: 48000,
   ytd_net: 162000,
   vacation_paid_days_remaining: 12.5,
+  vacation_days_taken_this_year: 2,
   vacation_saved_days_by_year: { [`${CURRENT_YEAR - 1}`]: 5 },
   opening_semester_liability: 42000,
   opening_semester_liability_avgifter: 13196.4,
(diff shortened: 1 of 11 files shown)


jakobwennberg-oss:
Reviewer finding worth weighing at merge time (adversarial review, non-blocking):

**Known limitation: year-close entitlement interaction.** `commitVacationYearClose` in `lib/salary/semesterberedning.ts` writes next-year rows with `next_year_entitled` (prorated over the new semesterar, honors a recorded future `employment_end`) and then immediately calls `syncVacationLedgerForEmployees`, which after this PR re-derives `entitled_days` via `computeEntitledDays`. Result: the close's value is overwritten one call later. For statutory-basis employees the sync's Semesterlagen 7 \§ value is arguably more correct; for calendar-basis employees with a recorded future `employment_end`, the close's downward proration is replaced by the flat `vacation_days_per_year`, overstating entitlement (and eventually the next close's 2920 provision). This is invisible to the test suite because `semesterberedning.test.ts` mocks the sync.

Recommended follow-up (separate PR): reconcile `nextYearEntitled()` and `computeEntitledDays()` into one entitlement source of truth that honors `employment_end`. Note the lazy-seed path already ignored `employment_end` before this PR, so the gap is not new, only newly reachable via recompute.

PR #1404, opened 2026-08-04
fix(auth): provision invitees server-side when signups are disabled

### LINKED ISSUE #1335: Invite flow silently breaks when disable_signup: true — invited new users get the email but can never create an account
## Summary

When a Supabase project is configured with `disable_signup: true` (a supported, recommended posture for closed/single-tenant deployments), the company-invite flow sends the invitation email but the invitee can never complete onboarding. The email arrives, the link works, but account creation is rejected by GoTrue, so the person is stuck with no account, no membership, and no self-service recovery.

## Steps to reproduce

1. Set the project's auth config to `disable_signup: true`.
2. As an owner/admin, invite a brand-new email via `POST /api/company/members/invite`.
3. Invitee receives the email (Resend), clicks the link -> `/invite/[token]`.
4. Since they have no account, `handleAccept()` routes them to `/register?invite=token` (`app/invite/[token]/page.tsx`).
5. `/register` calls Supabase `signUp`, which returns *"Signups not allowed for this instance"*.
6. No `auth.users` row is ever created -> no `company_members` row -> the invitee cannot log in, and there is no "reset password" option (nothing to reset).

## Root cause

The invite design assumes self-service signup is enabled. The invite page's no-account branch hard-routes to `/register` (self `signUp`), and `POST /api/company/members/invite` sends the email regardless of whether the invitee will actually be able to register. There is no admin-provisioning path, so `disable_signup: true` makes the flow fail *after* the email is sent — a silent, unrecoverable failure.

Relevant code:
- Email always sent: `app/api/company/members/invite/route.ts` (send block, ~L139–L167)
- No-account branch -> `/register`: `app/invite/[token]/page.tsx` (`handleAccept`, and the "not signed in, no existing account" render branch)
- Acceptance requires an already-authenticated user: `app/api/team/accept/route.ts`

## Impact

- Any deployment running closed signup (a common security posture) cannot invite new users at all.
- The failure is invisible to the inviter: the UI reports `email_sent: true` and a pending invitation, so it looks successful.

## Proposed fix

Provision the invited user through the service role instead of relying on self-service `signUp`, so invites work independently of `disable_signup`. Two viable approaches:

1. **Admin-invite at invite time (preferred):** in `POST /api/company/members/invite`, when the email has no existing account, call `supabase.auth.admin.inviteUserByEmail(email, { redirectTo: ${APP_URL}/invite/${token} })` (or `admin.createUser` + a recovery link) using the service role. This creates the `auth.users` row up front, so `check_email_exists` then returns true and the invite page routes to `/login` — which works even with signup disabled.
2. **Detect and branch:** read `disable_signup` (or attempt `signUp` and detect the "signups disabled" error) and, when closed, switch the no-account CTA from "create account" to an admin-provisioned magic-link/recovery flow rather than `/register`.

## Suggested implementation notes

- Keep the existing `company_invitations` + cookie acceptance path; only change *how the auth user comes into existence*.
- Guard `admin.inviteUserByEmail` behind the same owner/admin permission check already in the route.
- Surface a clear error if email delivery isn't configured (mirrors the existing `email_sent` contract).
- Add a test alongside `app/api/company/members/invite/__tests__/route.test.ts` asserting an admin-provisioned user is created when the invitee has no account.

## Workaround (current)

Manually admin-create the `auth.users` row (service role bypasses `disable_signup`), send a recovery email, and insert the `company_members` row.

jakobwennberg-oss:
## Summary

Self-hosted installations that run GoTrue with `disable_signup: true` silently broke the invite flow: an invitee without an account was routed to `/register`, where `supabase.auth.signUp` fails with "Signups not allowed for this instance", surfaced only as a generic toast.

- New server-only env flag `AUTH_SIGNUPS_DISABLED` (documented in `.env.example`; no existing flag described closed signup, so one was introduced). Self-hosters set it to mirror their GoTrue config; hosted keeps it unset and every new code path is gated on it, so hosted behavior is byte-for-byte unchanged.
- `POST /api/company/members/invite`, when the flag is true: calls `serviceClient.rpc('check_email_exists', { email_to_check: email })` (same arg name as `app/api/team/accept`); if the invitee has no account, provisions it via `serviceClient.auth.admin.inviteUserByEmail(email, { redirectTo: <appUrl>/invite/<token> })` before the Resend email. Provisioning runs before the invitation row is written, so a failure leaves nothing half-created and the admin can retry cleanly (otherwise the retry would 409 on the pending invite). The email is the schema-lowercased one, matching the invitation row and GoTrue, so `/api/team/accept`'s exact email match holds.
- Response contract: `user_provisioned` is returned alongside `email_sent`. A provisioning failure returns 502 with a Swedish message mapped through `lib/errors/get-error-message.ts` instead of a silently-successful invite.
- `getErrorMessage` learns two GoTrue patterns: "Signups not allowed" (account creation is closed on this installation, contact the person who invited you or your administrator), so the `/register` dead end is explained even for flows that bypass provisioning, and "Error sending invite/confirmation/recovery email" (GoTrue SMTP misconfiguration) so the 502 above is actionable.
- The invite acceptance path (`company_invitations` + cookie + `POST /api/team/accept`) is untouched.

## How the GoTrue invite link lands (what I found and did)

- `app/(auth)/auth/callback/route.ts` already verifies `token_hash` + `type=invite` via `verifyOtp` (the accepted type union includes `invite`), so a template-style link (`/auth/callback?token_hash=...&type=invite&next=/invite/<token>`) established a session. But nothing routed the now-passwordless user to a password surface: the flow fell through to the generic branch and dropped them on the dashboard with no password and no membership (the pre-auth invite cookie is only written by the invite page's CTA, so it is absent here).
- Fixed with minimal wiring, no new UI: `type=invite` now redirects to `/reset-password`, the existing set-password surface (a live session puts that page directly in `set-password` mode, saving via `POST /api/account/password`). The invite token is parsed from `next=/invite/<token>` and persisted as the existing pre-auth invite cookie (same flags and 7-day lifetime as `buildInviteCookie` in `app/invite/[token]/page.tsx`), so the reset-password flow's existing invite handoff (`app/(auth)/reset-password/invite-handoff.ts`) accepts the company membership right after the password is saved.
- The default GoTrue invite template (`{{ .ConfirmationURL }}`) instead lands on `redirect_to` (`/invite/<token>`) with implicit-flow fragment tokens; the invite page's browser client picks the session up and shows the existing one-click "join" button, which also works with signup disabled (the user sets a password later via the reset flow). Both landing shapes therefore complete.

## Notes for self-hosters

Hosted is untouched because the flag is unset there. Self-hosted GoTrue must have SMTP configured for `auth.admin.inviteUserByEmail` to deliver its mail; if it cannot, the failure is surfaced in the invite response (502 with Swedish guidance), not swallowed.

## Verification

- `npx vitest run app/api/company "app/(auth)" lib/errors`: 10 files, 156 tests passed (includes new tests: flag off unchanged with no admin call, flag on + existing account skips provisioning, flag on + no account provisions with the invite redirect and `user_provisioned: true`, flag on + admin failure surfaces 502 and sends nothing, `email_exists` race treated as existing account, callback `type=invite` redirect + cookie tests, and both new error-mapper patterns)
- `npm test`: 986 files passed, 2 skipped; 12398 tests passed, 4 skipped (pre-existing skips)
- `npm run lint`: 0 errors (warnings are pre-existing)
- `npx tsc --noEmit`: no errors in touched files (pre-existing errors exist in unrelated test files on main)
- No database was touched; no migration.

Fixes #1335

## Summary by CodeRabbit

* **New Features**
  - Configuration option to disable public signups while keeping admin invitations active
  - Automatic server-side account provisioning for invited users when signups are disabled
  - Enhanced password reset flow for newly provisioned accounts

* **Improvements**
  - Better error messages for disabled registrations and email delivery failures

* **Infrastructure**
  - Restored email validation database function for account verification

### DIFF
--- a/.env.docker.example
+++ b/.env.docker.example
@@ -15,6 +15,18 @@ NEXT_PUBLIC_SELF_HOSTED=true
 # Optional dedicated HMAC secret; otherwise SUPABASE_SERVICE_ROLE_KEY is used.
 # SESSION_TIMEOUT_SECRET=
 
+# Set to true when public signup is turned off in your GoTrue/Supabase auth
+# config (GOTRUE_DISABLE_SIGNUP / "Allow new users to sign up" off). GoTrue
+# offers no clean server-side read of that setting, so this flag mirrors it.
+# When true, inviting a teammate who has no account provisions the account
+# server-side via the auth admin invite API (GoTrue must have SMTP configured
+# to deliver that mail) instead of relying on public /register, which GoTrue
+# would reject with "Signups not allowed".
+# The GoTrue redirect URI allow-list (URI Allow List / GOTRUE_URI_ALLOW_LIST)
+# must include /invite/* or the invite email's redirect silently falls back
+# to SITE_URL.
+# AUTH_SIGNUPS_DISABLED=false
+
 # Optional: WebSocket origin allowed for Supabase Realtime in the CSP.
 # Defaults to NEXT_PUBLIC_SUPABASE_URL with https:// replaced by wss://
 # (http:// by ws://). Set only if Realtime is served from another origin.
--- a/.env.example
+++ b/.env.example
@@ -31,6 +31,9 @@ CRON_SECRET=generate-a-random-secret
 # provisions the account server-side via the auth admin invite API (GoTrue
 # must have SMTP configured to deliver that mail) instead of relying on
 # public /register, which GoTrue would reject with "Signups not allowed".
+# The GoTrue redirect URI allow-list (URI Allow List / GOTRUE_URI_ALLOW_LIST)
+# must include /invite/* or the invite email's redirect silently falls back
+# to SITE_URL.
 # Hosted keeps this unset: public signup stays open there.
 # AUTH_SIGNUPS_DISABLED=false
 
--- a/.env.example
+++ b/.env.example
@@ -24,6 +24,16 @@ CRON_SECRET=generate-a-random-secret
 # NEXT_PUBLIC_SESSION_WARNING_MS=120000
 # SESSION_TIMEOUT_SECRET=
 
+# Self-hosted only: set to true when public signup is turned off in your
+# GoTrue/Supabase auth config (GOTRUE_DISABLE_SIGNUP / "Allow new users to
+# sign up" off). GoTrue offers no clean server-side read of that setting, so
+# this flag mirrors it. When true, inviting a teammate who has no account
+# provisions the account server-side via the auth admin invite API (GoTrue
+# must have SMTP configured to deliver that mail) instead of relying on
+# public /register, which GoTrue would reject with "Signups not allowed".
+# Hosted keeps this unset: public signup stays open there.
+# AUTH_SIGNUPS_DISABLED=false
+
 # ── Optional: extension features (core runs without these) ─
 # AI features
 # ANTHROPIC_API_KEY=
--- a/app/(auth)/auth/callback/__tests__/route.test.ts
+++ b/app/(auth)/auth/callback/__tests__/route.test.ts
@@ -85,3 +85,41 @@ describe('GET /auth/callback: recovery flow', () => {
     )
   })
 })
+
+describe('GET /auth/callback: admin invite flow (type=invite)', () => {
+  beforeEach(() => {
+    vi.clearAllMocks()
+  })
+
+  it('routes a verified invite to /reset-password and preserves the invite token from next', async () => {
+    verifyOtp.mockResolvedValue({ error: null })
+
+    const request = new NextRequest(
+      'http://localhost:3000/auth/callback?token_hash=abc&type=invite&next=/invite/gnubok_inv_tok123'
+    )
+    const response = await GET(request)
+
+    expect(response.status).toBe(307)
+    expect(response.headers.get('location')).toBe('http://localhost:3000/reset-password')
+    expect(verifyOtp).toHaveBeenCalledWith({ token_hash: 'abc', type: 'invite' })
+    // The company invite token is persisted as the pre-auth invite cookie so
+    // the reset-password handoff can accept the membership after the
+    // password is set.
+    expect(response.headers.get('set-cookie') ?? '').toContain(
+      'gnubok-invite-token=gnubok_inv_tok123'
+    )
+  })
+
+  it('routes a verified invite without an invite path in next to /reset-password without the cookie', async () => {
+    verifyOtp.mockResolvedValue({ error: null })
+
+    const request = new NextRequest(
+      'http://localhost:3000/auth/callback?token_hash=abc&type=invite'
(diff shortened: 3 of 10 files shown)


jakobwennberg-oss:
Follow-up commit 0eb036fc addresses the adversarial-review finding that check_email_exists does not exist anywhere: it shipped in PR #229 and was lost in the #244 migration consolidation before ever reaching prod (verified missing on the hosted production database directly). The flow this PR builds on was therefore silently getting null from the RPC on every deployment, and app/api/team/accept routed even existing-account invitees toward /register.

What the commit adds:

- Migration 20260804140000_restore_check_email_exists_rpc.sql restores the function exactly as originally shipped: SECURITY DEFINER over auth.users, EXECUTE revoked from PUBLIC plus explicitly from anon and authenticated, granted to service_role only, so it cannot be used for email enumeration. Applying this fixes the hosted prod behavior too, not just self-hosted.
- tests/pg/check-email-exists.pg.test.ts locks in existence, case-insensitive matching on both sides, false for unknown emails, and the grants (anon/authenticated denied, service_role allowed).
- .env.docker.example (the template docker-compose self-hosters actually use) gains the AUTH_SIGNUPS_DISABLED block, and both env templates now note that the GoTrue redirect URI allow-list must include /invite/* or the invite email redirect silently falls back to SITE_URL.
- New invite route test for the existsError branch: when the RPC fails (function missing), the route logs a warning and proceeds to provision anyway, treating GoTrue as the authority on duplicates.

Validated: npx vitest run app/api/company (32 passed), npm run lint (0 errors), npm run check:guards (passed). test:pg not run in this session (needs a database); CI's pg-real job picks the new test up.

Note for after merge: the migration must actually reach prod for the existing-account routing fix to go live there.

jakobwennberg-oss:
Compliance follow-up. The PII log finding is fixed in 88f89dae: the new provisioning-failure log now writes a masked address (first character + *** + domain, e.g. c***@example.com) instead of the raw invitee email, with a test asserting the raw address never reaches the log record. Note the logger's shared redaction (lib/observability/redact.ts) would already have replaced a raw address with [REDACTED_EMAIL] on both the stdout and sink paths; the call-site mask keeps a useful debugging hint while guaranteeing no raw PII enters the record.

The httpOnly: false cookie finding is acknowledged as pre-existing design, not changed here. The invite cookie must be readable by client surfaces (register and reset-password read it to carry the invite through auth flows), mirroring the client-set cookie in app/invite/[token]/page.tsx, so making it HttpOnly would break that handoff. The token is low-power: the server re-checks the invitation's TTL and only accepts it for an authenticated session whose email exactly matches the invitation, so possession of the cookie alone grants nothing. Hardening the whole invite-cookie mechanism (e.g. moving the handoff server-side) would be a separate refactor if wanted.

PR #1416, opened 2026-08-05
feat(analytics): identify form fields in replays, user-entered values stay masked

mattssonn:
# What and why

Follow-up to #1412. Replays now show the sidebar, but form pages are still fully masked, so you cannot tell WHICH field a user is interacting with (e.g. quantity vs description on an invoice). This tags the shared Label primitive (`components/ui/label.tsx`, imported by ~500 files, every form in the app) with `data-ph-unmask`: field labels are static i18n chrome. Typed values stay hidden via `maskAllInputs: true`; nothing about input masking changes.

Audited every `<Label>` with dynamic children for user data: the only real case is `LineDimensionFields.tsx`, where the label IS the user's own dimension name; that one gets `data-ph-mask`, which wins even when both attributes sit on the same element (maskTextFn checks mask first). Raw `<label>` elements (~40 sites) stay masked; follow-up if wanted.

## Checklist

- [x] `npm run lint` passes locally (check:lint 0 errors; tsc clean on touched files)
- [x] New or changed logic in `lib/` or `app/api/` has tests (none changed; components are outside the test scope)
- [x] New or changed UI strings exist in both `messages/sv.json` and `messages/en.json` (no new strings)
- [x] Migrations: none
- [x] Core builds with zero extensions enabled (no `@/extensions/` imports added)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved session replay masking behavior for form labels.
  * User-defined dimension names remain masked while standard labels can appear in replays.
  * Entered field values continue to be masked.

* **Documentation**
  * Clarified how masking and unmasking attributes interact, including override behavior.


mattssonn:
@coderabbitai review

PR #1419, opened 2026-08-05
fix(accruals): carry origin dimensions onto dissolution entries

jakobwennberg-oss:
## Problem
Accrual dissolution entries were built from schedule fields only (account + amount), so a project-tagged deferred invoice line lost its dimension tag in every monthly dissolution. Per-project P&L understated costs and inflated the "(Utan dimension)" column. Verified finding G10 from the 2026-08-05 aktivering/K3/projekt analysis.

## Fix
- New column `accrual_schedules.dimensions` (jsonb, default {}), migration `20260805110000_accrual_schedule_dimensions.sql`. Not applied anywhere; goes live via merge auto-apply.
- `createAccrualSchedule` persists the merged bag; `from-invoices.ts` builds it with the same `coerceDimensionsBag` + `mergeDimensionBags` helpers the origin entries use (supplier and customer side).
- `dissolutionLines` tags BOTH the P&L target line and the interim balance line. Symmetry verified against both origin generators: supplier `groupExpenseBuckets` and customer `generatePerRateLines` keep the merged bag on the 17xx/29xx interim line, so per-dimension views of the interim account net to zero.

## Limits
Pre-existing schedules have `{}` and keep today's untagged dissolutions; backfill from `origin_journal_entry_id` is a separate follow-up. `cancelSchedulesForSource` needs nothing: stornos reverse tags verbatim.

## Verification
`npx vitest run lib/bookkeeping` 1249 tests green (7 new); eslint clean; adversarial review APPROVE (independent re-verification of the symmetry claim and cache-flow through `select('*')`).

## Summary by CodeRabbit

* **New Features**
  * Accrual schedules now preserve accounting dimensions from originating invoice lines.
  * Invoice-level dimensions are applied automatically, with item-level values taking precedence.
  * Dimensions carry through to both sides of dissolution journal entries.
  * Schedules without dimensions continue using an empty, untagged configuration.
  * Accrual dissolutions can post successfully even when preserved dimension values are archived or no longer registered.

* **Tests**
  * Added coverage for invoice schedules, dimension merging and persistence, dissolution entries, and archived dimension handling.


jakobwennberg-oss:
Rebased onto current main (no conflicts; `types/index.ts` additions sit far apart from #1414's).

On the compliance bot's two findings:

**Backfill of pre-existing schedules: measured, not assumed.** The bot is right in principle that a schedule created before this migration keeps `dimensions = '{}'`, so its dissolutions book untagged while the origin interim line was tagged, and per-dimension views of 17xx/29xx would not net to zero for that schedule. I checked prod read-only before deciding:

```
active accrual schedules:                          1
all accrual schedules:                             2
schedules whose origin entry has tagged lines:     0
```

Zero rows are affected, so a backfill migration would be a no-op carrying real risk (it would have to join `origin_journal_entry_id` back to `journal_entry_lines` and pick one bag out of several lines). Keeping it as a follow-up, now with the number attached rather than as an open-ended "separate follow-up". If the count is ever non-zero on a self-hosted install, the same query identifies exactly which schedules need it.

**Dimension-code validation:** declined here as out of scope, and it is not a regression. The whole dimensions substrate stores unvalidated `{sie_dim_no: code}` bags on `journal_entry_lines` today (invoice defaults, item overrides, categorization, salary, the retag RPC); this PR copies the same bag the origin entry already wrote, so it cannot admit a value that could not already be booked. Referential validation belongs at the substrate level, as one change across every writer, not smuggled into an accrual fix.

jakobwennberg-oss:
A pre-merge law-compliance gate caught a **regression this PR introduced**, now fixed (commit `bb4ece49`).

**The defect:** putting the origin's dimensions bag on the dissolution lines made every monthly entry non-empty for `validateEntryDimensions`, which `createDraftEntry` called unconditionally. Before this PR the lines were untagged and hit the `union.size === 0` early return, so a dissolution always posted. After it, a company with dimensions enabled whose tagged project had since been **archived** got `DimensionValidationError`; `postDueInstallments` catches it, writes `last_error` and leaves the installment pending, so both the daily cron and the manual "Bokför förfallna" button retried the same impossible entry forever.

Concretely: a 12-month prepaid invoice tagged `{"6":"P001"}` in January, project archived in June. From July, six months of cost never reach 5xxx, 17xx/29xx stays overstated, **the trial balance still balances so no year-end blocker fires**, and the årsredovisning is filed understated.

**The fix:** a dissolution is the mechanical continuation of an already-posted decision, the same category as a storno, which the engine already bypasses. New `DIMENSION_VALIDATION_EXEMPT_SOURCE_TYPES = {'accrual'}` (a separate set; the existing rules-exempt set is about required/default/fixed policy and is untouched), checked in `createDraftEntry` only. The other two `validateEntryDimensions` call sites, `updateDraftEntry` and `replaceOpeningBalanceEntry`, keep validating: accruals never reach them, and exempting import or opening balances would mask real registry errors.

**The tag is kept, not stripped.** An archived value still exists in the registry and the cost genuinely belongs to that project, so the tag is correct data; dropping it would understate the project. A hard-deleted value is unreachable here because the origin entry references it and the API 409s on referenced values.

Verified by swapping in the pre-fix `engine.ts` and re-running the new test: it throws `DimensionValidationError` for both the archived and the unknown-value case, and passes with the fix. Rebased on current main; `lib/bookkeeping` + accruals suites green (1272 tests), check:guards passes.

PR #1420, opened 2026-08-05
fix(year-end): typed preflight blocker codes so remediation links render

jakobwennberg-oss:
## Problem
`BlockerRow` in the bokslut wizard regex-matched ENGLISH phrases ("draft journal entries", "voucher gap") against blocker strings that `validateYearEndReadiness` emits only in Swedish, so the remediation links (Visa utkast etc.) could never render. The voucher-gap branch also linked to `/bookkeeping/voucher-gaps`, which does not exist as a page. Verified finding G14.

## Fix
- `validateYearEndReadiness` now returns `blockers: {code, message}[]` (required field, compile-time forced for future push sites); `errors: string[]` is kept as a derived mirror so every existing consumer is unaffected. Public v1 compliance route response is byte-identical.
- `BokslutReadinessReport` gains additive `blockerItems`; `PreflightStep` matches on stable codes, English regexes deleted.
- Links now point only at surfaces that exist; the voucher-gap and trial-balance branches were removed as dead code (no page to target; the blocker text itself still explains the problem).

## Known small compromise
`DRAFT_ENTRIES` links to plain `/bookkeeping` (the `?status=draft` param does not filter the list; the list opens on the Verifikat tab). A tab-deep-link is a separate improvement.

## Verification
`npx vitest run lib/core/bookkeeping lib/bokslut app/api/bookkeeping/fiscal-periods` green; check:guards green; adversarial review APPROVE.

## Summary by CodeRabbit

* **New Features**
  * Year-end readiness checks now provide clearer, categorized blockers for issues such as unposted transactions, unbalanced accounts, period status, and opening balances.
  * Blockers can now guide you directly to relevant areas for resolution.
  * Unsupported issues remain visible with their explanatory messages.
* **Bug Fixes**
  * Improved blocker classification for more accurate readiness results.
  * Preserved existing error messages while improving consistency between validation results and displayed blockers.


jakobwennberg-oss:
Heads-up on sequencing: #1414 just merged and adds one more blocker to `validateYearEndReadiness` (unbooked bank transactions), so this branch needs a rebase. Two mechanical follow-ups when you do:

1. **New blocker to convert.** `year-end-service.ts` now pushes a blocker for unbooked transactions plus a fail-closed variant when the check itself errors. Both are still `errors.push(...)` and need typed codes in this PR's scheme (suggest `UNBOOKED_TRANSACTIONS` and something like `UNBOOKED_CHECK_FAILED`). The MCP classifier already tags them `kind: 'unbooked_transactions'`, so the code name lines up. A sensible remediation link is `/transactions` (that page exists, unlike `/bookkeeping/voucher-gaps`).
2. **`readiness-aggregator.ts` moved.** It now also emits `ar_reconciliation_mismatch` / `ap_reconciliation_mismatch` reminders (accrual companies only) and takes `accounting_method` from the settings select. Reminders are a separate array from blockers, so `blockerItems` should drop in unchanged, but the file has shifted.

Also: #1415 originally patched the same `PreflightStep` regexes this PR replaces. I've **removed that hunk from #1415** so it no longer competes for the file: this PR owns the blocker links, and #1415 is now just the årsredovisning warning-list unblur. Your typed-code approach is the right one, and you were right that `/bookkeeping/voucher-gaps` has no page (it exists only as an API route), which the regex fix would have surfaced as a link to nowhere.

jakobwennberg-oss:
Rebased on main and #1414's two blockers converted, per your instructions.

- `UNBOOKED_TRANSACTIONS` and `UNBOOKED_CHECK_FAILED` added to `YearEndBlockerCode` (12 codes, 12 push sites). Both messages kept **byte-identical** to #1414, and the fail-closed semantics are untouched: a throwing check still produces a blocker and `ready: false`.
- MCP classifier now maps `validation.blockers` through `YEAR_END_BLOCKER_KIND: Record<YearEndBlockerCode, string>` instead of regexing `validation.errors`. Public `kind` values are byte-identical, and both new codes map to `unbooked_transactions`, which is exactly what #1414's regex already produced for both messages. The old wording heuristic is kept as a **fallback** for unmapped codes, since "someone adds a blocker without touching the map" is precisely the failure that created this task.
- `UNBOOKED_TRANSACTIONS` links to `/transactions` ("Visa transaktioner"). `UNBOOKED_CHECK_FAILED` gets no link: it is a retry condition.
- `readiness-aggregator.ts` verified after the merge: the new `ar_reconciliation_mismatch` / `ap_reconciliation_mismatch` entries go into the separate `reminders` array (accrual companies only), so `blockerItems` dropped in unchanged.

The compliance bot's "check nothing was left calling the removed `errors` reference" is verified by enumeration: 11 `blockers.push` + 1 early-return array + 9 `warnings.push`, and the only remaining `errors` references are the two deliberate `errors: blockers.map(b => b.message)` mirrors. "Inga bokförda verifikationer i perioden" was always a `warnings.push`, so the rename never touched it.

Full suite green (12668 tests), check:guards passes. One known cosmetic limit carried over: "Visa utkast" lands on `/bookkeeping`, which opens the Verifikat tab; the list has no URL param for the Utkast toggle, so a real deep-link is its own change.

PR #1421, opened 2026-08-05
fix(bokslut): remove uppskjuten skatt on obeskattade reserver in juridisk person (K3 29.37)

jakobwennberg-oss:
## What this became
Opened as a framework gate for the `uppskjuten_skatt` POST kind (analysis finding G12). During the session the underlying booking itself was verified against the swedish-asset-accounting skill and escalated: **K3 29.37 keeps obeskattade reserver at GROSS in juridisk person; the 79.4/20.6 split belongs to koncernredovisning.** The old disposition booked Dr 8940 / Cr 2240 = 20.6 percent of gross 21xx ON TOP of the gross reserves: the tax portion was double-counted (result charged twice, 2240 overstated). Founder decided 2026-08-05: remove the booking. This PR now does that removal end to end, plus the K3-to-K2 dialog from the original scope.

## Removal
- `buildDispositionsProposal`: no K3 latent-tax step; `buildLatentTaxProposal` deleted.
- POST route: `uppskjuten_skatt` kind removed from schema, ORDER map and execution (old clients get schema 400); the interim framework gate died with the feature it gated.
- `DispositionsStep`: kind case removed. `DispositionKind` union updated.
- K2-to-K3 switch no longer seeds 2240/8940 into the chart; rows already seeded for switched companies stay (audit history).
- `latent-tax-calculator.ts` reduced to `LATENT_TAX_DEFAULT_RATE`, still used for the analytical soliditet presentation in build-data; file doc now records the 29.37 rationale.

## Blast radius
Verified read-only on prod before deciding: 0 companies on K3, 3 posted 2240/8940 entries total (SIE-import shaped, untouched by this change), and INK2 adds the whole 89xx range back at field 7651 so the old error was tax-neutral.

## Kept from original scope
K3-to-K2 reversion now opens a confirmation dialog listing system consequences (2240/8940 balances and komponentavskrivningar are not unwound automatically; K3 arsredovisning content no longer applies). Legal wording about switch permission deliberately omitted: BFNAR 2012:4 is unsourced in repo skills (DECISIONS.md).

## Needs before merge
Visual sign-off of the dialog. Suggested merge order: this PR before #1423 (whose note text describes the post-removal policy).

## Verification
`npx vitest run lib/bokslut lib/errors app/api/bookkeeping components app/api/company/current`: 1143 tests green (framework test now pins that NO proposal touches 8940/2240); eslint + check:guards green; DECISIONS.md records both the escalation and the resolution.


jakobwennberg-oss:
Rebased onto current main. Two review points addressed, one accepted as a real refinement:

**"The removal is broader than the rule" (elective vs prohibition): fair, and now recorded.** 29.37 makes gross vs split an *election* for the legal entity, not a prohibition, so I have reworded the decision log: what ships is a product **default** (gross), not a claim that the split method is unlawful. Two things make the removal still correct as the default:

1. The deleted code implemented **neither** election. It credited 2240 while the balance sheet still carried 21xx gross, so the tax portion existed twice. A company that had genuinely elected the split method would not have gotten the split method from it either.
2. Nobody is on the split method: 0 companies on prod use K3 (verified read-only), so no existing accounting policy is being foreclosed.

Offering the split as a disclosed policy option is a legitimate future feature; it would have to ship with the matching redovisningsprincip note **and** a balance sheet that presents the reserves net. Logged as such.

**"Confirm nothing re-proposes or double-counts legacy 2240/8940": verified.** After this PR the only remaining references outside tests are the BAS catalog rows and `build-data.ts:1001-1002`, which reads the two accounts to *disclose* opening/closing balances in the note. No calculator, proposal or engine path touches them, so legacy balances are disclosed and never re-proposed. The note's wording is being made consistent with the new policy in #1423.

**Decision-log scoping.** You were right that the first entry described files not in this diff. It now names the PR that carries each item, and records that `K3_REQUIRED_FOR_UPPSKJUTEN_SKATT` was written and deleted **within this same PR** when the decision removed the feature it gated, so it never shipped.

Merge order still matters: **this PR before #1423**, whose note text describes the post-removal policy.

PR #1422, opened 2026-08-05
fix(assets): block Ej K2 accounts for K2 companies and fix immaterial defaults

jakobwennberg-oss:
## Problem
A K2 company could create an "Immateriell tillgang" asset with zero warnings, and the dialog defaulted it onto 1010/1019, exactly the accounts K2 forbids (BFNAR 2016:10 punkt 10.4 bans capitalizing egenupparbetade immateriella). No framework gate existed on the assets routes. Verified finding G2, the analysis's clearest compliance trap.

## Fix
- Server gate on POST and PATCH `/api/assets`: for non-K3 companies, any RESOLVED asset or accumulated account that is `k2_excluded` in the BAS reference is rejected with 422 `K2_EXCLUDED_ACCOUNT` (bilingual message pointing to the framework setting). Checked on resolved accounts so the category defaults cannot sneak through either.
- Shared helper `lib/bokslut/assets/k2-account-guard.ts`; account-level gating, no category-model change (forvarvad/egenupparbetad split stays a parked follow-up).
- Dialog: K2 companies get a purchased-intangible default (1090/1099) plus a hint that egenupparbetad utveckling requires K3; K3 aktiebolag choosing 1010 get a one-line note about fond for utvecklingsutgifter (2089, ARL 4 kap 2), wording scoped to aktiebolag after review.
- Account 1081 (Pagaende projekt for immateriella) was missing its `k2_excluded` flag in both the TS reference and the original 20260225103139 backfill: flagged in TS + data migration `20260805112000_flag_1081_ej_k2.sql` for existing chart rows. 1260/1269 stay unflagged: no asset category range reaches them; full k2_excluded-list reconciliation against official BAS marks is a separate follow-up.

## Needs before merge
Visual sign-off of the dialog hints.

## Verification
`npx vitest run app/api/assets lib/bokslut/assets` 68 tests green (K2+1010 rejected, K2+1081 rejected, K2+purchased accepted, K3+1010 accepted, PATCH equivalents); check:guards green; review verdict after fix round: the flagged residual hole (1081) is closed in this revision.

## Summary by CodeRabbit

* **New Features**
  * Added validation preventing K2 companies from using accounts restricted to K3 accounting.
  * Added framework-specific defaults for immaterial assets: 1090/1099 for K2 and 1010/1019 for K3.
  * Added clearer guidance when accounting-framework restrictions apply.
  * K3 component functionality now requires the K3 framework.

* **Bug Fixes**
  * Corrected account 1081 classification as unavailable under K2.
  * Asset creation and updates now consistently enforce framework rules.

* **Tests**
  * Expanded coverage for restrictions, category changes, defaults, and updates.


jakobwennberg-oss:
Rebased on main. The bot's third finding was right and is fixed; the others are answered below.

**Wrong legal citation on non-intangible accounts: fixed.** `k2ExcludedAccountMessage` cited BFNAR 2016:10 punkt 10.4 (egenupparbetade immateriella) for *every* `k2_excluded` account, including 1370, 2240, 8940 and friends that are excluded for entirely different reasons. The boundary is now **derived from the BAS data** (`k2_excluded && account_class === 1 && account_group === '10'`), which resolves to exactly 1010, 1011, 1012, 1018, 1019, 1081, so it tracks `bas-data` instead of a hardcoded list. Those get the 10.4 citation; the other 21 accounts get a correct generic message that claims only what is verifiable (the account is marked Ej K2 in BAS and requires K3) without inventing a paragraph reference. Both languages come from one function so sv and en cannot drift.

Reachability was real, not theoretical: POST range-checks BAS overrides in Zod, PATCH does not, so `PATCH { bas_asset_account: '1370' }` reached the gate and got the wrong citation. That path is now pinned by a test that fails against the previous code, plus a guard-level test asserting no account outside kontogrupp 10 ever cites 10.4.

**On the other findings:**
- *Fond för utvecklingsutgifter (2089) is not booked:* correct, and deliberately out of scope. This PR gates accounts; the fond flow is a known gap from the same analysis (nothing books 10xx/38xx either) and is queued as its own work. The dialog hint is informational only, which is why it is phrased as a consequence rather than a promise.
- *1081 might cover purchased-but-incomplete intangibles:* the repo's own domain reference lists 1081 under "Accounts to hide/disable in K2 mode" (`swedish-year-end-closing/references/k2-vs-k3.md`), which is the authority I am required to follow over inference about account naming.
- *No remediation for assets already booked on 1010-1019/1081:* true, no detection was added. Per BFL those would need a rättelse, not silent reclassification, so a detection report is the right shape for it and belongs in its own change.

Full suite green (12688 tests), check:guards passes (naive-ore-round down 3 from baseline).

jakobwennberg-oss:
The same pre-merge law gate caught a **regression this PR introduced**, now fixed (commit `c80aedd1`): the gate was blocking something **lawful**.

K2 forbids only *egenupparbetade* immateriella tillgångar; **acquired ones may be recognized** (`k2-vs-k3.md:26`). But `asset-service.ts` still resolved category `immaterial` to 1010/1019 for everyone server-side, and only `CreateAssetDialog` compensated by sending an explicit override. `EditAssetDialog` sends just the changed fields and has no account inputs, so a K2 aktiebolag recategorizing a purchased licence or trademark to "Immateriell tillgång" hit the raw defaults and got a 422 telling it to move the company to K3. The previous revision even encoded that refusal as an intended test.

**Fix, both halves:**

1. The server-side default is now framework-aware in one place: `defaultAccountsForCategory(category, framework)` returns the acquired pair 1090/1099 for `immaterial` whenever the framework is not K3. Both `createAsset` and the recategorization realign in `updateAsset` use it, so create and edit get the lawful default without the client knowing, and the client-side override in the create dialog was removed so the two surfaces cannot drift.
2. The 422 no longer asserts the company's framework (the companies read can fail transiently) and no longer prescribes a framework change. It now says the account is reserved for egenupparbetade utvecklingsutgifter, which require K3, and that an acquired intangible belongs on 1090.

The unlawful case still rejects. The test matrix now pins: K2 on defaults → 1090/1099 accepted; K2 explicitly overriding onto 1010/1019/1081 → 422; K2 recategorizing → accepted; **K2 moving a legacy 1010 asset onto a purchased pair → accepted** (the remediation path a company with a pre-existing non-compliant asset needs); K3 on 1010 → accepted; 1370 → generic message with no punkt 10.4 citation.

Rebased on current main; assets/bokslut/errors suites green (676 tests), check:guards passes.

PR #1423, opened 2026-08-05
fix(arsredovisning): truthful K3 policy note and GRANSKNINGSUTKAST watermark

jakobwennberg-oss:
## Problem
Two verified findings (G3, G4). The K3 redovisningsprinciper note unconditionally asserted policies the engine does not implement: uppskjuten skatt "enligt balansrakningsmetoden for temporara skillnader" and "finansiella leasingavtal redovisas som anlaggningstillgang" (no code capitalizes leases). A user signing the PDF signed a false policy statement. Meanwhile every K3 report is internally classified AR-K3-DRAFT-ONLY and can never be finalized or filed, yet the PDF downloaded with no draft marking.

## Fix
- Note text now states the decided policy (founder decision 2026-08-05, K3 29.37): uppskjuten skatt hanforlig till obeskattade reserver sarredovisas INTE i juridisk person; obeskattade reserver redovisas inklusive uppskjuten skatteskuld. Engine-side removal of the old 8940/2240 disposition lives in #1421; merge that first. The 2240-movement note remains and is already conditional on actual balances (imported/legacy companies keep their disclosure).
- Leasing: samtliga leasingavtal kostnadsfors linjart som operationella. Komponentavskrivning paragraph untouched.
- K3 PDF gets a fixed, rotated GRANSKNINGSUTKAST watermark on every page (react-pdf fixed element, low opacity; no `break` props). K2 PDF untouched. Visually verified in-session: watermark on all pages without obscuring content.
- Wizard: always-visible quiet banner for K3 explaining draft-only status and the manual alternative, mirroring capabilities.ts reason strings.
- Drive-by in the same surface: the K3 PDF header said "2025-01-01: 2025-12-31" (a date range laundered into a colon by the em-dash sweep); now "till", matching the K2 template.

## Verification
`npx vitest run lib/bokslut/arsredovisning` 143 tests green; eslint clean; adversarial review APPROVE; PDF visually inspected.


jakobwennberg-oss:
The compliance bot caught a real one, and the fix round then caught two more of the same kind. Rebased on main; three commits now.

**The contradiction (bot's finding).** `buildK3RedovisningsPrinciper` denied that deferred tax on obeskattade reserver is separately recognised, while `buildUppskjutenSkattNot` disclosed exactly that split with 2240 balances. A signed årsredovisning cannot do both. Both notes are now driven by **one signal** derived once in `build-data.ts` from the same trial balance, so no code path can produce the pair. A company with no 2240/8940 activity gets the gross-method policy and no movement note; a company carrying a legacy or imported balance gets a policy paragraph that acknowledges the recognised liability and points at the note that details it. Neither branch claims *how* the balance was measured, since an imported provision may have been booked at 21.4 or 22 percent by another system.

**Two more of the same class, found while fixing:**

1. *Fail-open denial.* When the 2240/8940 figures could not be read, the code warned and then passed "no recognised deferred tax", so the note printed the affirmative **denial** on data it had just admitted it could not read. Deferred tax is now a tri-state (`none` / `recognized` / `unknown`) and the read-failure path prints only the going-forward policy, never a denial.

2. *Leasing asserted more than it could know.* The paragraph claimed "Årsredovisningen avser en juridisk person och någon koncernredovisning upprättas inte", a categorical fact about the entity that the same document contradicts as soon as the user declares a parent company (build-data emits a Koncernförhållanden note). That clause is gone; the 20.29 basis for the treatment stays, since that is a statement about the rule being relied on rather than about the group. Separately, the blanket "samtliga leasingavtal redovisas som operationella" now yields to the balansräkning: if leased assets sit on 1260/1269, the paragraph describes what is booked instead of asserting a simplification the numbers contradict.

**On the source trap:** the 20.29 exemption is genuinely sourced, just not where my brief said. It is in `swedish-asset-accounting/references/leasing-and-disposal.md:32` ("In the juridisk person, financial leases MAY be accounted for as operating") and `swedish-asset-accounting/SKILL.md:52`, not in `k2-vs-k3.md`. The code comment now points at the right file.

**Still open, flagged not fixed:** note 4's third line is hardcoded "Årets förändring (8940)" even on the fallback branch where no 8940 row exists and the change is derived as `closing - opening`, which mislabels the account; and 8940 also carries movements against 1370, so opening + change is not guaranteed to equal closing for a company using both. Both are pre-existing and belong with the wider K3 note-matrix work.

151 tests green in `lib/bokslut/arsredovisning`, eslint and tsc clean on the changed files. Merge after #1421, whose engine change this text describes.

PR #1426, opened 2026-08-05
feat(invoice-inbox): show receipt fields on receipts, invoice fields on invoices

jakobwennberg-oss:
Founder observation from the live Underlag view, looking at a WhatsApp receipt from Lennart & Bror Kött.

## The problem

The rail showed the same twelve fields for every document, so a receipt got **Fakturanr**, **OCR/Referens**, **Bankgiro**, **Plusgiro** and **Förfallodatum** as five empty boxes. None of those concepts exist on a kassakvitto: it is already paid. A perfectly extracted receipt therefore read as a failed extraction.

The date was mislabelled as well. On a receipt the document date is the purchase date, not a "Fakturadatum".

## The change

When `extracted_data.documentKind` is `receipt`:
- the date field is labelled **Inköpsdatum**
- the five invoice-only fields fold behind a quiet **"Visa fakturafält (N)"** link

Invoices, government letters and unclassified documents are completely unchanged.

## Safety when the classification is wrong

1. **A field holding a value is never hidden.** A hybrid restaurangnota that really does carry an invoice number still shows it, and the hidden count drops accordingly.
2. **The fold is one click from being undone**, so nothing is unreachable.

## Where the logic lives

`lib/documents/inbox-field-visibility.ts`, not the component. This repo is a node-only unit project with no jsdom or testing-library, so display rules are only testable when extracted as a React-free module (same reasoning as `channel-context-notes.ts`).

## Verification
- `npm test`: 12682 passed
- `npm run lint`: 0 errors
- `npx tsc --noEmit`: 417, exactly the main baseline
- `npm run build`: succeeds
- New test covers: invoices untouched, five hidden on receipts, relabelling, value-never-hidden, showAll restoring everything, and no mutation of the caller's list

## Related, not fixed here

The date on that receipt is genuinely empty in the data (`invoiceDate: null`), which is a separate extraction issue I could not reproduce outside production. This PR makes the field read correctly; it does not fill it.

## Summary by CodeRabbit

* **New Features**
  * Receipt documents now display a focused set of relevant fields by default.
  * Populated fields remain visible, even when not part of the default set.
  * Added an option to reveal all available fields, with the number of hidden fields shown.

* **Bug Fixes**
  * Receipt field labels now use receipt-appropriate terminology, including the purchase date.

### DIFF
--- a/components/extensions/general/InvoiceInboxWorkspace.tsx
+++ b/components/extensions/general/InvoiceInboxWorkspace.tsx
@@ -2335,6 +2335,8 @@ export function EditableFieldsList({
   // from the extraction) and flips to user-verified once the user edits it:
   // mirrors the create form's AiFilledIndicator. Reset when switching items.
   const [edited, setEdited] = useState<Partial<Record<FieldKey, boolean>>>({})
+  // Per-document fold for the invoice-only fields on a receipt.
+  const [showAllFields, setShowAllFields] = useState(false)
   const timersRef = useRef<Partial<Record<FieldKey, ReturnType<typeof setTimeout>>>>({})
   // Last-known server values per field. Used to detect when the server
   // normalises a value (currency upper-cased, whitespace trimmed) so we can
@@ -2352,6 +2354,10 @@ export function EditableFieldsList({
     setDrafts(seeded)
     lastServerRef.current = seeded
     setEdited({})
+    // The "Visa fakturafält" fold is per document: without this, expanding
+    // it on one receipt leaves the invoice fields open on the next one,
+    // which reads as if that document had them too.
+    setShowAllFields(false)
     return () => {
       for (const t of Object.values(timersRef.current)) {
         if (t) clearTimeout(t)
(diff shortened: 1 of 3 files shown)


jakobwennberg-oss:
Both CodeRabbit findings triaged.

**1. `showAllFields` leaking across items — valid, fixed** in 019e16b1 (pushed after the reviewed commit). Rather than keying the state on the item id, I reset it inside the existing `itemId` effect alongside `drafts` and `edited`, so all three per-document states are cleared in one place and cannot drift apart. Same outcome, one mechanism instead of two. PR Agent independently flagged the same thing.

**2. `beforeEach` with `vi.clearAllMocks()` + event-bus reset — skipped deliberately.** That convention exists for suites that mock modules or emit events. This one tests a pure function: no mocks, no bus, no shared state to leak. Every comparable suite in the same directory agrees:

| suite | beforeEach | eventBus | vi.* |
|---|---|---|---|
| channel-context-notes.test.ts | 0 | 0 | 0 |
| core-receipt-matcher.test.ts | 0 | 0 | 0 |
| link-documents.test.ts (mocks) | 0 | 0 | 7 |

Importing the event bus into a test that never touches it would add a dependency purely to satisfy the letter of a rule whose purpose is isolation the suite already has.

PR #1430, opened 2026-08-06
fix(supplier-invoices): reverse credit notes on paid kontantmetoden invoices

jakobwennberg-oss:
## The bug

The supplier credit flow gated the reversing verifikat on `accounting_method === 'accrual'` in all three surfaces (dashboard route, v1 route, pending-operations commit).

That is correct only while the original is still **unpaid**: nothing reached the ledger, so there is nothing to reverse and recognition rightly waits for cash.

It is wrong once the invoice has been **paid**. Under kontantmetoden the payment verifikat already booked the expense and claimed the ingående moms on 2641 (ruta 48). Crediting it marked the invoice `credited` with **zero accounting trace**, leaving both the cost and the moms deduction overstated for as long as the credit stands, and leaving nothing for a later refund to reconcile against.

The customer side has never had this bug: `creditNoteNeedsJournalEntry()` books the reversal whenever the original reached the ledger, whatever the configured method. This restores the symmetry.

## Verification of the premise

Confirmed against `.claude/skills/swedish-vat` and the customer-side implementation before writing code, not from memory.

`createSupplierCreditNoteEntry`'s existing shape already suits the cash case: the **2440 debit leaves a claim on the supplier** that the refund payment clears, exactly as the customer side leaves a 1510 credit for a refund owed to the customer. No new entry shape was needed.

## Change

New `supplierCreditNoteNeedsJournalEntry()` in `lib/bookkeeping/booking-mode.ts`, the mirror of the customer predicate. Reverses when the original carries any booked-ness signal: `registration_journal_entry_id`, `payment_journal_entry_id`, `status === 'paid'`, `paid_at`, or a non-zero `paid_amount`.

The v1 route's GDPR-minimised projection had dropped exactly those columns; they are restored with a comment explaining why. `status` alone is too weak: it misses a part-paid-but-booked original, which is precisely the shape of the pre-#1413 rows.

## Tests

- 7 new predicate cases (each booked-ness signal in isolation, öre noise, null rows).
- 2 new route cases: unpaid cash original still skips; paid cash original now reverses.
- The paid-cash route test was **verified to fail against the old gate** before being kept, so it is real coverage rather than a green that proves nothing.
- Full suite green (12,757), lint 0 errors, `check:guards` passes, no new tsc errors.

## Not in scope

Per-installment proportional cash recognition, and the prod-side question of whether any already-credited paid cash invoices need a repair entry (a scan for the #1413 corruption shape found **zero** affected rows, but this is a distinct shape and was not scanned).


jakobwennberg-oss:
## Compliance review triage

**Finding 2 (reversal amount vs actually-booked amount for partial payments): checked, no change needed.**

The bot flagged this as "not verifiable from the diff alone", so here is the verification:

- `createSupplierInvoiceCashEntry(supabase, …, items: SupplierInvoiceItem[], …)` books the cash-method payment entry **from the invoice items**.
- `createSupplierCreditNoteEntry(supabase, …, creditNote, items, …)` is called with the **ORIGINAL invoice's items** (that is what the `// Pass the ORIGINAL items` comment is about).

Both derive from the same basis, so the reversal matches what the payment verifikat actually posted, line for line. There is no pro-rating to do: as documented on `cashPartialBlockReason` (#1413), the cash entry always books the full invoice and never takes a payment amount, so "the amount actually booked" *is* the full item set.

**Note on the `null` fail-open:** all three call sites fetch `original` and return 404 before reaching the predicate (`SI_NOT_FOUND` in both routes, `'Supplier invoice not found'` in the commit path), so `null` is unreachable there. The `null` branch exists only to keep the predicate total.

**Note on the GDPR projection:** deliberate and called out in the diff comment. The four re-added columns (`registration_journal_entry_id`, `payment_journal_entry_id`, `paid_at`, `paid_amount`) are the booked-ness signals the fix requires; they are booking metadata on a row the handler already reads in full, not personal data about a new subject. `status` alone was tried and rejected: it misses a part-paid-but-booked original.

**Finding 1** confirms the fix closes a real ingående-moms overstatement.

PR #1432, opened 2026-08-06
feat(bokslut): kontantmetoden year-end cut-off for fordringar and skulder

jakobwennberg-oss:
## The gap

Under kontantmetoden nothing reaches 1510/2440 during the year. But BFL still requires that **fordringar och skulder are booked at räkenskapsårets utgång**, and that conversion did not exist anywhere in the codebase.

Consequences today: every cash company's balance sheet omits its open invoices at year-end, and the AR/AP tie-outs added in #1414 are permanently "unreconciled" by construction for all 334 cash companies on prod.

## Verification of the premise

Confirmed from the skills before writing code, not from training data:

- `swedish-accounting-compliance/references/bfl-bfnar.md`: *"Kontantmetod still requires that fordringar/skulder are booked at bokslut."*
- `swedish-vat/references/vat-compliance-reference.md`: *"Bokslutsmetoden (cash) available for ≤ 3M SEK; VAT recognized on payment **except at year-end**."*

## The moms account question, resolved

This was previously parked on a founder decision (2641 vs 2647/2648). The BAS 2026 chart settles it:

| Account | Name | In momsdeklaration? |
|---|---|---|
| 2641 | Debiterad ingående moms | **yes, ruta 48** |
| 2647 | Ingående moms omvänd betalningsskyldighet | domestic reverse charge, unrelated |
| **2648** | **Vilande ingående moms** | **no** |
| **2618/2628/2638** | **Vilande utgående moms 25/12/6 %** | **no** |

The `vilande` accounts are seeded in the chart and deliberately absent from `ACCOUNT_RUTA` / `ACCOUNT_TO_BOX`, so moms parked there stays out of the declaration until payment, which is exactly what bokslutsmetoden requires. **Booking cut-off moms to 2641 would claim the deduction a period early** and is the actual error this avoids. 2647 was considered and rejected.

## Entry shape

```
Fordringar: Debit 1510      / Credit 30xx / Credit 2618|2628|2638
Skulder:    Debit 4-6xxx
            Debit 2648      / Credit 2440
```

Two **aggregate** verifikat, each reversed on day 1 of the next period.

Deliberately **not** per-invoice, and deliberately **no** `invoices.journal_entry_id` link: the payment flows route on that link, so per-invoice linking would send every new-year payment down the accrual clearing path against a receivable the vändning has already removed, booking the settlement twice. Leaving it unset means a new-year payment still books the normal kontantmetoden cash entry at the real payment date. The day-1 reversal is what makes that safe.

## Correctness details

- **Outstanding is computed from payment DATES**, not `remaining_amount`. An invoice settled in January was still a fordran on 31 December; reading `remaining_amount` would silently shrink the cut-off every day the bokslut is delayed.
- **Net is derived as `outstanding - vat`**, so the two legs always add back to the receivable whatever rounding the source rows carry.
- **Integer öre with largest-remainder allocation** (`distributeOre`) when splitting a net across several expense accounts: rounding each share independently drifts an öre per bucket and the DB balance trigger would reject the entry.
- A zero-moms treatment carrying a stray moms amount (data error) folds into revenue rather than inventing a moms account, so the verifikat still balances.
- Uses the canonical `roundOre` from `lib/money.ts`; the antipattern guard count is unchanged from main (630).

## Surfacing

A bokslut wizard reminder at **warning** severity, not a blocker. Promoting it to a blocker would newly block every cash company mid-bokslut, which is a separate call to make deliberately rather than as a side effect of this PR.

## Tests

18 cases on the pure core (balance invariants on uneven amounts, per-rate vilande routing, the vändning netting 1510 to exactly zero, leap-day and broken-fiscal-year rollover, degenerate weights) plus 4 aggregator cases covering the cash branch, the accrual no-op, and graceful degradation.

Full suite green (12,757), lint 0 errors, `check:guards` passes, no new tsc errors. **No migration** (reuses the existing `year_end` source type).

## Not wired yet

`postKontantmetodCutoff()` is implemented and tested but not yet exposed through a wizard action or MCP tool: posting real verifikat into users' ledgers wants an explicit UI step and your sign-off on the blocker-vs-warning question above. **Nothing posts to any ledger from this PR.**


jakobwennberg-oss:
## Compliance review round 2: triage

Three findings were data-dependent. I checked each against production rather than reasoning about it.

**Credit notes silently excluded — not a defect, verified.**
All 22 credit notes on prod carry `document_type = 'invoice'`:

| document_type | is_credit_note | rows |
|---|---|---|
| invoice | false | 10,370 |
| invoice | **true** | **22** |
| proforma | false | 2 |
| delivery_note | false | 1 |

The filter drops only the 2 proforma and 1 delivery_note, which is the intent. Credit notes are inside the set and reduce the receivable through their negative totals, exactly as the comment says.

**Vilande account numbers — verified, and the suggested alternative does not exist.**
The bot only had the SKILL.md quick reference, which does not enumerate these. The BAS 2026 chart in `lib/bookkeeping/bas-data/class-2-equity-liabilities.ts` does:

- `2618` Vilande utgående moms, 25 % • `2628` … 12 % • `2638` … 6 % • `2648` Vilande ingående moms

There is no 2617/2627/2637 in the chart. (I guessed those myself first; the chart corrected me.) Their absence from `ACCOUNT_RUTA` / `ACCOUNT_TO_BOX` is the whole mechanism, and there is a test asserting nothing lands on 2611/2641.

**Reverse charge — real hardening, fixed in e509150d.**
All 123 RC supplier invoices on prod carry `vat_amount = 0`, so no RC moms can reach 2648 today and no one-sided reverse charge was ever possible. But that was an *implicit data invariant*, not an enforced one. `CutoffPayable` now carries `reverseCharge` and forces cut-off moms to 0 for those rows, so a stray amount can never be routed into the single vilande bucket. The self-assessed output/input pair stays with the payment entry, after the vändning. Two tests added.

**Truncated underlag references.** The affärshändelse being booked is the bokslutspostering itself (one event, dated period end), not the individual invoices, so this is not a samlingsverifikation of many same-day events. The note is supplementary to the reskontra, which is the underlag. Improved so a truncated note explicitly names the reskontra as the full specification rather than implying the listed subset is everything.

**Warning vs blocker.** Deliberate, and called out in the PR description. Promoting it would newly block all 334 kontantmetoden companies mid-bokslut, which should be an explicit decision rather than a side effect of this PR. There is no hard gate at lock time today, so the bot is right that a company *can* close without posting the cut-off. That is the status quo, not a regression, and it is the founder's call. Flagging it as the open question on this PR.

jakobwennberg-oss:
## Compliance review round 3: triage, and stopping here

The findings have plateaued into repeats of items already verified above. Answering the one new item, then re-stating the answers for the record.

**Period-lock gap on the closing period — not a gap, verified.**
`createJournalEntry` does enforce it: the `enforce_period_lock()` trigger (migration `20240101000017_enforcement_triggers.sql:96`) rejects any write when `fiscal_periods.is_closed = true OR locked_at IS NOT NULL`, and the engine validates the period before mutating.

Worth explaining why `assertReversalPeriodPostable` exists at all given that: the trigger protects both periods, but it fires *at write time*. For the **next** period that is too late, because the cut-off entry has already committed by then, leaving exactly the un-reversed 1510/2440 inflation this module is built to prevent. The preflight moves that specific failure ahead of the first write. The closing period needs no such preflight: it is the first write, so the trigger rejecting it is already all-or-nothing.

**Already answered above, unchanged:**

- *Vilande account numbers* (3rd raise): verified against the BAS 2026 chart at `lib/bookkeeping/bas-data/class-2-equity-liabilities.ts` lines 1868/1956/2044/2121: `2618` / `2628` / `2638` Vilande utgående moms 25/12/6 %, `2648` Vilande ingående moms. The skill's SKILL.md quick-reference table is a summary and does not enumerate them; the chart is the source of truth. On the 2650 concern: the vilande accounts are never touched by the moms settlement, which drives off `ACCOUNT_RUTA`, so nothing leaks into the clearing account.
- *Credit notes* (3rd raise): checked against production. All 22 credit notes carry `document_type = 'invoice'`; only 2 proforma and 1 delivery_note are excluded by that filter.
- *Aggregate traceability* (3rd raise): the affärshändelse booked is the bokslutspostering itself, dated period end, not the individual invoices, so this is not a samlingsverifikation of many same-day events.
- *Reverse-charge timing*: flagged by the bot as "unsure, not asserting a hard defect". This PR does not change RC timing in either direction: RC invoices carry `vat_amount = 0` (all 123 on prod), so the cut-off never posted an RC moms leg before this change either. The `reverseCharge` flag added here only makes that existing behaviour explicit so a stray amount cannot produce a one-sided RC. Whether RC self-assessment should be pulled to the affärshändelse date under bokslutsmetoden is a real question, but it is a pre-existing question about the payment entry, out of scope here, and I would rather not change moms timing on the strength of an "unsure" flag.

Stopping the review loop here: three rounds, one new legitimate finding in the last two (stray moms on momsfri treatments, fixed in 555e96bc), and the rest repeats.

PR #1437, opened 2026-08-06
feat(sandbox): seed payroll, articles and a year of ledger history; calm the connect CTAs

jakobwennberg-oss:
The sandbox showed neither Löner nor a usable set of reports, and the "connect X" surfaces were oversized boxed cards. Founder-requested, 2026-08-06.

## Sandbox: payroll

`Löner` was invisible because `pays_salaries` was false, not because data was missing: `DashboardNav` gates `/salary` and `/salary/employees` on `entityType === 'aktiebolag' || paysSalaries`.

- Flips `pays_salaries` + `employer_registered`. Both seeded employees are `employment_type: 'employee'`: an enskild firma may employ staff, it just may not put its own owner on payroll.
- Two employees (månadslön 38 000, timlön 245/tim), one **Bokförd** run for last month and one **Utkast** for this month.
- The three verifikat the booked run must have posted (7210 → 2710/1930, 7510 → 2731, 7290+7519 → 2920+2940), linked via `salary_entry_id`. A run in status `booked` that posted nothing would be a lie.
- Skatteavdrag comes from the real Skatteverket 2026 tables (`FALLBACK_TAX_TABLES_2026`, table 34 col 1), not a schablon. The draft run ships calculated, so its live "Beräkna om" would otherwise have jumped ~4 600 kr away from the sibling booked run.
- `seed_chart_of_accounts` only lays down its 7xxx personnel block for aktiebolag, so the seed now creates the missing accounts from `getBASReference`, the way `ensureSalaryAccountsExist` does.

## Sandbox: articles + ledger history

- Artikelregister (5 rows, one at 6 % because a printed book genuinely is, ML 9 kap.).
- Year-to-date ledger, January through last month, so Resultatrapport, Balansrapport, Nyckeltal and Momsrapport show a real shape rather than a flat line.
- **Quarterly momsredovisning** cleared to 2650 and paid on the SFL 26 kap. 26 § deadline. Without it the demo collected VAT all year and never remitted it: moms read 155 813 kr "att betala" and the bank balance climbed without limit. Now 12 834 kr. The 2611+2641+2650 shape is what `get_vat_declaration_totals` classifies as a settlement and excludes, so the next declaration is not double-counted.
- History is exempted through `journal_entry_no_doc_required`, the same way the SIE-import opt-in treats imported books: its kvitton live in the previous system. Unflagged it put **"Verifikat utan underlag: 39"** on the demo's first screen.
- History is numbered before the invoice and payroll vouchers, so the series runs forwards through the year (`next_voucher_number` hands out numbers in call order). Writes are batched: ~130 round trips down to ~45.

## Connect CTAs

- **Bank picker**: a two-column grid of ~95px bordered logo cards becomes flat 48px hairline rows, Lucide icons instead of hand-rolled SVGs, `hover:bg-secondary/60` instead of `hover:border-primary`, and a quiet inline connecting state.
- **Cloud backup**: each provider collapses from a ~160px boxed panel to one row; the BFL 7-year note is shown once per section and names only configured destinations. OAuth scope copy stays visible rather than hiding behind a "?".
- **Hem first-run**: only the active step argues its case, but every not-done step keeps a reachable, focusable action. The Skatteverket nudge becomes one quiet sentence instead of a boxed card with a hero-weight title.

Both settings surfaces now speak the locked "Fönster" language (flat hairline rows, never boxed cards).

## Mobile assistant FAB

A fresh open is desktop-only (`hidden md:flex`): the bottom nav already carries an Assistent tab behind the same `identity.isVerified` gate. A **collapsed** session keeps its handle on every viewport except `/chat`, because the nav tab opens a fresh chat surface rather than restoring a minimized conversation, and hiding it would strand a session mid-booking.

## Security fix found along the way

`/api/salary/runs/[id]/payslips/send` had **no `guardSandbox`** while every sibling send route had one. Seeding a booked salary run put "Skicka lönebesked" one click from an anonymous visitor, with the live Resend service behind it on hosted. Guarded, with a test asserting the mail provider is never called; seeded employee emails are `null` as a second lock.

## Verification

- `npm run lint` clean (0 errors; the 2 warnings on `BankSelector` are pre-existing on main)
- `npm test`: **12 921 passing, 0 failing**
- `npm run check:guards` passing (and `naive-ore-round` improved by 6)
- **No migrations**, so nothing to apply to prod on merge
- Visually verified against a real seeded sandbox at 1440px and 390px, no horizontal overflow at either

`tests/schema/no-phantom-columns.test.ts` ceiling raised 360 → 375 with a documented reason: the pure-row-builder + spread-at-call-site pattern (the one `customers.ts` / `pending-operations.ts` already use) is opaque to that scanner, and writing the columns out again at the call site would duplicate every builder's shape.

## Summary by CodeRabbit

* **New Features**
  * Added richer sandbox bookkeeping and payroll data, including articles, ledger history, employees, salary runs, payslips, and vouchers.
  * Sandbox payroll now uses the 2026 fallback tax tables.
  * Redesigned onboarding checklist with numbered progress steps and contextual actions.
  * Improved cloud backup and bank connection interfaces with compact provider and bank lists.

* **Bug Fixes**
  * Prevented payslips from being emailed from sandbox environments.
  * Improved agent trigger visibility and positioning across screens.
  * Simplified the Skatteverket promotional section for clearer presentation.


jakobwennberg-oss:
Addressed in 02a86d43.

**Fixed**

- **Major, unchecked Supabase errors** (`route.ts`): both call sites now destructure and throw. This was a real hole: a failed `chart_of_accounts` re-select would have written `account_id: null` onto every ledger-history and salary voucher line, and a failed `next_voucher_number` would have inserted a *posted* verifikat with no number, which is a gap in the verifikationsserie (BFNAR 2013:2). A null voucher number is now rejected explicitly rather than passed through.
- **A-004 markup arithmetic**: 4860 → 5400 is 11.1 %, not 10 %. Kept the price (it is the realistic number) and dropped the false percentage from the note, since the note is user-visible copy in an accounting product.
- **Test name contradicted its assertions**: leftover from switching the seed off a schablon and onto the real Skatteverket table. Renamed.
- **Ledger-history doc said "4 to 6" verifikat per month**: the quarterly momsredovisning and its payment add a seventh in March, May and June. Now "4 to 7", matching the test.
- **Bank picker a11y**: good catch. `Loader2` is `aria-hidden`, so loading and connecting had no text equivalent and a failed bank fetch was never announced. Added `role="status"` with an `sr-only` label for both, and `role="alert"` on the error line.

**Declined, with reason**

- **Confirm before disconnect** (`CloudBackupCard`): disconnect was unconfirmed before this PR as well, so this is a behaviour change rather than a regression introduced by the redesign. Worth doing, but it belongs in its own change where it can be applied consistently to the other disconnect paths too.

PR #1443, opened 2026-08-06
feat(ux): Bucket A defaults pass: remove choices the system already knows the answer to

jakobwennberg-oss:
Ships Bucket A of the 2026-08-06 choice-removal audit (dev_docs/choice_removal_audit.md): the all-lens-clean removals where the code already computes the answer but asks anyway, plus the moms_period silent-failure fix.

## What changes

**Batch booking** (`BatchCategorySelector`, `BulkBookInboxDialog`)
- VAT no longer resets to a hardcoded `standard_25`: a new default "Enligt kategori" option omits the treatment so the server derives it per category (exempt bank/card fees, 12% representation). This was an active moms-error source: 25% claimed on momsfria bankavgifter unless caught every time.
- An explicit "Ingen moms" now books `exempt` instead of silently collapsing to the derived default (which meant 25% on most expense categories).
- Reverse charge is never derived, exactly as before.

**Fiscal period is derived, not picked** (`JournalEntryForm` embedded, `BookDirectlyDialog`)
- The period is a total function of the entry date (Swedish periods never overlap); both surfaces now render it as text. The embedded Select allowed picking a period that disagreed with the date; BookDirectlyDialog's silent `periods[0]` fallback (an actual wrong-period booking bug) is now a blocking explanation.

**Cost-account prefill from the company's own history** (`BookDirectlyDialog`, `NewSupplierInvoiceForm`, new `?counterparty=` mode on `GET /api/settings/counterparty-templates`)
- The deliberately-blank cost account prefills from a counterparty-template hit (same tiered matcher the booking flows use) or `supplier.default_expense_account`, through `handleAccountChange` so konto-default moms rides along. Expense-shaped templates only (P&L 4xxx-8xxx debit, 19xx credit), confidence >= 0.5, never a generic seed (the old 5010 incident is the reason there is no fallback), provenance line shown, supplier switches un-plant.

**Single-company login skips the picker** (`/select-company`)
- Exactly one membership + no new TIC engagements + no pending invite + fresh enrichment: auto-forward to `/`. The in-app "Lägg till företag" links pass `?choose=1` so the picker stays deliberately reachable. Byrå/multi-company users untouched. (The review workflow caught that `setActiveCompany`'s cookie write throws during Server Component render, which would have shipped this silently dead; the legacy write-only cookie hint is now best-effort.)

**Filing surfaces default to the period you can actually file**
- VAT declaration opens on the declaration that is actually open: monthly tracks the real 12th/17th deadline (M-2 until it passes, then M-1; over-40M always M-1), quarterly opens on the last ended quarter (`lib/vat/period-defaults.ts`, tested). Helårsmoms `FyPicker` gains `preferLatestEnded`.
- SIE export left unchanged on purpose: `FiscalYearSelector` already auto-selects after load; the audit's "opens empty" was fetch latency.

**moms_period silent failure**
- A VAT-registered company without `moms_period` gets ZERO VAT deadlines, silently (`deadline-config.ts` needs an exact match). The declaration page's dead-end now collects the answer inline through the same `PUT /api/settings` validation (gated on `vat_number` being present, since the coherence rule would 400 otherwise).

**Granskning** (`/pending`)
- The Godkänn pill commits directly for low/medium risk; high risk keeps the ConfirmationDialog whose warning carries real information. **Note for review:** this is a deliberate deviation from locked convention 10 on this one surface, on the argument that the review row IS the confirm-up-front surface; if you disagree, say so and the dialog comes back with one line. Chat-side "Godkänn alla N" is deferred: it needs ApprovalCard's state lifted (assistant-redesign seam 8.8), not a button.

## Verification
- Full unit suite: 12,959 passed. New tests: `lib/vat/period-defaults` (deadline-aware cases) and the `?counterparty=` route mode (401/list/match/null/400).
- `check:guards`, `check:lint` clean; production build passes.
- 17-agent adversarial review over the diff: 13 confirmed findings, 10 fixed in-branch (incl. the dead auto-forward), 2 deferred as follow-ups (AiFilledIndicator generalization, BulkBookInboxDialog whole-file i18n), 1 flagged above (convention 10).

No migrations. Not visually signed off: the changed surfaces (batch dialogs, Bokför direkt, momsdeklaration empty state, /pending pills) want a founder look before merge.

## Summary by CodeRabbit

* **New Features**
  * Added automatic VAT treatment options for transaction and bulk-booking workflows.
  * Added account suggestions based on previous counterparty and supplier bookings.
  * Added inline VAT reporting-period setup with smarter latest-ended-period defaults.
  * Added counterparty template lookup and matching.
  * Automatically select a sole eligible company during onboarding, with an option to choose manually.
* **Improvements**
  * Low- and medium-risk operations can be approved directly; high-risk actions retain confirmation.
  * Fiscal periods now derive from entry dates and appear as read-only.
  * Company assignment remains reliable if cookie updates fail.
* **Documentation**
  * Added decision-log entries and expanded automated coverage.


jakobwennberg-oss:
Review triage (f044770e):

**Fixed**
- Stale bulk selection after direct commit: `commitOp` now prunes the op id from `selectedIds` (pending/page.tsx:871 finding).
- Detail-panel Godkänn now has the same risk gate as the row pill: low/medium commit directly, high keeps the dialog (pending/page.tsx:1443 finding).
- Accounts-not-loaded race (NewSupplierInvoiceForm:655, the real catch of this round): automatic account fills are now *requested* and applied by an effect that waits for both the BAS chart and the request with fresh closures, so a prefill can no longer land before the chart loads and leave a VAT-free konto on the 25% row default.
- Timezone-dependent test dates: local-time `Date(y, m, d)` constructors; suite verified under `TZ=America/New_York`.
- Stale "ML 11 kap 8 §" citation dropped from the views comment (compliance bot). The same citation inside the pre-existing `/api/settings` error string is untouched here: separate cleanup.

**Dismissed, with reasons**
- *Duplicate destructuring "cannot compile" (marked Critical)*: false: each `const { status, body }` sits in its own `it` block scope. The file type-checks in the project build and its tests run green in CI.
- *Localize new strings (route 400, BookDirectlyDialog, BulkBookInboxDialog, VAT-view copy)*: these files are hardcoded-Swedish throughout today (BookDirectlyDialog, BulkBookInboxDialog and the reports views have no i18n at all; the route 400 is a programmatic guard no UI surfaces). Localizing only the new lines would leave mixed-language files; the whole-file migrations are recorded as follow-ups in DECISIONS.md.
- *FyPicker `preferLatestEnded` falls back to newest started when nothing ended + `toISOString` day-boundary*: the fallback preserves the exact pre-branch behavior for first-year companies (an empty picker would be strictly worse for them), and the date comparison matches the file's existing `preparePeriods` convention: changing one line would leave the file internally inconsistent. Both noted for a follow-up sweep.

Also tracking from the compliance bot's notes: the `'none'`→derived-VAT collapse still exists in QuickReviewDialog/CategoryExpandedDialog (documented follow-up), and the `/^[4-8]/` prefill gate admits 8xxx financial-income accounts in theory (matcher is name-based, low probability; tighten if observed).

PR #1445, opened 2026-08-06
fix(import): treat a voucher-less SIE file as a no-op, not a failed migration

jakobwennberg-oss:
## The incident

A Fortnox migration for a new account (CashLeads Media AB, 2026-08-06 21:38 CEST) died with the generic "Migreringen misslyckades / Något gick fel. Försök igen." after the bookkeeping had in fact imported: 2024 (2,098 verifikationer + IB) and 2025 (2,055) completed, but the 2026 file contained **zero vouchers** (nothing booked in Fortnox this year), the finalizer's 0-entry safety net flipped it to `failed`, the wizard aborted before the customers/suppliers/invoices phase, and the real Swedish error text was swallowed by `getErrorMessage`'s pattern heuristic on the way to the UI.

## Fixes

1. **`lib/import/sie-import.ts`**: `finalizeImportRecord` only downgrades a 0-entry run to `failed` when the file actually contained vouchers (parsed count travels in the `MigrationDocumentation` object). A voucher-less file completes as a no-op with the warning "SIE-filen innehåller inga verifikationer för räkenskapsåret...". The Lookma-case retry loop (vouchers existed, all skipped on unmapped accounts) is unchanged, including when no documentation object is passed.
2. **`ArcimMigrationWorkspace.tsx`**: messages that are already user-facing Swedish (server envelopes via `apiError()`, `ImportResult.errors` via `UserFacingError`) now reach the result step verbatim instead of being re-tested against the Swedish-pattern heuristic and collapsing to the generic fallback.
3. **`lib/errors/get-error-message.ts`**: the heuristic learns the import-error family (`verifikation`, `importera|importen`) so other surfaces rethrowing engine messages keep the real reason too.

One pinned behavior deliberately changed: a balance-only file whose IB is skipped by the continuation guard (company already has posted entries) now also finalizes as a no-op success instead of `failed`, because re-running the identical file can never produce a different outcome and the old error text ("granska varningarna ... konton som behöver mappas") pointed at the wrong cause. Genuinely unmapped balance-only files are still refused at stage time by the coverage guard. Logged in DECISIONS.md.

## Verification

- `npm test`: 12,957 passed (full suite)
- `npm run lint`: 0 errors
- New tests: empty-file no-op keeps `completed` + warning; voucher-skipping file still flips to `failed`; the guard sentence passes `getErrorMessage` verbatim

## Recovery for the affected account

After this deploys, "Försök igen" in the wizard completes the migration: 2024/2025 re-import in replace-mode (no user-created entries exist), 2026 passes as an empty no-op, then the API phase (customers/suppliers/invoices) runs and the consent is accepted.

## Summary by CodeRabbit

- **Bug Fixes**
  - Empty SIE files now complete successfully as no-op imports instead of failing.
  - Imports with declared or unparseable vouchers continue to report failures.
  - Balance-only continuation cases no longer fail unnecessarily.
  - Migration errors now preserve clear, detailed messages, including Swedish validation and import messages.
  - Added a warning when an empty import completes without creating entries.
  - Zero-entry import errors are now shown with their original detailed message instead of a generic error.


jakobwennberg-oss:
Triage of the Swedish compliance review findings (40fee026):

1. **Truncated/corrupted file masked as empty: fixed.** `finalizeImportRecord` now cross-checks the raw file content: the no-op path only engages when the content declares no `#VER` records at all (`/^\s*#VER\b/m`). A file whose vouchers were swallowed by a separator/encoding mismatch (the parser only warns in that case, it does not error) keeps flipping to `failed`, and its guard message points at the parser's "#VER-rader hittades men inga verifikationer kunde tolkas" warning. Pinned by a new test.

2. **Behandlingshistorik for the no-op: already durable, no change.** The reasoning is persisted in three places tied to the import record: the `sie_imports` row itself (`status='completed'`, `transactions_count=0`), the `migration_documentation` JSONB (systemdokumentation per BFNAR 2013:2, carrying `vouchers.total: 0`), and, because the import now completes, the source SIE file is archived to storage under the 7-year retention path, so the emptiness claim is verifiable against the original file. The warning string is UI messaging on top of that record, not the record itself.

jakobwennberg-oss:
Triage of the round-2 review comments (3a2c3b15):

- **CodeRabbit / verbatim envelope messages (major): declined by design.** Routing extracted server messages back through `getErrorMessage`'s heuristic is exactly the mechanism that produced this incident: a curated Swedish reason got replaced by the generic fallback. Extension routes return user-facing Swedish messages by repo convention ("User-facing errors are Swedish"); a technical string in an envelope's `message` field is a route bug to fix at the route, not something the wizard should mask. Scope is deliberately limited to the migration wizard's failure surface.
- **CodeRabbit / DECISIONS.md missing the raw #VER condition (minor): fixed** in 3a2c3b15.
- **Compliance / partial-parse scenario reaching the no-op branch: unreachable by construction.** The branch is gated on `journalEntriesCreated === 0`; any run that imported at least one voucher can never enter it. Partially dropped vouchers keep flowing through the pre-existing skipped-voucher accounting (warnings + `details.skippedVouchers`), unchanged by this PR.
- **Compliance / warnings visibility for the continuation-guard no-op: already satisfied.** Both the migration wizard and the manual import result step render `result.warnings` in the amber warning card alongside the success state, and the machine-readable reason (`vouchers.total`, IB skip) persists in `migration_documentation`.

PR #1450, opened 2026-08-07
fix(booking): explicit Ingen moms in quick review books exempt, not the derived 25%

jakobwennberg-oss:
Closes the remaining half of the 'none'-collapse gap the Swedish compliance bot flagged on #1443.

**The bug:** QuickReviewDialog collapsed the UI sentinel `'none'` ("Ingen moms") to `undefined` on the wire. The server then re-derived the category default, so a user who explicitly chose no VAT on a 25%-default category got 25% moms booked anyway — while the dialog's preview showed no VAT line. The dialog's auto-`'none'` for class-2 account overrides hit the same collapse, booking a stray moms line on liability re-bookings.

**The fix (deviation test, not a blanket map):**
- `'none'` as the *seeded default* (exempt categories like bank fees, templates without a treatment) stays off the wire: untouched bookings remain byte-identical to before.
- `'none'` as a *user deviation* goes as explicit `'exempt'`: books no VAT line, records the classification the momsdeklaration should see (per `buildMappingResultFromCategory`'s own note), and lands income on 3004 (momsfri) instead of a rate-bearing revenue account.

Logic lives in `resolveExplicitVat` (`lib/transactions/quick-review-defaults.ts`) with tests covering all three cases. CategoryExpandedDialog carries the same collapse but has zero references (dead code) — left for a hygiene sweep, documented in DECISIONS.md.

Verification: helper + transactions suites green (227 tests), `check:lint`, `check:guards` clean. No migrations, no API changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved VAT handling in Quick Review.
  * Existing bank-fee bookings with no VAT remain unchanged.
  * Explicitly selecting a no-VAT option now records the exemption correctly.
  * Selected VAT treatments are preserved during confirmation.
  * Prevented unintended VAT changes when confirming transactions with seeded defaults.
* **Tests**
  * Added regression coverage for default VAT handling and explicit user changes.


jakobwennberg-oss:
Compliance-bot follow-up: verified, not just asserted. `grep -rn "CategoryExpandedDialog"` and `grep -rn "CategoryExpanded"` across `app/`, `components/`, `lib/`, `extensions/` return zero hits outside the component's own file — no static imports, no dynamic imports, no string references. The component is unreachable; no user can hit its residual collapse. Its deletion is queued as hygiene (choice-removal audit follow-up list) rather than widening this diff.

PR #1455, opened 2026-08-08
fix(migration): Visma pagination + chunk-insert resilience (the '300 misslyckades' case)

jakobwennberg-oss:
## What

Two fixes behind the ElvaSmultron support case where a Visma/Spiris migration imported customers and suppliers but reported **300 misslyckades** for both sales and supplier invoices.

### 1. Visma eAccounting pagination was broken (`lib/providers/visma/client.ts`)
The client paginated with OData `$top`/`$skip`, which eAccounting **silently ignores** (its convention is `$page`/`$pagesize`, default 50, max 1000, per developer.vismaonline.com/docs/pagination). Every request returned page 1, and `getPaginated` appended it `Meta.TotalNumberOfPages` times.

Verified in prod: the affected company has 50 customer names inserted exactly 3x each (151 rows, 51 distinct); both invoice fetches returned exactly 6 x 50 = 300 rows of first-page repeats. Beyond duplicates, **collections larger than one page were silently truncated**: only the first 50 of ~150 customers ever came over.

Also added: stop when a page comes back empty, so a buggy/stale `Meta` can never loop or duplicate again.

### 2. Chunked entity inserts were all-or-nothing (`arcim-migration`)
One PostgREST statement per 500-row chunk means one bad row (here: intra-batch duplicate invoice numbers -> `23505` on `idx_invoices_company_invoice_number`) rejects the entire chunk, and the UI reported the chunk size as the failure count with a green icon and no error cause.

- in-run dedupe of customers/suppliers (org-number-else-name key) and invoices (invoice number / supplier+number key) before insert
- `insertWithPerRowFallback`: bulk insert fast path; on failure, per-row retry so healthy rows land and the offending row's DB error is captured
- empty invoice numbers now stored as NULL ('' collides on the partial unique index; NULLs are distinct)
- sales invoices now set `remaining_amount` (NOT NULL DEFAULT 0), so open migrated invoices no longer land as fully settled in AR aging
- result UI: all-failed steps get an error icon and show the first DB error message; arrival-number rollback removed (partial success makes rewinding hand out numbers that already landed; gaps in ankomstnummer are acceptable, unlike verifikationsnummer)

## Tests
- `lib/providers/visma/__tests__/client.test.ts` (new): $page/$pagesize sent, full traversal, empty-page stop
- `extensions/.../insert-fallback.test.ts` (new): bulk fast path, per-row fallback attribution, empty input
- `extensions/.../entity-mapper-invoice-numbers.test.ts` (new): NULL invoice numbers, remaining_amount
- Full suite for both dirs green (83 tests), lint clean, tsc clean for touched files

## Support context
User abe27581 (ElvaSmultron AB) can re-run the migration after deploy; correct totals should then be fetched and imported. Their existing 100 duplicate customer rows from the broken run are untouched by this PR (they plan to restart with a fresh company anyway).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved migration reliability by continuing successful imports when individual records fail.
  * Migration results now clearly identify completely failed entities and show failure counts with sample error details.
  * Prevented duplicate customer, supplier, and invoice records during a migration.
  * Corrected invoice number and payment balance handling, including empty invoice numbers and paid invoices.
  * Improved Visma data retrieval to consistently process all available pages without duplicates or premature termination.

### DIFF
--- a/components/extensions/general/ArcimMigrationWorkspace.tsx
+++ b/components/extensions/general/ArcimMigrationWorkspace.tsx
@@ -120,10 +120,10 @@ interface SkipReasons {
 
 interface MigrationResults {
   companyInfo?: { imported: boolean }
-  customers?: { total: number; imported: number; updated?: number; skipped: number; skipReasons?: SkipReasons }
-  suppliers?: { total: number; imported: number; skipped: number; skipReasons?: SkipReasons }
-  salesInvoices?: { total: number; imported: number; skipped: number; skipReasons?: SkipReasons }
-  supplierInvoices?: { total: number; imported: number; skipped: number; skipReasons?: SkipReasons }
+  customers?: { total: number; imported: number; updated?: number; skipped: number; skipReasons?: SkipReasons; errorSample?: string }
+  suppliers?: { total: number; imported: number; skipped: number; skipReasons?: SkipReasons; errorSample?: string }
+  salesInvoices?: { total: number; imported: number; skipped: number; skipReasons?: SkipReasons; errorSample?: string }
+  supplierInvoices?: { total: number; imported: number; skipped: number; skipReasons?: SkipReasons; errorSample?: string }
 }
 import AccountMappingStep from '@/components/import/AccountMappingStep'
 import type { AccountMapping, ImportResult, ParsedSIEFile } from '@/lib/import/types'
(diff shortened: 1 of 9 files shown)


jakobwennberg-oss:
Review triage:

- **Fixed** (pushed): stub-insert casualties now counted as `failed` (not `noMatch`) so the result row shows error status + the DB message; enrichment failures now feed `errorSample`.
- **Dismissed, RETURNING order**: the index pairing after a bulk `INSERT ... RETURNING` is the codebase's pre-existing pattern (the old orchestrator paired `insertedRows[i]` with `batch[i]` the same way). Postgres preserves input order for `INSERT ... VALUES` in practice (ModifyTable is not parallelized); a natural-key correlation rewrite would have to disambiguate duplicate names and carries more regression risk than the theoretical reorder. If we ever want it airtight, the fix is an explicit ordinal column in the payload, tracked as future hardening.
- **Dismissed, beforeEach reset nit**: the new tests use per-test local mocks / spyOn+restore, matching the merged idiom in `lib/providers/briox/__tests__/client.test.ts` and the pure entity-mapper suites.
- **Compliance swarm (A.8.28, advisory)**: the surfaced string is PostgREST's `error.message` first line (constraint name, no key values; `.details` is never shown), truncated to 140 chars, shown only to the company's own members in their migration result. Intentional: it is exactly what turns '300 misslyckades' into a diagnosable report.
- **Swedish compliance bot**: entity tables (customers/suppliers/invoices) are not journal writes; bookkeeping still flows exclusively through the SIE import engine and `commit_journal_entry`. Pre-existing architecture, unchanged by this PR.

Issue #1444, opened 2026-08-06
Log trusted-role committed_at overrides to behandlingshistorik


mattssonn:
Follow-up from the Swedish accounting review on PR #1439.

Migration 20260806160000 lets backend writers (service_role, direct SQL) preserve a preset committed_at on the draft-to-posted transition, which the seeding flows use to backdate demo history. Nothing records that the transition timestamp was overridden: BFNAR 2013:2 kap 8 wants a log of who did what, when. If a trusted-role connection is ever used against a live company, there is no audit row showing real vs. preset commit time.

Suggested shape: on preserve, write an audit row (actor role, entry id, preset value, wall-clock now()) to a durable table, e.g. the existing audit_log.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1490, opened 2026-08-10
feat(audit): log trusted-role committed_at overrides to behandlingshistorik

jakobwennberg-oss:
## Summary

Implements #1444 (follow-up from the Swedish accounting review on PR #1439).

Migration `20260806160000` lets trusted writers (service_role, or no JWT claims: direct SQL, maintenance, pg tests) preserve a preset `committed_at` on the draft-to-posted transition, which the seeding flows use to backdate demo history. Nothing recorded that the transition timestamp was overridden. BFNAR 2013:2 kap 8 wants behandlingshistorik showing who did what, when: if a trusted-role connection is ever used against a live company, there must be a durable row separating real from preset commit time.

**What the new migration (`20260810121000`) does:**

1. `audit_log` accepts a dedicated action `COMMITTED_AT_OVERRIDE` (reusing `SECURITY_EVENT` would put attack semantics on a sanctioned backdate). The TS `AuditAction` union and the audit-trail Zod filter learn it too.
2. A `SECURITY DEFINER` writer `log_committed_at_override()` inserts the row: entry id, acting user, preset value, wall-clock `now()`, and the JWT role, plus the `gnubok.actor_type`/`actor_label` GUCs. It cannot be a plain INSERT in the trigger: `audit_log` has RLS with no INSERT policy, and `SET ROLE service_role` in the pg harness has no BYPASSRLS. EXECUTE is revoked from `anon`/`authenticated` so PostgREST cannot expose it as an audit-noise RPC.
3. `set_committed_at()` calls the writer in its preserve branch. The stamping branch is byte-identical to `20260806160000`; end-user posts still always get the tamper-proof `now()` stamp and produce no override row.

## Test plan
- [x] Full `pg-real` suite against a fresh supabase/postgres:15.8.1.060 replay of all migrations: 1083 passed. The only 2 failures (`get_unlinked_1930_lines` date filter, `invoice-deliveries` preparing-attempt) also fail on an unmodified-main baseline replay in the same local environment, i.e. pre-existing local flakes, not this change.
- [x] New pg tests: override row written for postgres and service_role preserves (content asserted), no row on the member-stamp and no-preset paths, writer not executable by client roles.
- [x] `npx vitest run lib/api` (472 passed), `npm run lint` (0 errors)

## Summary by CodeRabbit

- **New Features**
  - Added `COMMITTED_AT_OVERRIDE` audit events when a trusted process preserves a preset commit timestamp.
  - Audit records include the affected entry and preserved timestamp details.
  - Audit trail filters now support searching for committed timestamp override events.

- **Bug Fixes**
  - Draft-to-posted entries now receive the current timestamp unless a trusted workflow explicitly preserves an existing value.
  - Restricted timestamp override recording to authorized application workflows.


jakobwennberg-oss:
Triage of the compliance-review findings: all three target the preserve mechanism itself, which shipped in #1439 (migrations 20260806150000/160000) — this PR adds the audit trail #1444 asked for and strictly improves observability of that mechanism. The hardening suggestions (explicit seed-intent gating instead of role inference, stray-preset handling, stronger operator attribution on the no-claims path) are now tracked in #1495 for a founder call. Note on attribution: the no-claims audit row is distinguishable in-row today via actor_type='system' + new_state.jwt_role='none'; user_id anchors the row to the entry owner, which is also what makes it visible under audit_log's owner-scoped SELECT policy.

PR #1528, opened 2026-08-12
feat(documents): dedupe intake channels on content, not just provenance

jakobwennberg-oss:
## What

Prerequisite PR 1 of the flows build plan (§6, "substrate before any flows UI"): content-level dedupe on document ingest. Every ingestion path already computes and stores `sha256_hash` — but only WhatsApp ever read it back. The manual upload had **zero** dedupe, and Resend inbound + mail hunt dedupe on provenance keys only, so the same receipt forwarded to two inboxes, re-hunted by a sweep, or uploaded twice became a second archived document and a second Underlag row. With the hunt live (392 items in 4 days across three channels), that's the plan's "unbounded duplicate generator".

## How

- **`uploadDocument` gains opt-in `dedupeByContent`**: before storing, look up a current-version document in the same company with the same SHA-256 and return it (marked `deduplicated`) instead of archiving a copy. **Deliberately NOT a DB unique index**: archival callers (sent invoices, filings, bank exports) legitimately store repeating bytes. The SELECT-then-insert race is accepted exactly as in the WhatsApp precedent (loser stores a copy; nothing corrupts). Decision logged.
- **`uploadAndExtract` enables it for every inbox channel.** On a hit: adopt the oldest inbox item for that document (callers always get a real `inbox_item_id` — the WhatsApp caller feeds it straight into `updateItemContext`), or fall through and file an item against the *existing* document when the content entered the archive outside the inbox. `DocumentDuplicateSkipped` lands in processing history.
- **Mail hunt (`ingest.ts`) skips outright**: its provenance key catches the same message re-hunted; the content check catches the same receipt arriving through *another* inbox (forwards are common — the plan's explicit requirement).
- **WhatsApp untouched**: its own pre-check also drives the duplicate reply to the sender.
- **No migration**: hash column + index exist since the original archive schema.

## Tests

3 new: dedupe hit returns existing doc, touches no storage, fires no `document.uploaded` (so no re-extraction); dedupe miss stores normally; mail-hunt skips filing on deduplicated content. 456/456 pass across documents, receipt-hunt, invoice-inbox, whatsapp-inbox. tsc/eslint clean on the diff (5 pre-existing tsc errors verified present on clean main).

## Summary by CodeRabbit

* **New Features**
  * Added optional content-based deduplication for uploaded documents.
  * Matching uploads now reuse existing documents or inbox items instead of creating duplicates.
  * Duplicate mail-hunted attachments are skipped automatically.
  * Duplicate activity is recorded for visibility.

* **Bug Fixes**
  * Prevented redundant storage, extraction, and inbox entries for duplicate document content.
  * Existing archival uploads remain unaffected.
  * Duplicate detection failures now stop processing safely rather than creating potentially duplicate records.


jakobwennberg-oss:
Review triage, round 2: applied the payload-PII fix (mailbox removed; note the mechanism differs from the bot's claim — the PII validator only matches personnummer/orgnr digit shapes, so an email would NOT have been rejected; the fix enforces the module's documented contract instead) and the test assertions on the DocumentDuplicateSkipped trail incl. a history-outage failure path. **Skipped** "propagate duplicate inbox-item lookup failures": ingestMailCandidate's documented contract is per-attachment resilience ("a single unreadable attachment must not abort a night's hunt"); the throw is caught by that boundary by design, fail-closed holds (nothing is filed), the error is warn-logged, and the next sweep retries because no item exists under the provenance key.

PR #1544, opened 2026-08-12
feat(inbox): visible differens + double-click balance fill in Bokför direkt

mattssonn:
# What and why

User feedback (screenshot of an unbalanced USD reverse-charge booking): when hand-computing VAT in the inbox "Bokför direkt" dialog, the remaining debit/credit gap had to be worked out by hand; the existing indicator was tiny, muted, and below the fold.

This brings the dialog to parity with the main verifikat editor (`JournalEntryForm`):

- **Differens in the Summa row**: while the entry is unbalanced, the remaining gap renders red left of "Summa" and both sum figures turn red. The old small "Diff" line below the table is removed (one indicator).
- **Double-click balance fill** (ported from `JournalEntryForm`, same opt-in semantics): double-clicking a debit/credit field fills the amount that balances the entry; no-op when balanced or when the fill belongs on the other side. A hint line advertises the gesture while unbalanced.
- `roundOre` used for the diff math per the antipattern guard.

Display + client-side prefill only; the engine and server validation still reject unbalanced entries. Auto-populating on account pick was deliberately NOT done: `JournalEntryForm` tried it and reverted it as surprising when splitting across rows (see code comment there); the double-click gesture is the established pattern.

Migrations: none.
UI strings: hard-coded Swedish, consistent with this dialog's Swedish-only surface (journal editor category per i18n rules), so no `messages/*` keys.

## Checklist

- [x] `npm run check:lint` and `npm run check:guards` pass locally (component change only; no lib/api logic)
- [ ] New or changed logic in `lib/` or `app/api/` has tests (n/a: client component only, repo has no component tests)
- [ ] New or changed UI strings exist in both `messages/sv.json` and `messages/en.json` (n/a: Swedish-only surface)
- [x] Migrations (if any): none
- [ ] Commits are signed off


mattssonn:
@coderabbitai review

mattssonn:
@coderabbitai review

PR #1546, opened 2026-08-12
fix(inbox): carry the matched transaction amount into manual booking

mattssonn:
# What and why

User-reported regression (2026-08-12, same day as #1524): after matching an inbox document to a bank transaction and pressing "Bokfor manuellt", the Andra kontering dialog opened with two blank rows and no amount. #1524 swapped the matched-state manual path from BookDirectlyDialog (which seeded lines from the matched transaction's SEK amount via buildPrefillLines) to EditKonteringDialog, which is seeded only from the booking proposal, and an unknown supplier has no proposal. On a foreign-currency invoice the SEK figure then existed nowhere on screen.

This restores the old prefill semantics inside the new flow:

- `suggest-booking` now returns, on every empty-proposal branch (`no_mapping`, `currency_unsupported`, engine failure), the matched bank row's SEK amount and date plus `fallback_lines`: a balanced two-row skeleton (settlement account on one side, blank cost row for the user to pick on the other), mirroring what buildPrefillLines seeded before the swap. No VAT split, same as the old matched-transaction path.
- The SEK amount goes through `resolveSekAmountOrNull`: a foreign row with neither a stored SEK value nor a rate still opens blank instead of relabeling EUR as SEK.
- The dialog shows the matched transaction's amount and date beside the title, so the kronor figure stays visible even if the user clears the rows.
- Empty proposals now carry `entry_date` = the bank date, so the dialog no longer falls back to the document date (BFL 5 kap: the day the money moved).

Read-only advisory surface: nothing posts without the form's own balance check and confirm step. Migrations: none.

Tests: new `fallback-kontering.test.ts` (6 cases: direction, foreign SEK resolution, refuse-to-guess, ore rounding) and three new/extended cases in `suggest-booking.test.ts` covering all three empty branches. 29 tests green locally; lint 0 errors; check:guards passes; tsc clean on changed files.

## Checklist

- [x] `npm run lint` and targeted `npx vitest run extensions/general/invoice-inbox` pass locally (full `npm test` has the known pre-existing Windows-local SEK-label suite failure, green on CI)
- [x] New or changed logic has tests
- [x] Migrations: none
- [x] Core builds with zero extensions enabled (no core imports from `@/extensions/`; the new helper lives inside the extension)
- [ ] Commits are signed off


mattssonn:
@coderabbitai review

mattssonn:
@coderabbitai review

PR #1549, opened 2026-08-12
fix(providers): stop requesting unapproved Fortnox scopes that broke every connect

mattssonn:
# What and why

PR #1541 (voucher attachment import) added `archive` and `connectfile` to the Fortnox `DEFAULT_SCOPES`. The registered Fortnox app does not have those scopes approved in the Fortnox Developer Portal, so Fortnox rejects the authorize request with `invalid_scope` ("An unsupported scope was requested") before the user can even log in. This broke EVERY Fortnox connect in production within minutes of the #1541 deploy; confirmed in Vercel runtime logs (three failed attempts by a real user at 22:45-22:48 UTC on 2026-08-12, deployment `dpl_ApBP1JfgFP4myB51SzsuQBjMaG7J`).

This PR:

- Removes `archive`/`connectfile` from the connect-time scope request, with a comment and a pinning test so they are not re-added until the portal registration has them approved.
- Keeps ALL of #1541's attachment-import logic intact. The feature already degrades gracefully: a Fortnox 403 becomes `PROVIDER_DOCUMENT_SCOPES_REQUIRED` and the wizard shows the reconnect follow-up card instead of failing the migration.
- Adds `charset=utf-8` to the OAuth callback HTML responses (error and success pages). Without it browsers parse the UTF-8 bytes as Latin-1 and render Swedish text as mojibake ("rÃ¤tt behÃ¶righeter"), which is also what the toast in the wizard re-displays.

Known interim behavior, deliberate: until the Fortnox portal app has the two scopes approved (and they are re-added to `DEFAULT_SCOPES`), the optional "underlag från Fortnox" follow-up after a migration will report missing permissions, and reconnecting will not clear it. That is the cost of restoring the connect flow for everyone; the migration itself completes normally.

Migrations: none.

Test coverage: updated `lib/providers/fortnox/__tests__/oauth.test.ts` to pin the reduced scope set; full Fortnox + arcim-migration + document-import-flow suites pass locally (217 tests).

## Checklist

- [x] `npm run lint` passes locally; affected vitest suites pass (217 tests)
- [x] No new UI strings
- [x] No migrations
- [x] No core imports from `@/extensions/`

## Summary by CodeRabbit

- **Bug Fixes**
  - Fortnox connections no longer fail due to unsupported default permissions.
  - Attachment imports remain available, with clearer handling when access is denied and guidance to reconnect.

- **Improvements**
  - Fortnox now requests only approved default permissions; additional attachment-related access can be enabled separately after approval.
  - OAuth success and error messages now display correctly with UTF-8 encoding.


mattssonn:
@coderabbitai review

PR #1559, opened 2026-08-13
fix(transactions): anchor the tx-pinned document to the verifikat on booking

mattssonn:
## What

Adds a pinned-document leg to the shared booking-time underlag propagation (`lib/transactions/inbox-underlag.ts`). A document pinned directly on a bank transaction (`transactions.document_id`) can exist without any `invoice_inbox_items` row (documents-page link, staged MCP attach), so the inbox-item loop alone never anchored it: booking then produced a verifikat reporting Underlag saknas while the file sat pinned to the transaction (2026-08-13 user report; 72 rows across 10 companies repaired by hand, 42 restored, 30 parked in locked periods).

## Why this shape

The leg lives in `propagateUnderlagForBookedTransaction`, so every booking path gets it at once: /book, bulk-book, categorize, link-to-existing-voucher, and the attach-after-booking entry point. The pin is always re-read inside the helper after booking, never taken from the caller's transaction snapshot: an adversarial review (3-skeptic pass) refuted the caller-supplied variant twice with one root cause, a pre-booking snapshot misses a doc attached mid-booking, and an attach-path opt-out re-opened the gap for bulk-booked transactions where `transactions.journal_entry_id` stays NULL (verifikat hangs off `transaction_voucher_links`). Attach-after-bulk-book now anchors via voucher-link resolution (regression-tested). Never steals: a doc already serving as another verifikat's underlag is left alone (DB immutability trigger backs this at the DB layer); the bulk-book RPC's own doc linking makes the leg an idempotent no-op there.

## Migrations

None.

## Tests

- 6 new tests in `lib/transactions/__tests__/inbox-underlag.test.ts`: pin resolution, never-steal, shared pin+inbox doc links once, pin failure does not block inbox items, attach-after-bulk-book anchors against the samlingsverifikat.
- All transactions suites green (34 files, 532 tests), lint 0 errors, check:guards clean.

## Staging click-test

1. Attach a document from the documents page (not the inbox) to an unbooked bank transaction.
2. Book the transaction with Bokför manuellt.
3. Open the verifikat: the underlag is attached (before: Underlag saknas).

## Known out-of-scope follow-up

The v1 API categorize routes (`app/api/v1/.../categorize`, `batch-categorize`) book without calling the shared helper at all and can still reproduce the bug for API-key clients; needs its own issue.

## Summary by CodeRabbit

- **Bug Fixes**
  - Documents attached to transactions during booking are now reliably linked to the resulting journal entry.
  - Existing document links remain protected from reassignment.
  - Duplicate document links are avoided, including when documents are shared with inbox items.
  - Booking continues successfully if document linking encounters an error.
  - Improved handling for bulk booking and voucher-linked journal entries.


mattssonn:
Duplicate of #1560, which is merged (d9dddba68). Same fix implemented independently; this one also had red CI (executors.test.ts queue desync). Worth salvaging later: the attach-after-bulk-book samlingsverifikat test, and the note that v1 categorize/batch-categorize still bypass the shared helper.

PR #1563, opened 2026-08-13
fix(whatsapp): make every silent disposition observable, stop pure silence for linked senders

### LINKED ISSUE #1552: [WhatsApp] Intake can swallow a receipt in seven places, none of them observable
## Problem

A user reported (2026-08-12) that sending a receipt over WhatsApp produced no reaction at all. Nothing came back, and nothing about the account explained it. The same flow works from other numbers, and the webhook itself is healthy: 22 inbound requests in 24 hours, all 200.

The reason it could not be diagnosed is the actual bug. Silence is a legitimate, deliberate outcome in at least seven places, and none of them leave a trace anyone can look up for a given sender:

- muted link: after STOP, every message except START is silence, photos included (`index.ts:257`, `:287`)
- unknown sender over the pre-binding quota: `check_and_increment_whatsapp_sender_quota` returns not-ok and we return (`index.ts:143`)
- quota RPC itself failing: fails closed and stays silent by design (`index.ts:138-142`)
- unknown sender whose greeting is throttled: `greetingThrottled()` caps the M1 reply per hour and per day (`index.ts:100-118`)
- stale interactive tap after the question closed (`index.ts:293`)
- inbound row missing its link or media reference: marked error, no reply (`lib/process-inbound.ts:392-397`)
- link revoked between arrival and processing: marked skipped, no reply (`lib/process-inbound.ts:401-407`)

Every one of these is a defensible product decision on its own. Together they mean the channel can swallow a receipt with no signal to the sender, no signal to the user in the app, and nothing support can query afterwards. For an intake channel where the whole promise is "photograph it and forget it", a swallowed receipt is worse than a rejected one: the user believes the underlag is filed.

Deliberate constraint to design around: for unknown senders we never persist content or raw payloads (`index.ts:120-124`). That is right and must stay. This is about metadata, not messages.

## Solution

Make every silent disposition observable at metadata level, and stop being silent where the sender is known.

1. Record the disposition on every inbound message we decline to answer: sender hash, disposition kind, reason, timestamp. No body, no media, no profile name. Enough for support to answer "what happened to my message" from the phone number alone.
2. For a linked sender, replace the silences that are not deliberate policy with a reply. Muted is deliberate (STOP means stop). A revoked link, a missing media reference, or a dropped media fetch is not: those are failures we should own out loud.
3. Surface the last inbound event in the WhatsApp settings panel: when we last heard from this number and what happened to it. The panel already renders link state and muted (`WhatsAppLinkPanel.tsx:164-166`), so this is the natural place.
4. Alert on the shape that matters: a linked number whose messages all end in silence, which is what this report was and what nothing detected.

## Why

- The reported case is still unexplained. We cannot tell the user why, and with today's code we could not have told them at any point after the fact.
- WhatsApp intake is being positioned as the low-friction way to file underlag. A channel that can silently drop a receipt undermines exactly that promise, and the user has no way to notice.
- The same blind spot cost a day of guessing on the upload half of this report (#1550, #1551): the cause turned out to be measurable in minutes once there was something to measure.

## Where

**Files:**
- `extensions/general/whatsapp-inbox/index.ts:244-300`: `classify()`, where the silence dispositions are decided
- `extensions/general/whatsapp-inbox/index.ts:129-143`: unknown-sender quota, both the over-quota and the fail-closed returns
- `extensions/general/whatsapp-inbox/index.ts:100-118`: `greetingThrottled()`
- `extensions/general/whatsapp-inbox/lib/process-inbound.ts:392-407`: missing link/media and revoked link, already writing a status that nothing reads back
- `components/extensions/general/WhatsAppLinkPanel.tsx:164`: the settings surface that should show the last inbound event

**Related:**
- `whatsapp_messages.processing_status` already carries some of this; the gap is that unknown senders and classify-time silences never reach a row at all
- Retention: whatever is recorded must expire with the existing WhatsApp retention cron, and must not become a way to reconstruct message content

## Acceptance

- For any phone number, support can answer "we received a message at T and did X with it, for reason R" without reading message content.
- A linked sender never gets pure silence except when muted.
- The settings panel shows when we last heard from the linked number and what happened.
- Nothing new persists message bodies, media, or raw payloads for unknown senders.

Reported alongside the upload failure fixed in #1550. That half turned out to be a platform body limit; this half is unrelated and unexplained.

jakobwennberg-oss:
Closes #1552.

## The reported case, traced as far as the data allows

With the reporter's phone number (shared privately for troubleshooting), prod shows:

- The number IS linked and healthy: active link since 2026-08-07 15:17, not muted, not revoked, `default_company_id` null.
- Their message arrived 2026-08-12 07:23:48 and reached linked-sender processing: the burst debounce was armed, and the ack claim ran at 07:25:45.
- The conversation ended `idle` with an empty context: no company question was ever staged.
- **No `invoice_inbox_items` row exists for it.** The only WhatsApp items since 08-11 belong to other senders.

So the receipt's processing ended in `skipped` or `error`, a notice (M15/M17/M18/duplicate) was attempted (the row had both its link and media references, which excludes the silent pre-reply paths), and that notice never reached the sender. Today, nothing records why: `sendText` never throws, a failed send is a `log.warn` and a `delivery_status='failed'` row with no reason, and Meta's `failed` delivery statuses had their error payload dropped at the parser.

One query closes the specific case (CLI access to `whatsapp_messages` was blocked in-session; run in the SQL editor):

```sql
select direction, message_type, processing_status, delivery_status,
       error_message, attempts, media_mime, created_at
from whatsapp_messages
where phone_link_id = 'bd53f552-a1c5-407f-837b-5d2ed63a2205'
order by created_at desc limit 40;
```

After this PR, that answer is readable without SQL, and the sender is no longer left in silence.

## What changed

1. **Unknown-sender declines persist content-free trace rows** (over quota, quota RPC failing closed, greeting throttled, plus `done` rows for greeted/bad-code replies): wamid, phone hash, message type, disposition. No body, no media reference, no raw payload, no profile name. Capped at 20 rows per hash and day; the existing 30-day unknown-sender retention deletes them. The wamid dedupe also stops a Meta redelivery of an already-answered bad-code/greeting message from earning a second reply.
2. **Linked-sender deliberate silences record their reason** on the skipped row (muted, stale interactive tap, ignorable type).
3. **Non-policy silences now reply**: a row missing its media reference sends M18 through the link's reply address; a link revoked between arrival and processing gets the M1 unlinked copy, throttled like the unknown-sender greeting. Muted stays silent: stopp means stopp.
4. **Send failures keep their reason**: outbound rows store the Graph error detail on failure, and `failed` delivery statuses from Meta store their error code/title instead of being dropped.
5. **The settings panel shows the last inbound event** ("Senaste meddelande: 12 augusti 2026 09:23 · hoppades över: samma fil finns redan") from a closed server-derived enum (`lib/last-event.ts`, so no internal error text reaches the client), plus one attn line when the latest reply never left our side.

## Acceptance (from the issue)

- For any phone number, support can answer "we received a message at T and did X, for reason R" from `whatsapp_messages` alone, without message content. ✔
- A linked sender never gets pure silence except when muted (or when no reply address can be resolved at all, which the row now records). ✔
- The settings panel shows when we last heard from the number and what happened. ✔
- Nothing new persists message bodies, media, or raw payloads for unknown senders. ✔

Not in this PR: the proactive alert on "linked number whose messages all end in silence" (issue idea 4, not in acceptance). The panel surfacing plus recorded reasons are the prerequisite for it.

## Notes

- No migration: declines reuse `whatsapp_messages` (see DECISIONS.md), so there is no orphan-version risk and retention/RLS already apply.
- 190 extension tests green (6 new), lint clean, `check:guards` green, sv/en key parity verified.
- Visual: the new panel row uses the stock `SettingsRow` + `AttnLine` primitives; founder visual sign-off still applies before merge.

## Summary by CodeRabbit

- **New Features**
  - WhatsApp link status now shows the latest inbound event, its timestamp, and whether the latest reply failed.
  - Added clearer translated statuses for message processing, duplicate messages, rate limits, unsupported files, muted chats, and failed replies.
  - Improved responses for missing media and revoked links.

- **Bug Fixes**
  - WhatsApp delivery failures now provide clearer error details.
  - Unknown-sender declines are tracked with deduplication and daily limits without storing message content.


jakobwennberg-oss:
Review triage (af6165ef):

**Fixed**
- greetingThrottled now fails closed when the throttle window cannot be read (CodeRabbit, conversation.ts).
- Decline-trace day cap now applies only to 'skipped' declines; 'done' traces always insert, so the wamid dedupe prevents duplicate M1/M2 replies even past the cap (CodeRabbit, index.ts). Their volume is bounded upstream (greeting throttle 3/day, pre-binding quota 200/day).
- company-question.test.ts mocks match the widened SendTextResult (CodeRabbit, graph-api.ts finding).

**Declined, with reasons**
- Fully atomic cap/dedupe reservation and atomic greeting throttle: both would need a migration + RPC for what are soft observability bounds. Worst case under racing webhook deliveries is a few extra metadata rows or a duplicate greeting, and the greeting race predates this PR. Noted as best-effort in the code comment; not worth the schema surface.
- Event-bus clear in the new test describe: the file uses no event bus anywhere (uploadAndExtract is mocked); adding a clear only to the new block would be inconsistent with the file's three existing describes.

**Compliance bot notes** (advisory, no violations): the revoked-link media drop is the intended consent semantics (media stays re-sendable from the phone; the M1 reply now says how to relink), and the 30-day unknown-sender purge cannot catch rows that later belong to a filed item, because trace rows never gain a phone_link_id or inbox_item_id retroactively; linking always creates new rows.

Issue #546, opened 2026-05-20
Önskemål - Kunna skicka E-faktura via PEPPOL


ZilanCiftci:
Möjligheten att Skicka E-faktura via PEPPOL är önskvärt. Det är en förutsättning för oss som fakturerar mot offentlig sektor (dom kräver att man skickar fakturorna som e-faktura). Inte läst i detalj, men finns t.ex. qvalia.com som har multi-tenant lösningar. Skickar idag mina fakturor via Fortnox, och lite av en förutsättning innan jag går över till er är att det fungerar även här. Vet att det involverar en kostnad för er, och är självklart villig att betala för fakturautskicken (fortnox tar t.ex. 3kr / faktura)

ZilanCiftci:
Jag har inte själv provat qvalia.. Jag satt bara och undersökte själv ett tag hur jag själv skulle kunna implementera e-fakturering från mitt egenbyggda BMS .. kom aldrig till skott.

mattssonn:
PR #1585 implements and documents the mergeable provider-agnostic foundation only: constrained Peppol BIS Billing 3 UBL XML generation, Swedish preflight, authenticated company-scoped export API, and a UI download that explicitly says the invoice was not sent.

#546 must remain open after that PR merges. The exact blocker to send/receive is an external product and commercial decision: select and contract a certified Peppol access-point provider, confirm multitenant/reseller terms and pricing, obtain sandbox and production credentials/API documentation, choose participant registration and lookup behavior, define delivery receipt/webhook semantics, and decide how each company configures its participant identifier. Scheme 0007 can be used only for non-personal organization numbers in the current foundation. Sole traders and any personnummer-derived identifier need a separately configured scheme 0088 GLN so Accounted does not publish personal identity data.

Once those inputs exist, the remaining implementation is provider adapter delivery, release-pinned official XSD/EN 16931/Peppol validation, durable status and immutable audit storage, participant discovery, and a separately scoped inbound flow. No provider credentials or network delivery have been invented in PR #1585.

mattssonn:
Provider-neutral transport foundation: #1595.

[...]

PR #1580, opened 2026-08-13
fix(mcp): correct e-invoice capability guidance

### LINKED ISSUE #1577: MCP skills falsely advertise Peppol / EN 16931 support
## Problem

Two MCP skill texts claim capabilities that Accounted does not implement:

- `extensions/general/mcp-server/skills/invoicing-rules.ts` says Accounted renders EN 16931-compliant XML on demand.
- `extensions/general/mcp-server/skills/customer-onboarding.ts` says Accounted handles Peppol when a customer has an endpoint configured.

The repository has no Peppol send/receive path and no EN 16931 XML renderer. The assistant can therefore promise customers a feature that does not exist.

## Fix

Replace both claims with accurate product guidance: Accounted currently supports PDF/email invoices; users who deliver an e-invoice through an external provider can mark the invoice as sent. Keep the strategic implementation gap tracked in #546.

Add or update a regression test if the MCP skill suite has an appropriate product-truth guard.

## Acceptance criteria

- Neither skill claims that Accounted renders EN 16931 XML or sends/handles Peppol.
- Guidance points users to the supported external-delivery + mark-sent flow.
- Relevant tests, lint, and skill-generation checks pass.

Found while verifying #546.

mattssonn:
## Summary
- replace unsupported Peppol and EN 16931 product claims with the supported external-provider plus mark-sent flow
- correct the same implication in the MCP-exposed Swedish invoice compliance atom and regenerate its registry seed
- add a product-truth guard covering static workflow skills and the exposed invoice atom

#546 remains the implementation gap for native e-invoicing support.

## Verification
- npx vitest run --project unit extensions/general/mcp-server/__tests__/skills.test.ts
- npx vitest run --project unit extensions/general/mcp-server/__tests__/claude-plugin-refs.test.ts lib/bokslut/arsredovisning/__tests__/arsredovisning-k3-pdf.test.ts
- npx eslint extensions/general/mcp-server/__tests__/skills.test.ts extensions/general/mcp-server/skills/invoicing-rules.ts extensions/general/mcp-server/skills/customer-onboarding.ts
- npx --yes --package=tsx tsx scripts/generate-skill-bodies.ts --check
- production migration reconciliation: no remote-only versions; only this new seed is local-only

Full unit suite note: 14,063 tests passed; an unrelated origin/main guard file, scripts/checks/__tests__/format-currency-sek-label.test.ts, fails to parse when run alone. Four load-sensitive 5-second timeout failures passed when isolated.

Closes #1577


mattssonn:
Verified the generated registry seed contains the corrected external-provider plus gnubok_mark_invoice_as_sent guidance (migration line 4527), and npm run skills:check passes in CI. The existing generator freshness guard already compares every atom body hash against the committed manifest/seed workflow, so the focused product-truth test remains on authored active skill text rather than parsing generated SQL.

PR #1582, opened 2026-08-13
fix(salary): skatteavdrag rounding trio (whole kronor, ,50 table pick, import prefix guard)

mattssonn:
Three small statutory rounding fixes in the salary module, follow-ups from the tax table review.

## 1. Percentage skatteavdrag in whole kronor (SFF 2011:1261 22 kap. 1 §)
`calculateJamkningTax` / `calculateSidoinkomstTax` returned öre-precision amounts; skatteavdrag is stated in whole kronor with öretal dropped, the same rule `taxForRate` already applies to percent brackets. Computed in integer öre and hundredths of a percent because flooring the raw float product loses a whole krona when float noise lands an exact result just below an integer (`1000 * 0.007 === 6.999999999999999`, so 0,7 % jämkning of 1 000 kr would floor to 6 kr).

**Scope note:** the two inline flat-30% branches in `calculation-engine.ts` (unverified F-skatt, no-table fallback) had the identical defect and now route through `calculateSidoinkomstTax`. This is two lines beyond the reviewed plan, same statutory rule; logged in DECISIONS.md.

Live impact: at most 1 kr lower withholding per payslip for jämkning/sidoinkomst employees. Already-paid runs are unaffected (amounts are persisted at run time).

## 2. Table selection at exactly ,50
`fetchKommunTaxRates` used `Math.round(rate)`, sending 32,50 to table 33. Skatteverket's rule: a fractional part of at most 50 öre picks the lower table, 51 öre or more the higher (32,49 → 32, 32,50 → 32, 32,51 → 33). Compared in hundredths so float noise cannot decide the boundary. Latent today (no kommun sits exactly on ,50 for 2026).

## 3. Two-week rows fail the monthly import loudly
`scripts/import-tax-tables.ts` only checked position 3 for B/%, so a two-week row (`14B29`) would silently merge into the monthly fallback data if the wrong Skatteverket file were used. The day-count prefix must now be `30`; a 14-row throws. `main()` is guarded behind a direct-execution check so `parseLine` is importable by tests.

## Testing
- New unit tests per case incl. Skatteverket's own ,50/,51 example, the öre-drop boundary 41 999,99 × 25 % → 10 499 (not 10 500), and the float-noise case 0,7 % of 1 000 kr → 7 kr.
- `npx vitest run lib/salary scripts/__tests__/import-tax-tables.test.ts`: 536 passed.
- `npm run lint`: clean.
- Pure functions, no migrations, no data impact; rollback = revert per commit.


mattssonn:
Skeptic pass (3 adversarial agents, author-independent) completed before merge:

- **correctness: REFUTED, then fixed in 521806f53.** Negative taxable income (deductions exceeding pay) hit Math.floor rounding away from zero: calculateSidoinkomstTax(-0.01) returned -1 kr where öretal bortfaller means -0. Now Math.trunc with -0 normalized, plus regression tests.
- **compliance/money: COULD-NOT-REFUTE.** Salary voucher stays exactly balanced (net absorbs the öre); whole-krona 2710 credit now matches AGI FK001 (which already rounds), removing a pre-existing öre residue between ledger and skattekonto.
- **regression: COULD-NOT-REFUTE.** Only caller of the changed functions is calculation-engine.ts; no öre-precision contract anywhere downstream (schemas, MCP, payslip, pain001); fallback regeneration from the committed 2026 TXT is byte-identical; 712 tests across lib/salary + app/api/salary pass.

Residual notes (pre-existing, not this PR): absolute-amount/förhöjt jämkningsbeslut unsupported; jamkning_percentage schema allows >2 decimals which quantize to hundredths; negative withholding itself remains engine-legal as before.

mattssonn:
@coderabbitai review

PR #1583, opened 2026-08-13
fix(payroll): normalize KU10 organisation numbers

### LINKED ISSUE #1410: KU10: does Skatteverket want a 10- or 12-digit organisationsnummer?
`lib/salary/ku/ku10-generator.ts` emits the company's org number into `<Organisationsnummer>` after stripping separators, but **without folding the 12-digit form down to the canonical 10 digits**. A company stored as `165560125790` files with 12 digits; one stored as `5560125790` files with 10.

Every other Skatteverket- and Bolagsverket-bound path normalizes to 10 (or converts deliberately to the 12-digit redovisare form). KU10 is the odd one out.

## Status

Pre-existing behaviour, not introduced by the invariants work (#1364). The old `replace('-', '')` did the same thing. It is now **pinned by a test** so the current behaviour is explicit and a change has to be deliberate:

`lib/invariants/__tests__/org-number-cross-path.test.ts` → *"PINNED: KU10 passes a 12-digit stored org number through unfolded"*

## Why it was not simply fixed

The `swedish-payroll` skill does not cover the KU10 identity digit format. Its reference files were checked; none addresses it. CLAUDE.md forbids answering Swedish domain questions from training data, so the behaviour was pinned rather than guessed at.

The Swedish compliance review bot reached the same conclusion independently on #1364, flagging it as *"unsure, flag for verification against a primary Skatteverket source (KU10/kontrolluppgift schema docs)"*.

## What to check

The KU10 / kontrolluppgift XML schema documentation from Skatteverket: does `<Organisationsnummer>` for the uppgiftslämnare take 10 digits, 12 digits, or either?

## Why it matters

A wrong-length identifier in a kontrolluppgift is a filing-integrity risk: it either fails schema validation at upload or attributes the uppgift to the wrong identity.

## Definition of done

Either confirmation that the current passthrough is correct (in which case add the source to the pinned test's comment so nobody re-opens this), or a fix that normalizes, with the pinned test updated to assert the new behaviour.

mattssonn:
## Summary
- normalize KU10 employer organisation numbers to the 12-digit `16`-prefixed form required by Skatteverket KU schema 12.0
- reject malformed KU organisation identities without adding check-digit validation beyond the XSD
- correct the confirmed FK201 element spelling to `UppgiftslamnarId`
- replace the pinned open question with a sourced cross-path contract and focused generator tests

## Primary source
- Skatteverket: Testtjänst och teknisk beskrivning, KU schema 12.0 and official 2026 KU10 examples

## Verification
- focused KU10 and org-number suites: 22 passed
- changed-file ESLint: passed
- full ESLint: 0 errors
- full Vitest: 14,037 passed; unrelated PDF and skill-discovery tests exceeded fixed timeouts on the shared Windows host, and one unrelated script-check suite hit a runtime load error. CI is the authoritative full-suite run.

Resolves #1410

### DIFF
--- a/DECISIONS.md
+++ b/DECISIONS.md
@@ -3,6 +3,7 @@
 One line per decision: `[YYYY-MM-DD] <decision>: <why>`. Appended by agents and humans when a non-obvious choice is made (approach picked over an alternative, dependency declined, action stopped by a CLAUDE.md rule). Read before re-litigating a past decision.
 
 [2026-08-13] E-invoice product-truth correction covers the MCP workflow skills and the MCP-exposed swedish-invoice-compliance atom: the atom's "for Accounted e-invoice generation" heading made the same unsupported product claim as issue #1577, so all active guidance now directs external delivery followed by gnubok_mark_invoice_as_sent; Peppol implementation remains tracked in #546.
+[2026-08-13] KU10 organisation identities normalize to the 12-digit `16`-prefixed form and receive structural XSD validation without a new Luhn gate: Skatteverket KU schema 12.0 requires that shape, names FK201 `UppgiftslamnarId`, and explicitly leaves check-digit validation outside the schema.
 
 [2026-08-03] Issue #317 derives löneväxling pension and SLP at the shared salary-entry boundary from the frozen salary_runs.calculation_params rate snapshot: this fixes dashboard, MCP, and v1 booking without a schema migration, and prevents a live config change from altering an already reviewed run between calculation and posting.
 
--- a/lib/invariants/__tests__/org-number-cross-path.test.ts
+++ b/lib/invariants/__tests__/org-number-cross-path.test.ts
@@ -22,6 +22,7 @@ import type { IxbrlArsredovisningInput } from '@/lib/bokslut/ixbrl/types'
  */
 
 const AB_10 = '5560125790'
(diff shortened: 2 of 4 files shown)


mattssonn:
Manual primary-source verification completed against Skatteverket's official [Testtjänst och teknisk beskrivning](https://www.skatteverket.se/foretag/skatterochavdrag/kontrolluppgifter/testtjanstochtekniskbeskrivning.4.233f91f71260075abe8800073614.html), specifically its linked KU schema 12.0 ZIP and 2026 KU10 example ZIP.

- `SKV_COMMON_12.0.xsd` defines `OrganisationsnummerTYPE` with maximum length 12 and the structural pattern `16\d{2}[2-9]\d{7}`. Its annotation explicitly states that schema validation does not calculate the check digit.
- `Kontrolluppgifter_COMPONENT_12.0.xsd` defines FK201 as `UppgiftslamnarId`, extending the shared identity type.
- Skatteverket's official 2026 KU10 examples use the 12-digit `16`-prefixed value in the sender, case owner, and FK201 identity positions.

This is why the implementation normalizes 10 digits to `16` plus 10 digits, accepts an already-correct 12-digit value, performs the XSD structural checks, and does not add a new Luhn gate at export time.

Issue #1457, opened 2026-08-08
Per-account VAT treatment (momskod) picker: any konto to the right momsdeklaration ruta


jakobwennberg-oss:
## Problem

Migrating users bring charts of accounts whose momskoder cannot be expressed in Accounted. Support case (Olof, ElvaSmultron AB, migrating from Spiris/Visma):

- Their real sales accounts are 3041 ('Försäljn tjänst 25% sv', 4.27 MSEK lifetime) and 3051 ('Försäljn varor 25% sv'), plus 3043/3052/3053. None are in `ACCOUNT_RUTA`, and after SIE import `default_vat_rate` is NULL, so **ruta 05 underreports** until the user manually sets Standard moms on each account. The #1261 name-inference declines here ('25% sv' does not match '25 % moms').
- Their EU purchases live on 4056-4058 ('Inköp varor 25% EU', Visma convention, 115 postings). There is NO mechanism, manual or automatic, to feed rutor 20-24 from custom purchase accounts: the hardcoded 4515/4535/... list is all there is. **Ruta 20 basis stays 0 while 2614/2645 carry the VAT**, which trips the RC completeness checks.
- They asked explicitly: 'Går det att få till så att man kan sätta momskod på egna konton? Jag ser att det går att sätta moms men inte t.ex. omvänd moms.'

Current state:
- `chart_of_accounts.default_vat_code` is a write-only dead column (accepted by API/MCP, read by nothing).
- `default_vat_rate` does two jobs: booking-line default AND dynamic ruta 05 membership (class 3 only).
- Rutor 07 (VMB) and 08 (frivillig uthyrning) remain unmappable; 41 is fixed for standard BAS 3231-3233 in #1456 but custom RC accounts still cannot reach it.
- SIE `#SRU`/`#KTYP` are parsed then discarded (`lib/import/account-sync.ts` recomputes), and the account-mapping step offers no VAT controls.

## Proposal (Tier 2 from the analysis)

1. `default_vat_treatment` enum on `chart_of_accounts`, reusing the existing treatment vocabulary: `standard_25 | reduced_12 | reduced_6 | exempt | reverse_charge_domestic | reverse_charge_eu_goods | reverse_charge_eu_services | export_goods | export_services | vmb | rental_voluntary`, each mapping to one ruta (05/42/41/35/39/36/40/07/08 + purchase-side 20-24/50 for class 4-6 accounts).
[...]

PR #1593, opened 2026-08-13
fix(vat): complete account treatment enforcement

mattssonn:
## Summary

- make per-account VAT treatment authoritative across web, import, reporting, and MCP flows
- preserve booking defaults while deriving rates only when unset, including commit-boundary validation
- add safe class-aware migration normalization and explicit SIE/Arcim review
- include inactive accounts in historical VAT filing and keep reverse-charge basis checks rate-specific

## Verification

- focused Vitest regression suite: 570 passed
- npm run lint: 0 errors (271 existing warnings)
- npm run check:lint
- npm run check:guards
- production Next.js build
- staging migration chain and constraint verified; no production database writes

Follow-up to #1588 for #1457.


mattssonn:
Accounting advisory disposition:

- The 25 percent reverse-charge booking rate is an editable default only when no stored rate exists. This follows #1457's explicit requirement to derive a sensible rate when treatment is chosen and the rate is unset; stored 12 or 6 percent defaults are preserved across API, MCP, import, and pending-operation updates.
- Explicit treatment overriding BAS fallback is the requested #1457 behavior. Commit 9e2708e9f appends a superseding DECISIONS.md entry to remove the inherited contradiction from #1588.
- Class 5-6 accounts require review when the label yields a VAT-treatment suggestion. Requiring confirmation for every ordinary class 5-6 expense account would create false mandatory reviews without a treatment to confirm.
- 4425-4427 are the repository's existing BAS domestic-service reverse-charge accounts for ruta 24, mirrored in moms-box-mapping, declaration mapping, filing-gate rate groups, and regression tests.

No production database write was performed.

mattssonn:
Final compliance follow-up:

- Treatment/rate choices are not silently persisted in the import flows: suggested class 3-6 treatments require the row's explicit Confirm VAT review action. Add/Edit and API/MCP callers explicitly select or send the treatment; 25 percent is only derived when no booking rate exists, and remains editable.
- Account metadata changes are audit logged by the existing audit_chart_of_accounts trigger from migration 20240101000018, including UPDATE old/new data. Posted journal entries remain immutable; this feature changes declaration classification metadata and does not edit verifikationer.
- Already stored 12 or 6 percent booking rates are preserved, and the filing gate matches reverse-charge basis by rate.

No unresolved inline review threads remain.

PR #1611, opened 2026-08-13
fix(salary): declare, book and pay AGI in whole kronor (SKV per-sats computation)

mattssonn:
# What and why

A user's first lönekörning showed an AGI payable with both kronor and ören, while Skatteverket deals in whole kronor. Investigation found three connected defects, all fixed here with one shared computation:

- **AGI XML rounded instead of truncating.** Öretal bortfaller (SFF 2011:1261 22 kap. 1 §): every amount field now truncates. FK487 is additionally computed the way Skatteverket computes IK587 (kontroll B_006): per avgiftssats on the whole-krona per-IU underlag sums, with youth/växa monthly-cap splits, in exact integer arithmetic (new `lib/salary/declared-avgifter.ts`). FK497 is the sum of the per-IU truncated taxes, so the HU always agrees with its IUs.
- **The salary booking credited 2731 with exact öre**, leaving a perpetual residual after the whole-krona skattekonto draw. The avgifter voucher now credits 2731 with the declared amount and the remainder to 3740 (Öres- och kronutjämning); the 7510 cost side stays öre-exact. Preview and booking share the identical split.
- **The LB payment file and TaxPaymentPanel paid/showed öre.** `agi_declarations` now stores the declared whole-krona totals (aligning stored = filed = booked = paid, which also lets the skattekonto auto-settlement match SKV's draw to the öre); legacy öre rows keep paying öre-exact so pre-deploy bookings still clear 2731.

Hardening that fell out of three adversarial /skeptic rounds: review overrides route all legs through the same per-category truncation (an operator's adjustment can never book as fake utjämning); basis-only overrides are inert on money totals (they never reach the filed IUs); the v1 book route gains override parity with book-run; F-skatt rows ignore avgifter overrides on every surface; `tax_withheld_override` requires whole kronor; booked runs show their actual posted verifikat (voucher numbers included) instead of a recomputed projection.

Accepted, documented divergences (DECISIONS.md): amount overrides (FoU) cannot reach SKV's beslut until the FoU underlag fields are emitted in the HU (follow-up); regenerating a legacy-booked period's AGI surfaces that booking's öre residual on 2731 instead of hiding it as skattekonto överskott.

Migrations: none.

## Checklist

- [x] `npm run lint`, `npm test` pass locally (known Windows-local SEK-label suite failure excepted; green on CI)
- [x] New or changed logic in `lib/` or `app/api/` has tests (declared-avgifter suite, booking split cases incl. the SKV counterexamples, preview route both branches, payment file both eras, AGI generation öre/override cases)
- [x] New UI strings in both `messages/sv.json` and `messages/en.json`
- [x] No migrations
- [x] No imports from `@/extensions/` in core code

## Summary by CodeRabbit

* **New Features**
  * Payroll calculations now support declared employer contributions by category, including applicable caps and exemptions.
  * Booked payroll runs display their actual posted vouchers, while unbooked runs show journal previews.
  * Journal entries separately show employer-contribution rounding differences.

* **Improvements**
  * Tax and employer-contribution amounts are handled consistently in whole kronor, while legacy fractional amounts remain supported.
  * Tax-payment totals use stored AGI declaration values when available.
  * Tax overrides require whole-krona values, and employer-fee overrides are ignored for F-tax employees.

* **Documentation**
  * Updated payroll booking guidance for contribution and rounding entries.


mattssonn:
@coderabbitai review

PR #1613, opened 2026-08-14
fix: false popup-blocked toast on Visa dokument + scope SKV reconnect line to skattekonto source

jakobwennberg-oss:
Two small UX fixes from feedback 2026-08-14 (screenshot: Granskning slide-over).

## 1. "Kunde inte öppna dokumentet" fired even though the tab opened

`DocumentViewButton` called `window.open(url, '_blank', 'noopener,noreferrer')` and treated a null return as popup-blocked. But `window.open()` returns null **by spec** whenever `noopener` is in the features string — success included — so the destructive toast fired on every successful open. Now opens without the features string and severs the reverse channel manually (`tab.opener = null`), the same pattern `lib/browser/deferred-tab.ts` uses. The toast now appears only on a genuine block.

## 2. SKV reconnect line showed on /transactions permanently

The "Skattekontots transaktioner hämtas inte förrän du anslutit igen…" attn line rendered whenever the connection needed renewal, regardless of source filter — permanent noise. It now shows only when the source picker is on **Skatteverket** (the rows it actually explains). The skattekonto page keeps its own reconnect line, so the state is still discoverable there.

## Verification

- `eslint` on both files: 0 errors (3 pre-existing warnings untouched)
- `tsc --noEmit`: no errors in changed files
- No i18n changes (reuses existing keys), no migrations

## Summary by CodeRabbit

* **Bug Fixes**
  * The Skatteverket reconnection message now appears only when viewing Skatteverket transactions, including when no transaction data is returned.
  * Improved document preview opening and popup detection, while preserving an error message when the preview cannot be opened.

### DIFF
--- a/app/(dashboard)/transactions/page.tsx
+++ b/app/(dashboard)/transactions/page.tsx
@@ -745,7 +745,11 @@ export default function TransactionsPage() {
   )
 
   const sourceItems = useMemo<ContextPickerItem[]>(() => {
-    const showSkvSource = skvRows.length > 0
+    // Keep the Skatteverket source pickable while reconnect is needed even
+    // though the rows fetch fails then (401 -> skvRows []): the reconnect
+    // attn line only renders under this source, so dropping the option would
+    // hide the reconnect path exactly when it applies.
+    const showSkvSource = skvRows.length > 0 || skvNeedsReconnect
     const items: ContextPickerItem[] = [
       {
         id: 'all',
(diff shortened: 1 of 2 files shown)


jakobwennberg-oss:
Confirmed and fixed in c94524873: the Skatteverket source now stays in the picker whenever `skvNeedsReconnect` is true, so the reconnect line stays reachable while the rows fetch 401s. `skvNeedsReconnect` already requires `connected=true`, so never-connected companies get no phantom source. No regression test added deliberately: this repo's test scope is lib/ + app/api/ (no component tests).

PR #1614, opened 2026-08-14
fix(import): preview and migration theater read all fiscal years, not just the newest

jakobwennberg-oss:
## Root cause

`/sie-data` in `extensions/general/arcim-migration/index.ts` parsed only the newest fiscal year's SIE file (`sieFiles[sieFiles.length - 1]`) for both the import preview and the returned `SIEData.parsed`. Mid-year provider exports have few or zero vouchers in the newest year. Accounts were already merged across all files for the mapping step; vouchers and years were not.

Two founder-visible symptoms on the first real Fortnox migration (One.com, 3 fiscal years, 4153 vouchers):

1. **Dishonest preview**: the connect step said "Hittade 740 konton och 0 verifikationer" while the import then landed 4153 vouchers from the older files.
2. **Empty theater**: `buildTheaterModel(sieData.parsed)` got the near-empty newest year, so the migration theater drew a hub plus two year rings and nothing else; no real animation ever happened.

**Correction to the brief**: the "Hittade X konton och Y verifikationer" string renders from `/preview`'s `sieStats` (which fetched with `latestOnly: true`), not from `/sie-data`'s `generateImportPreview` (that `preview` field is not consumed by the wizard at all). Fixing only `/sie-data` would have left symptom 1 on screen, so `/preview` is fixed too: it now fetches every allowed year and computes counts from the merged parse. Cost: one SIE export per extra year at connect time, the same work `/sie-data` repeats right after; honest numbers won over latency.

## Merged-parse semantics (`lib/import/sie-merge.ts`)

`mergeParsedSIEFiles(files)` is pure and browser-clean (no Node APIs):

- **accounts**: union by number, first occurrence wins the name (matches the previous mapping-step loop); missing `sruCode`/`accountType` filled from later duplicates
- **vouchers**: concatenated in file order (oldest first)
- **fiscal years**: union deduped by start+end, sorted oldest first, re-indexed so the newest year is `yearIndex 0` (stats then report the newest year's bounds, the single-file "current year" convention)
- **company name / orgnr**: most common non-empty value, first seen wins ties
- **balances (IB/UB/RES)**: concatenated as-is; each record keeps the yearIndex relative to its source file, documented as presentation-only (per-year balance math on the merged output is not meaningful)
- **dimensions**: deduped by dim number and (dim number, code); **issues**: concatenated

The actual import is untouched: it still posts each raw file separately.

## Validation choice

`/sie-data` still validates only the newest file's parse, exactly as before. Rationale: (a) no previously accepted dataset is newly rejected; (b) `validateSIEFile` assumes single-file invariants (balance yearIndexes relative to ONE current year, cross-year P&L residual math) that the merged view deliberately does not preserve, so validating the merge would produce spurious findings; (c) older files never had a gate here and `/import-sie` surfaces per-file problems at import time.

## Theater pacing

`ArcimMigrationTheater` used to spawn rings, buckets and all four account waves in the first ~2s. Now the account waves are rationed through one shared gate: timers guarantee the full build by ~10s, and every real step-label change during the SIE phase (progress <= 55) births the next wave early. The gate has a 1.8s floor between births because the wizard emits three step labels within the first second; ungated per-step spawns would have collapsed the build right back into the opening seconds. Honest narration untouched: labels and progress are still the wizard's real values, only canvas pacing changed. `prefers-reduced-motion` behavior unchanged (settled graph, no animation).

## Out of scope, noted

The manual SIE path (`app/(dashboard)/import/page.tsx`) holds a single `file` state and parses one file per import run; multi-file merging does not apply there.

## Gates

- `npx vitest run lib/import extensions/general/arcim-migration`: 48 files, 835 tests, all green (6 new merge tests included)
- `npx tsc --noEmit` stash-baseline diff: identical, zero new errors
- `npx eslint` on all touched files: clean
- `npm run check:guards`: pass; `package-lock.json` untouched

## Verification advice

A founder screenshot of the connect-step counts and the theater on the next real migration is the true end-to-end check; this cannot be exercised against a live provider from CI.


jakobwennberg-oss:
Compliance-finding disposition (GDPR Art.5(1)(c), medium, non-blocking): accepted with documentation. The connect-step preview now fetches all allowed years' SIE to compute honest counts — the founder-reported bug was precisely that latestOnly produced a false "0 verifikationer" for a dataset whose import then landed 4 153. The same full multi-year fetch happens unconditionally at the very next wizard step (/sie-data), so this changes WHEN the data transits memory, not WHAT is fetched or stored; nothing is persisted at preview time. The provider (Fortnox et al.) exposes no lighter per-year voucher-count endpoint — SIE export IS the stats source. Rationale recorded in DECISIONS.md in this PR.

PR #1625, opened 2026-08-15
fix(auth): move the BankID flow into a signed, user-gated, single-use cookie

mattssonn:
> Reworked through adversarial review. Earlier localStorage and sessionStorage ownership designs were discarded because they could not safely bind a resumed browser document to the person who started the flow.

## The report

A user tried to create an account with BankID on an iPhone four times. Each BankID return opened another Accounted tab. One tab reached "Verifierad med BankID / E-postadress / Skapa konto", while the tab the user was looking at showed the initial button. Production had no `auth.users` or `bankid_identities` row: identification succeeded, but signup was stranded.

## Cause

On iOS outside plain Safari, the BankID return URL can be handed to the OS and opened in a new browser tab. The in-flight session lived in `sessionStorage`, which is scoped per tab. Login largely hid this because it finishes without extra input and establishes a shared session cookie. Signup waits for an email address in the original tab, so it never landed.

## Change

The BankID session ID no longer reaches browser JavaScript. It lives in a signed `__Host-`, Secure, HttpOnly, SameSite=Lax cookie set by `/start`; `/poll`, `/complete`, `/link`, and `/cancel` read it server-side. Every tab on the origin can therefore find the same flow without exposing its bearer credential.

The security and race invariants are:

- **Explicit resume:** every flow found after a page load shows a "continue only if this was you" card. Nothing resumed is automatically consumed. Only the live component instance that called `/start` polls automatically.
- **Exact flow binding:** `/start` creates a random, non-secret `flowId`, signs it into the cookie, and returns it without returning the BankID session ID. A starting or explicitly confirmed tab sends `x-bankid-flow-id` on active calls. `/poll`, `/complete`, `/link`, and `/cancel` require an exact match. An older same-mode tab therefore cannot silently follow, complete, or cancel a newer flow that replaced the shared cookie. A stale cancel neither clears nor cancels the newer flow.
- **Mode and owner binding:** the request mode must match the signed cookie. A `link` flow also requires authentication at `/start`, pins the user ID, and rejects a different user at `/link`.
- **Single use before minting:** `bankid_consumed_sessions` has a database unique constraint claimed before `generateLink`. Concurrent completion requests cannot mint competing magic links. Non-`23505` claim errors fail closed.
- **One canonical cookie:** `__Host-` plus `Path=/` prevents subdomain or longer-path cookie shadowing. Duplicate cookies fail closed, malformed values return no flow, and all signed state has a fixed total lifetime.
- **No identity leak from probing:** a mount probe returns only flow metadata needed for confirmation, including `flowId`, never a personnummer or name. Names are returned only to an explicitly active signup flow. Logs contain only an eight-character session prefix.
- **Bounded lifetime:** the order window is 300 seconds. Completion can extend the email-entry window, but the signed `startedAt` imposes a 1,200-second total cap.
- **Safe settling:** dead-session polling does not issue an untargeted cookie deletion that could erase a newer flow. Terminal completion and matched cancellation settle the flow. The UI also has dedicated cancelling and completing states, so the user cannot relaunch during those races.

iOS keeps its return URL. Android keeps `redirect=null` as before.

## Tests and verification

- Cookie signing, tamper rejection, duplicate rejection, attributes, lifetime, and `flowId` validation
- Start, probe, active poll, completion, link, cancel, mode mismatch, owner mismatch, stale-tab replacement, and concurrent single-use cases
- Real-Postgres unique claim and deny-all RLS coverage for `bankid_consumed_sessions`
- Swedish and English message-key parity
- Local focused run: 11 files, 186 tests passed
- Local `npm run lint`, `npm run check:guards`, `git diff --check`, and production `npm run build` passed

The full local test command reaches 14,735 passing tests, then fails in an unchanged `format-currency-sek-label.test.ts` parser error that reproduces on `main`. The pull request's clean hosted unit shards are the authoritative full-suite gate.

## Required before merge

- Complete one real signup on an iPhone in Chrome, including the new-tab return and explicit continue card.
- Deploy migration first. Deploying the code before `20260815120000_bankid_consumed_sessions.sql` makes BankID completion fail closed with `session_invalid`.

No new dependency or environment variable is introduced. The cookie signing key is derived from the existing BankID/session server secrets.

## Residual risk

If one person abandons the exact still-live tab and another person takes over that same document without any navigation, no browser signal can distinguish the people. This is the ordinary unattended-live-tab risk. Any reload, restored tab, duplicated tab, or newly opened return tab goes through explicit confirmation.

## Summary by CodeRabbit

* **New Features**
  * Improved BankID login, registration, and linking flows with safer session handling.
  * Added options to continue, restart, or cancel interrupted flows.
  * Mobile BankID flows can resume with clearer controls for reopening the app.
* **Bug Fixes**
  * Improved handling of expired, cancelled, failed, or replaced flows.
  * Prevented duplicate completions, signups, and stale-tab actions.
* **Localization**
  * Added English and Swedish messages for BankID status, recovery, cancellation, and continuation actions.


mattssonn:
## Blocked by adversarial review — do not merge

Converted to draft. Two independent skeptic passes refuted this; findings verified against the code.

**1. The PR's central safety claim is false.** The body says only the *pending* handle is stored, so a record is "worth nothing until the legitimate holder authenticates." It is not. `sessionId` is an unauthenticated bearer credential: `POST /bankid/poll` (`skipAuth`) returns `user.personalNumber` verbatim, and `POST /bankid/complete {mode:'login'}` returns a `tokenHash` that `verifyOtp` turns into a full session, with MFA skipped for `bankid_linked` accounts.

Moving that handle from `sessionStorage` to `localStorage` widens the reach from "XSS in the /login or /register tab at that moment" to "any same-origin XSS, anywhere, decoupled in time, surviving browser restart". Because `login-client.tsx:106` defaults the panel to BankID, `BankIdAuth` mounts on every /login load and the new `resumePending` will claim a planted completed session, poll it, and sign the victim into the attacker's tenant with zero clicks. With `mode:'link'` it binds the wrong personnummer to the account — and that one needs no attacker at all, just a shared machine and an abandoned link flow inside the 6 minute TTL.

**2. Two concrete bugs in `bankid-pending.ts`.** The TTL check sits after the mode check, so a stranded `signup` record is never pruned by /login visits; and `now - ts` is negative for a future `ts`, so a planted record with a future timestamp never expires.

**3. Commit 2 regression.** The render ladder falls through to the waiting block for `status === 'complete'`, and the new "Öppna BankID-appen igen" button is guarded only on `session &&`. During the `/complete` → `verifyOtp` window the user sees "Öppnar BankID-appen..." plus a 44px button inviting the exact action they think is still needed. Tapping it navigates the tab to `app.bankid.com` with a consumed token, aborting the in-flight request and burning the single-use magic link, then returns them to a silent `idle` page. Same shape in `mode="link"`, where it can abort the `/link` POST.

**4. Mobile expiry went silent.** `record.ts` is fixed at persist time while `pollStartedAtRef` resets on every resume, so the TTL always beats `POLL_DEADLINE_MS` on mobile. The "BankID-sessionen löpte ut. Försök igen." branch is now unreachable there; the spinner just snaps back to the start button.

**What held up:** the lock design itself. Desktop QR is provably untouched (no record is ever written outside `isMobile()`), the settings-modal mount/unmount cycle is unchanged, mode-scoping prevents cross-panel hijack, and the two-point ownership check does close the double-`/complete` hazard.

**Direction.** The right fix is the one this PR's `DECISIONS.md` entry wrongly dismissed: put the sessionId in an `HttpOnly, SameSite=Lax, Secure, __Host-` cookie set at `/start`, have `/poll` and `/complete` read it from there, and pin `mode` server-side. Cookies are origin-wide by construction, so that solves the original stranded-tab bug outright — no client storage, no lock, no heartbeat, most of this PR deletes itself — while removing the bearer property instead of spreading it. Awaiting a call on scope before rebuilding.

mattssonn:
## Second adversarial round: BLOCK again

Both skeptics refuted the cookie rebuild. The cookie mechanics themselves held up (HKDF domain separation, no forgery found, `SameSite=Lax` holds against CSRF, mode pinning works as far as it goes, the `/poll` whitelist genuinely removes the personnummer, Path matching verified against every call site). What failed is the resume logic built on top, plus two claims in the commit message that are not true.

**Critical, introduced here.** `/link` pins the mode but never the user. `/start` is `skipAuth` and will open a `link` flow for anyone; `/link` binds to `requireAuth()` — whoever is logged in now. Shared browser: A starts "Koppla BankID", authenticates, and on iOS the return lands where `BankIdSettings` never remounts the component (`isLinking` resets on load), leaving a completed unbound link flow in the cookie jar. B logs in within 6 minutes, clicks "Koppla BankID", and the mount effect silently resumes A's flow — writing **A's personnummer against B's user_id**. A can then sign in as B, MFA skipped. Not possible before: the session id lived in per-tab React state.

**The signature does not bind the cookie to a browser.** It proves server-minted, not browser-bound, so an attacker can mint one with `curl`, authenticate it with their own BankID, and replay. And HttpOnly protects only the exact `(name, domain, path)` triple: JS can plant the same name at a **longer path**, RFC 6265 serializes longer paths first, and `readBankIdFlow` returns the first match. `clearBankIdFlowCookies` writes to the shorter path, so that shadow survives every `settle()` permanently.

**"Single-use server-side" is false.** `settle()` only sets `Max-Age=0` on the response; two concurrent requests carrying the cookie both reach `generateLink`, and the second invalidates the first `tokenHash`. This is a regression — on `main` two tabs held two different sessions, so it was structurally impossible.

**Client-side defects, all introduced here.** Tillbaka on signup fires `/cancel` fire-and-forget and remounts synchronously, so the mount effect resumes the still-live flow and bounces the user back onto the email form, which then dies with `session_invalid`. A stale hint cookie resumes a dead order into a 502 path that never clears the cookie, making BankID unusable for the full 6 minutes. A resumed desktop tab builds `autoStartToken: ''` and renders a dead "BankID på den här enheten" button. Any 404 is now terminal, including the dispatcher's own "Route not found" and Vercel's during a rolling deploy. The signup e-mail window shrank from TIC's ~30 minutes to 360s counted from `/start`, and the `account_exists` retry affordance is worth only whatever is left of it. `POST /bankid/cancel` has no tests.

Root cause of the client cluster: auto-resuming on mount from a client-readable hint fires without a gesture, races cancellation, resumes dead orders, and mounts in contexts where it should not. The fix direction is to drop the hint cookie and probe once via `/poll` (already returns a distinguishable `no_session`), and to never auto-resume a `link` flow. Awaiting a scope decision before a third attempt.

mattssonn:
## Review status: four adversarial rounds, final pass clean

This branch was rebuilt from `main` three times as `/skeptic` refuted earlier designs (localStorage+lock → login-fixation + deadlock; first cookie → wrong-user link binding + path shadowing; user-pinned cookie → session-restore auto-consume). The current design's last review round found three issues, all now fixed:

- **Auto-consume via session restore (security).** `sessionStorage` survives Ctrl+Shift+T / session restore / tab duplication, so the per-tab ownership marker could be inherited by a different person. Closed by gating auto-resume on the flow being **still pending** on first observation — an already-complete flow always routes to a confirm card, marker or not.
- **Avbryt/restart cookie race (correctness).** A `'cancelling'` state now blocks a new `/start` until `/cancel` resolves, so its untargeted clear can't delete a fresh flow.
- **Name disclosure to the pre-gesture probe.** `/poll` withholds `givenName/surname` from a probe; only the post-confirm active poll receives them.

A final verification pass could not refute any of the three and confirmed no test was weakened.

### One residual, flagged for your call (not fixed)

If a flow is **still pending** when a tab is reopened via Ctrl+Shift+T and the marker is inherited, the pending-gate will auto-resume it. To cause harm this needs: someone starts BankID, closes the tab *before authenticating*, another person reopens it within the 300s window, *and* the first person then authenticates that order on their phone. Much narrower than the complete-flow case the gate closes.

Closing it fully means requiring the "Fortsätt" confirm tap on **every** mobile return (including the legitimate same-tab reload), not just cross-tab ones. That's a product tradeoff — one extra tap on every mobile BankID sign-in vs. this narrow residual — so I left it as your decision rather than bake it in.

### Not verified
No iPhone here — the real-device signup on iOS Chrome is unverified. Deploy is **migration-first** (single-use consume fails closed) or BankID auth returns `session_invalid` until the migration lands. `pg-real` is green.

Left as a draft for your review; not merged.

PR #1629, opened 2026-08-16
fix(ux): actions update lists in place - no more takeovers, jumps and dead air

jakobwennberg-oss:
## Why

Founder: "it is feeling a bit glitchy when clicking around and especially when doing actions like deleting a row or booking something. Optimize for the user to feel a smooth experience."

A 4-agent loading-states sweep (dev_docs/loading_states_analysis.md, 37 findings) traced the glitch to one repo-wide anti-pattern: **single-row actions trigger whole-list skeleton/spinner takeovers** (hundreds of px collapse to a spinner box, scroll position lost, stagger-enter entrance replays over every row), deletes give zero feedback and then hard-jump, and the /transactions "exit animation" was filter-only: rows popped out the same frame, so the page's single most-used action always ended in a layout jolt.

This PR fixes the highest-leverage subset with one unified vocabulary: skeletons are for first paint only, refreshes reconcile behind the rendered list, and rows that leave visibly animate out.

## Per surface

**Never take over a rendered list for a background refresh** (skeleton/spinner reserved for an empty or foreign list; quiet inline cue near the list header where a reconcile is worth signalling):

- `/transactions` fetchTransactions: period changes and realtime echoes refetch behind the rendered rows (client-side bounds filters already show the new scope in the same frame); small Loader2 next to the FyPicker chip during scope reconciles
- `/pending` (Granskning) fetchOperations: approving/rejecting one row used to play list -> spinner -> list -> spinner -> list (direct refetch + realtime echo, each swapping the list for DataListLoading and replaying the stagger). Both listed "/pending" and "Att göra approvals" findings point at this one file; refetches are now silent reconciles with a quiet toolbar cue, tab switches still take over (different dataset)
- Kundfakturor list: bulk Bokför no longer collapses hundreds of rows to 3 skeleton stubs with a full stagger replay
- Leverantörsfakturor list: register/betalfil/approve-fallback refetches reconcile in place; also wrapped the fetch in try/catch/finally with an error toast, since a network failure used to stick the skeleton forever or silently render "no supplier invoices"
- Verifikat list: `JournalEntryList` gains a `refreshToken` prop and refetches in place (its own hasLoaded + dim-at-opacity-60 path); `/bookkeeping` no longer `key=`-remounts it, so creating a verifikat keeps expansion, selection, pagination and scroll

**/transactions row exit + delete feedback** (doc findings 1-2):

- Exiting rows (booked / ignored / deleted, bank and skattekonto) stay rendered through the existing 350ms window with a real exit transition: fast fade + the space closes by transitioning cell paddings/line metrics and a numeric max-height on the fixed-height cell spans (`.row-exit` / `.row-collapsible` in globals.css), pointer-events off; batch-select excludes exiting rows. `prefers-reduced-motion`: instant removal
- Delete routes through `processingId` (row spinner) + `exitingIds`, and decrements `totalUncategorizedCount` when the deleted row was pending, so the inbox badge no longer goes stale (realtime echo is not guaranteed for DELETE on a filtered subscription). History-view rows get the same exit fade

**FyPicker double-fetch** (doc finding 3): the initial fetch now waits for FyPicker's existing `onReady` (fires after its persisted-scope restore), so mount does ONE correctly scoped fetch instead of an unscoped fetch racing the restore's second fetch with a second skeleton (list -> skeleton -> list on every visit). Deviation from the doc's literal fix ("read the persisted period synchronously in initial state"): localStorage holds only the period id, not the period bounds, and a pre-set value suppresses FyPicker's restore onChange, which would leave the fetch permanently unscoped while the chip claimed a year. Same outcome, no new API (logged in DECISIONS.md)

**Pagination preserved on realtime echo**: background refreshes fetch `range(0, pagedCountRef.current - 1)` instead of resetting to the first 200 rows, so content loaded via "Visa fler" no longer vanishes (with the scroll jump) after any action

## Deliberately NOT in this PR (follow-ups, all in dev_docs/loading_states_analysis.md)

- Invoice/supplier-invoice/customer/article DETAIL page takeovers (fetchInvoice/fetchCustomer/fetchArticle isLoading swaps, useDestructiveConfirm's defeated pending state, per-row Attestera spinner)
- Batch-action parallelization (sequential per-row round trips; skv bulk path is the model)
- Toast-spam reduction on batch booking (per-row + aggregate toasts)
- Underlag inbox row-click flash (content -> empty state -> spinner -> document) and its delete/realtime churn
- Route-level loading.tsx silhouette mismatches (suppliers/articles/detail segments)
- Bank-sync router.refresh no-op and the misleading balances comment
- Per-row committingId on /pending approve pills; /pending optimistic row removal

## Gates

- Full vitest suite green: 14764 passed (incl. updated booking-feedback-parity guard: delete is now a counted decrement path, 6 -> 7)
- `tsc --noEmit` byte-identical to the origin/main stash baseline (zero new)
- eslint: 0 errors on touched files, warnings at baseline parity
- `check:guards` green, package-lock untouched, sv+en message keys added in both files

## Visual sign-off

NOT VISUALLY VERIFIED: this is motion/loading behavior; per house rule, no UI PR merges without visual sign-off. Suggested check on the preview: book + delete a row on /transactions (rows should fade-collapse, never pop), approve one row on /pending (list must not blink to a spinner), bulk Bokför on /invoices, create a verifikat on /bookkeeping (journal stays put), and reload /transactions with a persisted räkenskapsår (one skeleton, no list flash). Also verify once with "reduce motion" enabled (rows must disappear instantly).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed duplicate loading when changing the fiscal year in Transactions.
  * Improved refresh, stale-response, and error handling across invoice and pending-operation lists.
  * Preserved existing table contents during background updates.

* **User Experience**
  * Added clearer loading and refresh indicators across bookkeeping sections.
  * Added smooth transaction-removal animations with reduced-motion support.
  * Deleting transactions now shows progress and updates pending counts immediately.

* **Localization**
  * Added English and Swedish translations for loading and refresh states.


jakobwennberg-oss:
Triage round 2: comments 1-6 are the re-posted already-fixed findings (addressed in 2e55b113d). New findings 7 (foreign-tab failure can briefly render the other tab's rows) and 8 (company-switch data boundary + loadSkvRows response guard) are real but display-only rare-edge hardening with no data risk — queued as follow-ups in dev_docs/loading_states_analysis.md alongside the other deferred items, to ship as one batch rather than respinning this hot branch per finding.

PR #1633, opened 2026-08-17
fix(ux): smoothness follow-ups - detail pages, batches, toasts, and the last edges

jakobwennberg-oss:
Follow-up batch to #1629, shipping the documented deferred items from `dev_docs/loading_states_analysis.md` as one PR. Same vocabulary as #1629: first-load-only takeovers, background reconciles behind mounted content, row/button-level pending, sequence guards.

## Scope

**1. Invoice detail pages stop blanking on one-field changes**
- `invoices/[id]`: `fetchInvoice()` now shows the blocking spinner only before the first paint (or when the pager steps to a different invoice). Bokför / status change / finalize / payment / send refetch behind the mounted page; the acting buttons (Bokför, mark-sent for följesedlar, makulera proforma) show a spinner-in-button and the mutation handlers await the refetch so pending covers until the page reflects the new state.
- `supplier-invoices/[id]`: same first-load-only guard. The single `isProcessing` boolean became `processingAction` so the spinner lands on the clicked button only (Attestera, Bokför, Kreditera, Återupprätta, Ta bort); all other buttons still just disable. Delete previously had no pending state at all and permitted duplicate DELETEs; it now runs inside the confirm dialog (below) with the icon button spinning.
- Leverantörsfakturor **list** try/catch/res.ok: already fixed by #1629 (`fetchInvoices` has try/catch/finally + res.ok there); nothing to do.

**2. `useDestructiveConfirm` async onConfirm**
- `confirm(opts, action?)` can now carry the destructive operation: the dialog stays open with its existing (previously unreachable) `isLoading` spinner until the action settles, blocks dismissal meanwhile, and resolves `false` if the action throws. Backwards compatible: all existing `confirm(opts)` call sites are unchanged.
- Adopted at the two sites in scope: /transactions row delete and `supplier-invoices/[id]` delete. Batch confirms deliberately keep the close-then-run shape: their live feedback is the bulkbar counter, which a held-open modal would cover.

**3. Batch parallelization on /transactions**
- New `lib/concurrency.ts` `mapWithConcurrency` (worker pool, order-preserving, with tests). Batch categorize / ignore / delete run their per-row requests 5 at a time instead of strictly sequentially; the bulkbar counter ticks per completed row. Never an unbounded fan-out.

**4. Toast-spam reduction**
- Batch categorize rows run `silent`: no per-row "Bokförd" toast, no per-row generic failure toast; the exit animation, count decrement and state patch stay. ONE aggregate toast: "N bokförda" or "N bokförda, M misslyckades".
- **Undo:** per-row undo IS feasible today (each booked row storno-reverses via the same `/uncategorize` endpoint as the per-row Ångra), so the aggregate toast carries a single **Ångra alla** that pools the same endpoint over every booked row and reports once ("N ångrade" / partial). The endpoint call stays in one place (`undoCategorize`), pinned by `booking-feedback-parity.test.ts`.
- Deliberately NOT silenced: the interactive escalations (SI/CI match suggestion dialogs, duplicate warning, activate-account toast): they are the only way forward for those rows. Batch delete/ignore already emitted one aggregate toast each; unchanged apart from the pool.

**5. Underlag row-click flash (InvoiceInboxWorkspace)**
- `handleSelect` seeds the detail pane synchronously from the clicked list row (fetchItems returns full rows) and starts the document load in parallel with the detail GET, which hydrates `selected` on arrival. No more content → onboarding/empty flash → spinner → document; two serialized round trips became parallel. A stale-response guard (`docRequestRef`) keeps a slow item-A fetch from overwriting item B's pane.

**6. #1629 round-2 edge hardenings**
- `/pending`: a FAILED fetch for a tab whose rows aren't on screen now HOLDS the loading state (plus the error toast) instead of rendering the previous tab's rows under the new tab's header. Held rather than cleared-to-empty: "Inget att granska" after a failed load would be indistinguishable from a genuinely empty list.
- `/transactions`: company switch now clears `transactions`/`skvRows` (+ count/paging state) in the existing render-time reset block and bumps both fetch sequences, so old-company rows neither linger under the new company's header nor land later from an in-flight response. `loadSkvRows` got the same sequence-guard pattern as `fetchTransactions`.

## Not done (from the doc), and why
- Everything outside the six scoped items (FyPicker/realtime findings, list pages, register loading.tsx files, deadlines/recurring/mileage, kund/leverantör/artikel detail saves, underlag delete confirm() swap, bank-sync router.refresh): out of scope for this batch; several were already fixed in #1629.
- Batch confirm dialogs not switched to the async-confirm shape (see item 2 rationale).
- No new animation was added, so no new `prefers-reduced-motion` handling was needed; the row-exit path reused from #1629 already handles it (instant removal).

## Gates
- Full `npx vitest run`: green (1151 files passed, 14772 tests passed, 0 failed)
- `npx tsc --noEmit`: byte-identical to origin/main baseline (stash-diffed)
- eslint on touched files: 0 errors, warnings byte-identical to baseline
- `npm run check:guards`: green
- `package-lock.json`: untouched
- New lib helper has unit tests (`lib/__tests__/concurrency.test.ts`)

## VISUAL sign-off needed (preview checklist)
Per the standing rule, no UI PR merges without visual sign-off. On the preview:
1. Kundfaktura detail: Bokför a sent invoice (deferred booking) → page must NOT blank; button spins; verifikat row appears in place. Markera betald + send flows likewise.
2. Leverantörsfaktura detail: Attestera → only that button spins, page stays mounted. Ta bort → dialog shows spinner until it closes, no double-click double-DELETE.
3. /transactions: delete a row → confirm dialog spinner engages, then the row animates out.
4. /transactions batch: select ~10 rows → Bokför valda → counter ticks fast (pooled), rows animate out, ONE toast with "Ångra alla"; click Ångra alla → rows return, one result toast.
5. Underlag (/e/... invoice-inbox): click between rows → fields rail and preview must never flash the onboarding card or "Välj en post"; PDF starts loading immediately.
6. /pending: switch tab with network offline (devtools) → spinner + error toast, never the other tab's rows.
7. /transactions: switch company → list clears to skeleton, never shows the previous company's rows.


jakobwennberg-oss:
All 7 CodeRabbit findings verified against current code and applied in 2753ce0ec, with two deliberate deviations from the proposed fixes:

1. **Aggregate count (transactions batch)**: runCategorize now resolves `{ ok, journalEntryId }` and the toast aggregates on `ok`, as proposed. But "Ångra alla" targets only rows with an actual `journalEntryId`, not all `ok` rows: `POST /uncategorize` returns 400 for a transaction without a journal entry, so including successful flag-flip rows would produce guaranteed "kunde inte ångras" failures for rows that have no verifikat to reverse.

2. **Monotonic token (InvoiceInboxWorkspace)**: implemented for both the detail and the document reads. The suggested regression test (request two resolves before request one) is not added: this repo runs Vitest in the node environment with no component rendering (test scope is lib/ + app/api/), so an out-of-order-resolution test for a React component has no home here. The token pattern itself is the same one already used by the transactions page's skvFetchSeqRef.

Also updated `booking-feedback-parity.test.ts`, which pins the undo-guard shape, to assert the new shared `undoneIdsRef` guard.

PR #1634, opened 2026-08-17
fix(invoices): article pre-fills ROT/RUT and kundkort personnummer covers the claim

jakobwennberg-oss:
## User report (kalletoxic, 2026-08-17)

> Verkar inte som att RUT/ROT är fördefinierat även om jag väljer en artikel som har det satt på sig. Måste manuellt välja på varje rad. Dessutom verkar inte heller personnummer populeras upp på fakturan från kundkortet.

## Bug 1: article's ROT/RUT never reached the invoice line

`articles.housework_type` was documented as "pre-fills the invoice line" but the invoice editor never selected the column, and `applyArticle` never touched the line's deduction fields.

- `ArticleOption` + the picker query now carry `housework_type`.
- New `deductionTypeForWorkType()` in `rot-rut-rules.ts` derives ROT vs RUT from which Skatteverket work-type list the code belongs to (the lists are disjoint).
- `applyArticle` sets `deduction_type` + `work_type` with the same overwrite semantics the function already uses for description/price/unit: an article **without** a code clears the deduction, so switching a RUT-flagged row to a material article cannot keep claiming a deduction on material (HUSFL labor-only rule). Accrual is cleared when a deduction lands, mirroring the manual ⋮ menu rule.
- "Spara som artikel" round-trips the line's work type onto the created article, so the loop closes.

## Bug 2: kundkort personnummer never used for the claim

`customers.personal_number` is encrypted at rest and only ever reaches the browser as ciphertext or a mask, so a client-side prefill is impossible by design. The fix is a server-side fallback in `buildInvoiceWriteData` (shared by web, v1 API and MCP): **typed > stored draft > kundkort**.

- The kundkort value is decrypted (`revealStoredCustomerPersonalNumber`), expanded to 12 digits (new `expandPersonnummerTo12`: standard century inference, `+` over-100 marker, samordningsnummer day offset), Luhn-validated, then encrypted into the invoice columns exactly like a typed value.
- Invalid/unreadable kundkort values are treated as absent and fall through to the existing "Personnummer krävs" error rather than surfacing a cryptic invalid-personnummer error for a value the user never typed.
- The editor drops the required-mark and shows "Hämtas från kundkortet om fältet lämnas tomt" when the selected customer has a personnummer.

## Tests

- `expandPersonnummerTo12`: 12/10-digit forms, century boundary on today's date, `+` marker, samordningsnummer offset, mask rejection.
- `deductionTypeForWorkType`: ROT/RUT/unknown codes (incl. the IT-tjänster ROT→RUT move).
- `buildInvoiceWriteData`: fallback from encrypted + legacy 10-digit plaintext kundkort values, typed-wins, stored-draft-wins, invalid-kundkort-falls-through.

`npx vitest run lib/invoices lib/salary/__tests__/personnummer.test.ts`: 775 passed. ESLint: 0 errors on touched files. Both message files updated.


jakobwennberg-oss:
Triage of the compliance review findings:

- **Article round-trip tagging a material article** — declined. "Spara som artikel" always POSTs a **new** article; it never mutates an existing/shared one, so the described contamination path does not exist. The remaining case (a manually ROT-flagged line saved as a new article carries its work-type code) preserves an explicit per-line user choice, and the labor-only warning strip is visible on any flagged line. Dropping the round-trip would recreate bug 1 for user-created articles.
- **Century inference silently wrong** — declined. The 10→12 expansion is the standard deterministic semantics of the 10-digit form (no '+' = under 100, '+' = 100+), not a heuristic; any Swedish system expands it identically. The last four digits remain visible on the invoice and in the editor hint, and Husavdragstjänsten rejects a non-matching personnummer rather than silently paying out. Requiring a typed value for 10-digit kundkort rows would gut the fix for the most common stored form.
- **IT-tjänster 2026-07 note** — the test comment mirrors the pre-existing source comment in `rot-rut-rules.ts`, which mirrors the Begaran.xsd V6 element lists; no new claim introduced.
- **Individual-only gate** — accepted and fixed in 91933fec7: the kundkort fallback now requires `customer_type === 'individual'` (defense in depth; the Zod schemas already forbid `personal_number` on non-individuals but the DB does not). Typed personnummer unaffected.

PR #1655, opened 2026-08-17
fix(invoices): remaining_amount can no longer be inserted as 0 on an unpaid invoice

jakobwennberg-oss:
## Summary
Root-cause fix behind the `remaining_amount` backfill run on prod today (337 open invoices sat at `remaining_amount = 0` while unpaid; snapshot in `public._backfill_remaining_20260817`).

`remaining_amount` is `NOT NULL DEFAULT 0` and every payment surface (payment dialog, bank match, Stripe sync, agent mark-paid) reads it as the customer's open balance. Four writers omitted it, so their invoices looked settled: the dialog rejected every payment as an overpayment and the bank match saw nothing to clear.

- **Migration `20260817191708`**: `BEFORE INSERT` trigger `invoices_derive_remaining_amount`. When `remaining_amount` is NULL/0 on a real invoice (`document_type = invoice`, not a credit note) with `total > 0` and a status that still owes money, it becomes `total − paid_amount − deduction_total` (≥ 0). The ROT/RUT share is a 1513 receivable on Skatteverket, never the customer's, exactly as `buildInvoiceWriteData` computes it. INSERT only: settlement code owns updates and legitimately writes 0 on full payment.
- **pg-real test** `tests/pg/invoices-remaining-amount-guard.pg.test.ts`: derivation, explicit value respected, paid/prior/deduction arithmetic, drafts + overdue, paid/cancelled keep 0, credit notes and proformas untouched, never negative. Run green locally against a fresh `supabase/postgres:15.8.1.060` with all 643 migrations replayed.
- **Writers fixed too** (defense in depth): proforma → invoice conversion (dashboard route and MCP `commitConvertInvoice`), MCP `commitCreateInvoice` (every agent-created invoice was affected), sandbox seed (the recurring F-2026002/003 rows).

## Verification
- `npx vitest run app/api/invoices app/api/sandbox lib/pending-operations` → 58 files / 670 green; tsc clean; eslint 0 errors; `check:guards` green; `check-pg-test-coverage` recognises the test.
- Migration auto-applies to prod on merge (existing setup). No data change in the migration itself.

## Summary by CodeRabbit

- **Bug Fixes**
  - Invoice balances are now initialized consistently when invoices are created or converted.
  - Missing remaining amounts are calculated from totals, payments, and deductions.
  - Explicit zero balances for settled invoices remain unchanged.
  - Negative remaining balances are prevented, with correct zero balances for paid, cancelled, credited, and non-invoice records.

- **Tests**
  - Added coverage for unpaid, draft, overdue, paid, cancelled, credited, and deduction scenarios.


jakobwennberg-oss:
Review triage:
- **Applied**: draft seed row now carries `remaining_amount` + `paid_amount` like the rest of the batch (PostgREST union-of-keys normalisation, good catch; the new trigger would have rescued `remaining_amount` but the writer contract should not rely on it).
- **Declined**: `beforeEach` mock/event-bus reset in the pg-real test. That guideline is for the mocked unit suite; none of the 106 existing `tests/pg/*.pg.test.ts` files do it (they run against a real database with no mocks and no event bus), and this file follows their harness.

PR #1657, opened 2026-08-18
feat(mileage): prefill distance and purpose from earlier trips on the same route

mattssonn:
## What

Route memory in the korjournal trip dialog. Fran/Till autocomplete from the company's earlier trips (datalist), and when the entered from/to pair matches a previous trip, the one-way distance and purpose prefill from the most recent match. Only empty fields are filled, so nothing the user typed is overwritten; edit mode is untouched. A muted hint under the km field shows when the value came from route memory, and it clears as soon as the user edits the distance.

## Why

Beta feedback from a korjournal user: most real-world mileage logging is the same few routes repeated, and retyping the km every time is the main friction. This removes it for the recurring case without any external maps/routing API, works self-hosted, and is fully deterministic.

## Migrations

No.

## Test coverage

- New `lib/mileage/__tests__/route-memory.test.ts`: normalization (case/whitespace/aao), direction sensitivity, round-trip halving back to one-way with 1-decimal rounding, most-recent-wins ordering (trip_date then created_at), blank/no-match returns null, input order not mutated, distinct suggestion ordering.
- `npx vitest run lib/mileage` green (37 tests), `check:lint` and `check:guards` green.
- New UI string `mileage.route_prefill_hint` exists in both `messages/sv.json` and `messages/en.json`.

## Summary by CodeRabbit

* **New Features**
  * Added route memory to mileage trip creation.
  * Existing locations are suggested as you enter a route.
  * Matching previous routes can prefill distance and trip purpose.
  * Added a hint when details are automatically filled.
  * Added English and Swedish translations for the route-prefill message.

* **Bug Fixes**
  * Prefilled values reset when starting or copying a trip.
  * Manually changed values are preserved and no longer marked as prefilled.
  * Editing existing trips remains unaffected.


mattssonn:
@coderabbitai review

mattssonn:
@coderabbitai review

PR #1683, opened 2026-08-18
fix: unignore transactions during categorization

### LINKED ISSUE #1660: Fix batch-categorize crash on transactions_is_ignored_no_journal_entry constraint
**Priority:** high

`POST /transactions/batch-categorize` with `is_business: false` books the row as private (a real journal entry) and writes `journal_entry_id` without clearing `is_ignored`. For rows the user previously ignored, the update violates the DB check constraint `transactions_is_ignored_no_journal_entry` and the raw Postgres error escapes as a generic `INTERNAL_ERROR`. A customer reports 33 of 40 attempts failing when marking private card transactions that came in via PSD2. The single-transaction categorize routes have the same gap.

## Relevant code
- `app/api/v1/companies/[companyId]/transactions/batch-categorize/route.ts:365-383`: the offending UPDATE; never touches `is_ignored`; raw constraint text surfaces as INTERNAL_ERROR
- `supabase/migrations/20260529190000_transactions_is_ignored.sql:27-29`: constraint requires ignored rows to have no journal entry
- `app/api/transactions/[id]/ignore/route.ts:24-46`: ignore side guards on `journal_entry_id`, but categorize side never reverses `is_ignored`
- `app/api/transactions/[id]/categorize/route.ts:920-925`: same gap in the single-transaction route

## Next steps
- Decide semantics: categorizing an ignored row should either set `is_ignored: false` in the same UPDATE or return a typed error
- Apply the same fix to both single-categorize routes and the batch route
- Map the constraint violation to a typed error code instead of INTERNAL_ERROR as defense in depth
- Add tests: ignored row + batch categorize, ignored row + single categorize

---
**Gmail thread ID:** `1a00f995ff83dd5b`

mattssonn:
## Root cause

Categorization posts a journal entry before linking it to the transaction. The batch route and single-transaction categorization flows wrote `journal_entry_id` without clearing `is_ignored`, so previously ignored PSD2 rows violated `transactions_is_ignored_no_journal_entry`. The raw PostgreSQL constraint failure then degraded to `INTERNAL_ERROR`.

The post-then-link error path could also leave the newly posted journal entry orphaned, and the shared categorization core did not use a company-scoped compare-and-set update.

## Change

- Treat categorization as explicit intent to book: batch, single-transaction, `/book`, and shared-core updates now atomically set `is_ignored: false`.
- Map this exact PostgreSQL constraint to `TX_CATEGORIZE_IGNORED_CONFLICT` with HTTP 409, while preserving existing mappings for unrelated database errors.
- Compensate rejected links and CAS losers with engine-backed storno instead of editing posted journal entries.
- Add a company-scoped CAS to the shared categorization core.
- Add regressions for ignored rows, typed fallback behavior, unrelated error mapping, CAS races, and storno compensation.

## Impact

Users can categorize or mark previously ignored transactions as private without a database constraint crash. Concurrent or rejected post-then-link flows no longer silently strand posted journal entries.

## Validation

- Focused regression suite: 133 tests passed across 9 files.
- `npm run lint`: passed with the existing warning baseline.
- `npm run check:guards`: passed.
- Full four-worker local unit scope: 15,382 tests passed. One pre-existing Windows-only loader failure remains in `scripts/checks/__tests__/format-currency-sek-label.test.ts` with `SyntaxError: Invalid or unexpected token`; it reproduces independently of this diff.

## Residual risk

Storno compensation is best-effort. If reversal fails because of an external bookkeeping condition, the failure is logged and voucher-gap documentation is attempted so an operator can reconcile it manually.

No migrations or database writes are included.

Fixes #1660

## Summary by CodeRabbit

* **Bug Fixes**
  * Categorizing or booking an ignored transaction now restores it consistently across single, batch, and pending-operation workflows.
  * Added clear conflict responses when concurrent changes prevent categorization, with guidance to reload and retry.
  * Prevented orphaned journal entries when transaction updates fail or encounter conflicts.
  * Improved company-scoped transaction updates to avoid cross-company changes.
  * Improved recovery when reversing failed bookkeeping entries, including more accurate voucher-gap reporting.
* **Tests**
  * Expanded coverage for ignored transactions, race conditions, booking conflicts, and journal-entry recovery.


mattssonn:
@coderabbitai review

mattssonn:
@coderabbitai review

mattssonn:
Verified against the existing bookkeeping engine and audit path: reverseEntry allocates the correction voucher through getNextVoucherNumber, creates a source_type=storno entry linked with reverses_id, and atomically links the original through reversed_by_id. The journal_entries audit trigger records the commit and reversal with old/new snapshots. The is_ignored transition is transaction triage state that is cleared atomically with categorization, not a separate bookkeeping posting. Successful categorization emits transaction.categorized; any compensation emits journal_entry.reversed and is durably represented by the linked journal entries and audit rows. No additional code or migration is needed for this note.

mattssonn:
Addressed the latest review in c7e4e0466:\n\n- Batch categorization now uses the registry's Swedish conflict message.\n- The dashboard /book update failure now goes through the canonical localized errorResponse path.\n- The suggested later-stage unused-allocation tagging is intentionally not applied. reverseEntry never deletes the reversal header in those paths; it preserves it as status=cancelled and deletes only its lines. detect_voucher_gaps counts every non-draft header, including cancelled rows, so the allocated voucher number remains occupied and is not a gap. Tagging it as unused would create a false voucher-gap explanation. A regression test and DECISIONS.md entry now pin this invariant.\n\nFor the compliance advisory, an untagged storno failure means no unused reversal number is proven. The original posted voucher remains durable and visible in the journal register, including its unlinked state. The helper records a gap only when the engine proves that a sequence number was allocated before any reversal header existed; otherwise it logs manual reconciliation without fabricating a voucher gap.

mattssonn:
Compliance advisory disposition on the final head:\n\n1. The reversal insert already uses .insert(...).select().single(). Under normal PostgREST semantics, zero returned rows produce a non-null single() error, including RLS insert rejection; that path is tagged with the exact unused allocation. The separate data:null,error:null branch therefore represents a malformed or ambiguous client response, not a confirmed no-op. If the insert committed but the response was lost, marking the number unused would falsely document a live reversal header as a gap, so the conservative untagged result is deliberate.\n\n2. Every non-gap compensation failure leaves a durable ledger state:\n   - failures before allocation leave the original posted, unlinked voucher unchanged and visible for reconciliation;\n   - failures after reversal-header creation preserve that header with 
everses_id, normally as cancelled, while deleting only lines;\n   - a concurrent EntryAlreadyReversedError leaves the winning storno link plus the preserved cancelled competing header;\n   - only failures after durable sequence allocation but before any reversal header are tagged and written to oucher_gap_explanations.\nNo branch deletes a posted voucher or fabricates an unused-number explanation.\n\n3. The classification update in this PR is not a silent ledger correction: it is the same transaction categorization mutation, now atomically clearing is_ignored, and all affected flows emit 	ransaction.categorized. The shared/dashboard paths also retain their existing processing-history append behavior.

PR #1685, opened 2026-08-18
fix(invoices): make the Swish QR encode the amount to pay after ROT/RUT deduction

mattssonn:
# What and why

A user reported that on invoices with RUT work the Swish QR code does not match the amount to pay: it ignores the skattereduktion entirely.

Root cause: the PDF totals block and the invoice email state "Att betala" via `getAmountToPay()` (rounded total minus the ROT/RUT deduction, fakturamodellen), but `buildSwishQrDataUrl` still built the QR from the pre-deduction `getDisplayTotal()`. Since the Swish "Type C" payload locks the amount (editmask 0), a customer scanning a RUT/ROT invoice was asked to pay the full total with no way to correct it in the app: overpaying by the entire deduction (50% of labor incl. VAT for RUT, 30% for ROT).

Fix: swap the QR amount source to `getAmountToPay(invoice, company).toPay` so the QR, the printed "Att betala", and the email always agree by construction. A fully deducted invoice (toPay = 0) now renders no QR via the existing `amount > 0` guard. All seven render surfaces (dashboard send/preview/pdf, v1 send/pdf, MCP pending-operations commit, recurring invoices, issue-and-book) funnel through this one helper, so no call-site changes were needed.

Out of scope, logged in DECISIONS.md: the preview-pdf route builds a `previewInvoice` with no deduction fields at all (internally consistent, but the editor preview ignores ROT/RUT); that is a separate display gap to file as its own issue.

**Migrations:** no.

**Test coverage:** 5 new unit tests in `lib/invoices/__tests__/pdf-render-helpers.test.ts` asserting the exact Swish payload string: deduction applied, öresavrundning-before-deduction order, fully-deducted renders no QR, credit-note parity with `getAmountToPay`, and the no-deduction regression case. `npx vitest run lib/invoices`: 1069 passed; `app/api/invoices` + `lib/pending-operations`: 596 passed; lint 0 errors; `check:guards` green.

## Checklist

- [x] Title follows Conventional Commits (`fix:`)
- [x] `npm run lint` and targeted `npm test` suites pass locally
- [x] No new UI strings (customer-facing PDF only, no chrome text changed)
- [x] No migrations
- [x] Core builds with zero extensions enabled (no `@/extensions/` imports)

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected Swish QR codes to use the accurate amount payable, including deductions and öre rounding.
  - Prevented QR codes from appearing on non-payable or credited invoices.
  - Improved invoice PDF data to include correct totals, delivery dates, and deduction details while excluding sensitive information.

- **Tests**
  - Added regression coverage for Swish QR calculations and invoice PDF projections.

- **Documentation**
  - Documented the QR-code correction and invoice PDF behavior.


mattssonn:
@coderabbitai review

PR #1701, opened 2026-08-19
fix(rattelse): let a bank-side strike land on the linked bank amount

jakobwennberg-oss:
## Summary

Discord report (Sebastian, 2026-08-19): a bank-linked verifikat was booked as **1930 D 10 874,81 / 1930 K 10 874,81** (the credit should have been 2970 Förutbetalda intäkter). Striking the wrong 1930 K line in *Stryk rader i verifikatet* and adding 2970 K was refused:

> Raden mot konto 1930 kan inte ändras: verifikationen är kopplad till en banktransaktion eller betalning. Använd rättelseverifikat (storno).

The guard in `correct_entry_lines_inline` required the per-account net on the bank side to stay *unchanged* whenever the entry is linked to a transaction. That protects the common case (bank line right, contra line wrong) but also blocks the one rättelse that makes a mis-booked bank side match the feed again: here the 1930 net was **0** against a **+10 874,81** deposit, and the fix moves it *to* the bank amount.

## Change

New migration `20260819092408_inline_rattelse_bank_anchor.sql` redefines the RPC with the guard anchored to the external amount instead of the pre-state:

- A non-zero change on a 19xx / cash-ledger account is allowed **iff** the post-state net on that account equals the signed sum of the linked bank transactions resolved to it (once per transaction: split links count `allocated_amount`, otherwise `transactions.amount`; `cash_account_id` NULL resolves to the primary cash account, then 1930).
- No anchor resolves, or the post-state does not match: still refused, and the message now shows both amounts ("…kopplad till en banktransaktion på X kr, och kontots belopp efter rättelsen skulle bli Y kr…").
- Net-preserving strikes (description fixes) stay allowed as before.
- Reskontra sides (15xx / 24xx behind `invoice_payments` / `supplier_invoice_payments`) stay strictly net-preserving.
- Per account, not group net, so a wrong-bank-account booking (1930 vs 1940) remains a storno job (DECISIONS.md line added).

Nothing else in the function changes; the em dashes in the copied messages became colons per CLAUDE.md.

## Tests

`tests/pg/inline-rattelse.pg.test.ts` (+2 tests, 24/24 green against a fresh `supabase/postgres:15.8.1.060` with every migration applied):
- the Discord case: partial move refused with both amounts; full move 1930 K → 2970 K allowed; moving the now-correct bank side again refused
- split links on a 1940 cash account with one transaction carrying both the direct FK and a link row (no double counting): partial refused, full move allowed

`check-pg-test-coverage` passes; strike-lines route unit tests unchanged and green.

## Switch-on

Migration applies to prod on merge via the branching auto-apply. Sebastian can then redo the same strike in the dialog.


jakobwennberg-oss:
Triage of the compliance review notes:

- **NULL `cash_account_id` fallback**: agreed it's the edge to watch. Failure modes split asymmetrically: for a strike on a NON-primary account (say 1940) a NULL-tx resolves to primary/1930, so no anchor resolves on 1940 and the RPC keeps the old strict refusal (safe, pre-existing behavior; storno still available). The wrongly-permit direction requires a legacy NULL-cash-account transaction in a multi-account company AND a post-state that exactly equals the misattributed anchor sum to the öre — and the result is still a balanced, fully logged rättelse. For single-account companies (the overwhelming majority of NULL rows) the fallback is simply correct. Taking the suggested two-cash-account pg test as a follow-up rather than blocking: the current test suite already pins that an unresolvable anchor refuses.
- **`transactions.amount` immutability**: agreed, that's the load-bearing assumption. Nothing in the app edits a bank transaction's amount after import; any future feature that does must be reviewed against this anchor.

PR #1702, opened 2026-08-19
fix(salary): show the sysselsättningsgrad product next to the run salary input

jakobwennberg-oss:
## Summary

Discord report (Sagebro, 2026-08-19): "När jag lägger in löner så blir det en decimal mindre än vad det ska vara. Vill jag betala ut en bruttolön på 4 531 kr måste jag skriva in 45 310 kr."

Not a decimal bug. The three employees in that company are registered with **sysselsättningsgrad 10 %** (prod: `employees.employment_degree = 10`, `monthly_salary = 1`), and the engine correctly computes grundlön = månadslön × 10 %. The row in the run only shows the monthly-salary input, so the ×10 % was invisible unless you read the formula line in Beräkningsdetaljer. The user worked around it by typing 45 310 and booked August that way (numbers are right; the snapshot just carries the inflated monthly salary).

Two other companies have employees at exactly 10 % too (Lilltuna, Dala IT), so the value is probably not intentional in any of them. I could not find a code path that turns "100" into 10 (dialog defaults to 100, min 1 / max 100; API schema 1..100; no MCP/v1 creations for these rows), so this PR makes the consequence visible instead of guessing at the cause.

## Change

- `RunEmployeesTable`: for monthly employees with degree < 100 the row prints **"× 10 % = 4 531 kr"** directly under the monthly salary, both for the editable input (draft + write) and the read-only number. Degree 100 rows are unchanged.
- `NewEmployeeDialog` and the employee edit page: helper line under Sysselsättningsgrad, "Under 100 % räknas grundlönen som månadslön × sysselsättningsgrad." (sv/en).
- New keys `salary_run.degree_hint`, `salary_employee.form_employment_degree_hint` in both message files (parity checked).

## Visual

Verified via a throwaway `/sandbox` page rendering the component with a 10 %, a 100 % and a 50 % employee (deleted before commit). Rows below 100 % get one extra 11 px muted line right-aligned under the number; 100 % rows are pixel-identical to before.

## Tests

`npx vitest run components/salary lib/salary` 594/594, eslint + tsc clean on the touched files. No migrations.


jakobwennberg-oss:
Triage of the reviewer note: `degree.toLocaleString('sv-SE')` is deliberate, not a locale leak. The amount in the same hint comes from `formatCurrency`, which hardcodes `sv-SE` app-wide (lib/utils.ts), so the degree must match it; mixing en-US `12.5` with `4 531 kr` in one line would be the actual inconsistency. Degrees are integers in practice anyway. The gross-salary hint uses the same `monthly_salary × degree / 100` + `roundOre` expression as `lib/salary/run-calculation.ts:477` and `run-employees.ts:162`, so hint and booked grundlön cannot drift.

PR #1728, opened 2026-08-19
fix(enable-banking): reconnect supersedes the old connection and stops duplicate imports

jakobwennberg-oss:
User report (2026-08-18 EF session): renewing an expired bank connection via the bank list created a NEW connection, the old one kept nagging "needs renewal" until manually removed, and duplicate transactions appeared.

## What
- Migration `20260819175522_bank_connections_superseded_by.sql`: additive `superseded_by` + `superseded_at` + partial index (status stays 'revoked' for superseded rows, no CHECK change).
- `supersedeSiblingConnections()` runs in the OAuth callback for every finalized connection: IBAN-overlap match (active siblings without overlap are never touched), park-then-revoke ordering, shared EB sessions respected (`countLiveSiblings`), `last_synced_at`/`initial_sync_*` carried onto the survivor, transactions re-pointed in batches, leftover cash_accounts demoted, `bank_connection.superseded` event emitted.
- Fresh `/connect` now 409s (`EXISTING_CONNECTION`) only when a same-bank row in `expired/error/pending_selection` exists and no `force_new: true`; active same-bank rows never 409 (second-login coexistence stays possible). #1727 wires the client dialog.
- Dedupe hardening: `dedup_scope` persisted per account and carried across reconnects by IBAN (survivor's explicit scope always wins), so uid-regenerating banks stop minting new external ids. The frozen external_id FORMAT is untouched.
- Gap-fill probe includes superseded connection ids, so a renewal defaults to "continue where the last fetch stopped" instead of a full fiscal-year re-pull. Sync toast now reports "N redan importerade hoppades över".

## Not in this PR
Prod repair of already-existing duplicate rows (needs a dry-run script, separate task).

## Verification
212 extension/route tests pass, full suite green; lint 0 errors; check:guards pass. Migration applies via the merge auto-apply flow.


john-frandsen:
Nice fix — the supersede model is the right call, because reconsent-as-new-resource is provider-level behavior: Enable Banking (like GoCardless/Nordigen agreements) issues a fresh resource id on every renewed consent, so "reuse the old connection row" can never work reliably. A few observations from wrestling with the same class of bugs on Berlin Group AIS data:

**On `dedup_scope` carried by IBAN:** that's the correct anchor, but some banks regenerate `transactionId` on *every* fetch, or when a booking flips pending→booked, and a few return different ids per PSU session even for the same booked row. `internalTransactionId` is more stable where it exists, but it's not universal. A pragmatic second layer is a composite fallback hash — bookingDate + amount + currency + normalized remittance info + counterparty account — stored alongside the external id, so uid-regenerating banks stop minting "new" rows even if the explicit scope is lost (e.g. user removes and re-adds the bank). Keep the hash independent of `transactionId` entirely, otherwise the fallback inherits the same instability you're working around.

**On the gap-fill probe defaulting to "continue where the last fetch stopped":** one caution — some banks amend already-booked transactions late (pending→booked flips, corrected booking dates, late bookkeeping by a day or two). A pure forward cursor skips those amendments. A small rolling overlap re-fetch (re-pull the last 7–14 days each sync and let dedupe absorb the rest) is cheap insurance, and it also self-heals any rows that slipped through during the supersede window.

**On the batched re-pointing:** since re-pointed `transactions` rows may already be referenced by posted entries, making the UPDATE and the `bank_connection.superseded` event emission idempotent and re-runnable in one place will save pain if the OAuth callback retries mid-batch.

PR #1729, opened 2026-08-19
fix(skattekonto): bound the sync to the first räkenskapsår, add an ignore path, EF-aware avdragen skatt

jakobwennberg-oss:
User report (2026-08-18 EF session): connecting the personal skattekonto imported transactions from before the company existed (April 2025 vs registered Nov 2025); those rows are permanently stuck (unbookable, undeletable). "Avdragen skatt" auto-mapped to 2710 Personalskatt, which is wrong for an EF without employees.

## What
1. **Fetch bound**: sync passes `datumFrom = earliest fiscal_periods.period_start`, clamped to SKV's 555-day default window (omitted when older, avoiding felkod 2 past the 915-day max and never widening the window). Boundary tests included.
2. **Ignore path**: migration `20260819080000` adds `is_ignored` to `skattekonto_transactions` (+CHECK: an ignored row can never carry a journal_entry_id) mirroring `transactions.is_ignored`. New `PATCH /skattekonto/transaktioner/:id/ignore` (409 on booked). Ignored rows leave the work-list buckets but show as an "N ignorerade" band (no silent disappearance); booking/matching an ignored row is refused up-front (`ROW_IGNORED`) before any draft is created; saldo-coverage math still includes ignored upcoming charges (SKV draws them regardless). Rows predating the first räkenskapsår get a clear message instead of the opaque PERIOD_LOCKED error.
3. **EF-aware avdragen skatt**: migration `20260819080100` adds `requires_employer` to `skattekonto_rules`, set on the avdragen-skatt seed row and clones. For an EF that is not an employer (`employer_registered ?? pays_salaries`), the 2710 suggestion is withheld (manual review + a hint that this is usually the owner's private A-skatt); employer EF and AB keep 2710; EF preliminärskatt -> 2013 regression-guarded.

## Verification
347 extension tests pass, full suite green; lint 0 errors; check:guards pass. New pg tests for the CHECK/RLS need a `npm run test:pg` run. Migrations apply via merge auto-apply.
**Visual sign-off needed**: /skattekonto Ignorera/band, /transactions inbox Ignorera, book-dialog hint.

## Summary by CodeRabbit

* **New Features**
  * Skattekonto transactions can now be ignored and restored, with confirmation dialogs, undo actions, and visibility controls.
  * Ignored transactions are grouped separately and excluded from booking and matching until restored.
  * Synchronization now includes transactions from the earliest applicable fiscal period.
  * Added clearer guidance when employer registration is required for tax handling.
* **Bug Fixes**
  * Preserved ignore decisions during synchronization.
  * Added safeguards preventing ignored transactions from being booked or matched.


jakobwennberg-oss:
### Supabase Preview is red for a stale-branch reason, not a migration defect

**What happened.** This PR first pushed the two migrations as `20260819080000` / `20260819080100`. The preview branch applied them at that point. I then renumbered the files to `20260819200000` / `20260819200100` because the originals sorted *before* prod's latest applied migration (`20260819092408`), and Supabase skips migrations older than the last one on the remote unless you pass `--include-all`. Their own docs name this exact case ("Rebase problems: timestamps getting out of order after Git rebase") and prescribe the rename I did.

The side effect: the preview branch's `schema_migrations` still lists the two old versions, which no longer have files in the repo, so branching reports MIGRATIONS_FAILED.

**Prod is not affected, and I verified that rather than assuming it.** Prod's migration list ends at `20260819092408` and has never contained `20260819080000` or `20260819080100`. Only the ephemeral preview project (`ksmxouacsgqhkgpcyzhv`, `with_data: false`) holds the orphans.

**The fix is one action I could not perform:** reset the preview branch for this PR (Supabase dashboard: Branches -> fix/skattekonto-ef-scope -> Reset, or `reset_branch` via the Supabase MCP). It re-applies from the current files and the check goes green. I was blocked from running it, so it needs a human.

**Do not fix it by reverting the rename.** That would make the preview happy and leave the real problem: on merge, both migrations sort behind prod's head and would be silently skipped, shipping the code without the `is_ignored` column or the `requires_employer` rule.

pg-real passes on this branch as of the latest push.

jakobwennberg-oss:
### Correction: the out-of-order migration was NOT skipped

I claimed above that Supabase skips a migration whose version sorts before the remote head unless `--include-all` is passed. The evidence from this batch contradicts that, so correcting it rather than leaving a wrong belief in the record.

#1733's `20260819190000_notice_dismissals` merged *after* this PR's `20260819200000`/`20260819200100` were already applied, so at apply time it sorted behind prod's head. It applied anyway. Prod's `supabase_migrations.schema_migrations` now reads:

```
20260819092408  inline_rattelse_bank_anchor
20260819175522  bank_connections_superseded_by
20260819190000  notice_dismissals          <- applied despite sorting behind the head
20260819200000  skattekonto_transactions_is_ignored
20260819200100  skattekonto_rules_requires_employer
```

So this pipeline applies any unrecorded version regardless of order. What the rename here actually fixed was the *preview branch*, which had applied the two files under their old names and then could not reconcile once they were renamed. That was real and the reset cleared it. The prod-skip risk I used to justify the rename did not exist in this pipeline.

Worth knowing for future batches: renumbering to sort after the head is harmless, but it is not required, and renaming a migration *after* a preview branch has applied it is what breaks branching.

PR #1752, opened 2026-08-20
fix(agent): show account names on in-chat booking proposals

StDl71:
## Summary
- `/pending` already loads the chart name map into `AccountNamesContext`. The agent ApprovalCard reused `OperationPreview` without that context, so kontering showed bare account numbers.
- When the context is empty, fetch `/api/bookkeeping/accounts` once and fall back to the client BAS map (`getAccountName`) so chat proposals show names.
- Pass `selectedName` into the journal combobox so a prefilled account is not displayed as a number only.

## Test plan
- [ ] Open an agent booking proposal (not on `/pending`): each journal line shows the account name, not only `1930` / `6570`.
- [ ] Open `/pending` for the same company: names still come from the existing page fetch (no regression, no wrong name).
- [ ] Manual journal combobox with a prefilled account shows the name next to the number.
- [ ] Unknown / custom account numbers still render without inventing a name.

## Summary by CodeRabbit

* **Bug Fixes**
  * Account names now display correctly in journal-entry forms, including accounts that are not yet activated.
  * Operation previews now show resolved account names across categorization, VAT, voucher, and accounting tables.
  * Line descriptions are displayed alongside account names when they provide additional information.

### DIFF
--- a/components/bookkeeping/JournalEntryForm.tsx
+++ b/components/bookkeeping/JournalEntryForm.tsx
@@ -1539,6 +1539,10 @@ export default function JournalEntryForm({
                   value={line.account_number}
                   accounts={accounts}
                   catalog={catalog}
+                  selectedName={
+                    accounts.find((a) => a.account_number === line.account_number)?.account_name
+                    ?? catalog.find((a) => a.account_number === line.account_number)?.account_name
+                  }
                   notActivatedLabel={t('account_not_activated')}
                   onChange={(num) => updateLine(index, 'account_number', num)}
                   onCommit={() => focusDebit(index)}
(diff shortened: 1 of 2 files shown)


jakobwennberg-oss:
Thanks for this. Closing as superseded: #1812 (2fd58c412) gave the chat ApprovalCard the same AccountNamesContext the /pending page uses (useAccountNamesSource() plus the provider around OperationPreview), which was the headline of this PR. The remaining OperationPreview.tsx hunks now conflict with main and would drop the voucher line_description main renders next to the account name. The selectedName prop for the journal combobox is a small standalone improvement; a fresh PR with just that against current main would be welcome.

PR #1754, opened 2026-08-20
fix(enable-banking): stop paginating empty AIS continuation pages

StDl71:
## Summary
- `getAllTransactions` / `getAllTransactionsWithRaw` follow `continuation_key` up to 100 pages. That is correct when pages have transactions.
- Swedbank historic AIS (`strategy=longest`) can return an **empty** page plus a continuation key while it prepares. Following that key spun 100 empty pages in ~25s and ingested nothing.
- Stop when a page has a continuation key but no transactions. Keep `strategy=longest` (required by existing tests and other ASPSPs). The next cron picks up whatever the bank later materializes.
- Do not remove `strategy=longest` globally.

## Test plan
- [ ] `npx vitest run extensions/general/enable-banking/lib/__tests__/api-client.test.ts`
- [ ] New case: empty `transactions` + `continuation_key` → 1 fetch, not 100.
- [ ] Existing cap test still stops at 100 when every page has a transaction.
- [ ] Existing `strategy=longest` / no-strategy fallback tests still pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved transaction retrieval when an empty page includes a continuation marker.
  * Pagination now stops promptly instead of making unnecessary follow-up requests.
  * Prevents repeated empty requests while preserving transactions retrieved in later syncs.

### DIFF
--- a/extensions/general/enable-banking/lib/__tests__/api-client.test.ts
+++ b/extensions/general/enable-banking/lib/__tests__/api-client.test.ts
@@ -167,6 +167,31 @@ describe('api-client', () => {
 
       warnSpy.mockRestore()
     })
+
+    it('stops when a page is empty but still sends continuation_key', async () => {
+      const warnSpy = vi.spyOn(console, 'warn').mockImplementation(() => {})
+      fetchSpy.mockImplementation(() => {
+        return Promise.resolve(
+          new Response(
+            JSON.stringify({
+              transactions: [],
+              continuation_key: 'keep-going',
+            }),
+            { status: 200, headers: { 'Content-Type': 'application/json' } }
+          )
+        )
+      })
+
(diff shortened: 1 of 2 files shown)


jakobwennberg-oss:
Thanks for digging into the Swedbank slow-history case. Closing without merging, with the reasoning: Enable Banking's documentation (markets/se and the FAQ section on strategy=longest) says an empty transactions list together with a non-null continuation_key means the bank has not materialised that part of the history yet and the client must keep calling until no continuation key comes back; Swedbank historic retrieval 'may take up to several minutes'. Stopping on the first empty page would turn a slow first sync into a silent partial import, and the next cron only fetches the recent window, so the gap would never be filled. If the 100-page spin is the pain point, a bounded variant would be welcome: keep following the key, cap consecutive empty pages at N with a short backoff, and log the cutoff so the next sync resumes from the same position.

PR #1755, opened 2026-08-20
fix(transactions): prefill the counter account in manual booking

StDl71:
## Summary
- **Bokför manuellt** already resolved the cash account (e.g. 1930) but left the expense/income line empty, even though `POST /api/transactions/suggest-categories` already returns template and category hits.
- Prefill the counter account from that existing endpoint. Pick the non-bank side of the top template suggestion, else the category account. Resolve against the transaction's cash account, not a hardcoded 1930.
- Does not add a "Fråga assistenten" row action or change `general.help` tools.

## Test plan
- [ ] Open **Bokför manuellt** on a transaction that matches a template: counter account is filled; bank line still uses the cash account.
- [ ] Transaction with only a category suggestion (no template): counter is the suggested account.
- [ ] Transaction with no suggestions: counter stays empty (same as today).
- [ ] Prefill does not override a selected booking template (`preselectedTemplate`).
- [ ] Cash account other than 1930: the non-bank side of the template is still the counter.

### DIFF
--- a/components/transactions/TransactionBookingDialog.tsx
+++ b/components/transactions/TransactionBookingDialog.tsx
@@ -134,13 +134,17 @@ export default function TransactionBookingDialog({
   const [inboxPickerOpen, setInboxPickerOpen] = useState(false)
   const [bankAccount, setBankAccount] = useState<string | null>(null)
   const [bankAccountName, setBankAccountName] = useState<string | null>(null)
-  const [suggestedCounterAccount, setSuggestedCounterAccount] = useState('')
+  const [categorySuggestion, setCategorySuggestion] = useState<{
+    debit_account?: string
+    credit_account?: string
+    fromCategory: string
+  } | null>(null)
 
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


jakobwennberg-oss:
Thanks for this, and sorry for the slow turnaround. Closing it in its current shape rather than merging, because the vehicle no longer exists on main: #1935 replaced the cash-account useEffect this PR extends with useCashAccounts() + useMemo, and #1889 (preselectedAccount) + #1894 (Andra rader) reshaped the counter-line prefill, so the branch conflicts in four hunks of TransactionBookingDialog.tsx. The gap itself is still real (handleManualBooking passes preselectedAccount=null, so the counter line stays empty). A rebased version that calls /api/transactions/suggest-categories from the page-level handleManualBooking and passes the hit as preselectedAccount would be very welcome; note that getSuggestedTemplates returns the enskild firma debit_account, so an aktiebolag needs debit_account_ab.

PR #1756, opened 2026-08-20
fix(cash-accounts): rebind transactions when remapping a PSD2 ledger

StDl71:
## Summary
- `upsertFromPsd2` could overflow onto a second cash account (e.g. 1931) when 1930 was already taken. Remapping the same IBAN to 1930 **demoted** the overflow row (cleared `bank_connection_id`) instead of moving the imported txs.
- AI booking reads `transactions.cash_account_id` → `cash_accounts.ledger_account`, so those rows kept proposing 1931 after the UI showed 1930.
- Rebind `cash_account_id` onto the promoted row, then delete the duplicate. Rebind first so `ON DELETE SET NULL` cannot orphan the txs. Delete even when there are no linked txs (same leftover-mirror cleanup as before).

## Test plan
- [ ] `npx vitest run lib/cash-accounts/__tests__/service.test.ts`
- [ ] Duplicate with linked txs: txs move to the promoted cash account; overflow row is deleted (not demoted to a manual leftover).
- [ ] Duplicate with no txs: overflow row is still deleted.
- [ ] Primary flag still transfers to the promoted row.

### DIFF
--- a/lib/cash-accounts/__tests__/service.test.ts
+++ b/lib/cash-accounts/__tests__/service.test.ts
@@ -475,6 +475,8 @@ interface UpsertStub {
   rpcCalls: Array<{ fn: string; args: Record<string, unknown> }>
   /** .eq() filters applied to the linked-transactions probe. */
   transactionFilters: Array<{ col: string; value: unknown }>
+  /** Rebind writes from the overflow duplicate onto the promoted row. */
+  txRebinds: Array<Record<string, unknown>>
 }
 
 function makeUpsertStub(partial: Partial<UpsertStub> = {}): UpsertStub {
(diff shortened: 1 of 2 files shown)


jakobwennberg-oss:
Thank you for this fix, and sorry it took a while. The rebind idea was right; what held it back was that the unconditional rebind also moved booked and voucher-anchored rows (whose vouchers sit on the old 19xx account) and deleted a duplicate that the twin logic from #2010 may still need. It landed as #2023 (squash cc7050f6c's successor on main) with your commit cherry-picked and credited, plus a movable-rows guard mirroring #1570 and the demote-or-delete rule kept for rows that stay bound. Closing this one as superseded.

PR #1760, opened 2026-08-20
perf(agent): drop extended thinking on /chat and the empty auto-fire on open (audit quick wins)

jakobwennberg-oss:
## Why

WS1 **rip track, RIP-1** (your AI surface audit, 2026-08-18). Two independent, low-risk quick wins that trim the heavy chat runtime ahead of the full rip and align it with the local/OpenAI-compatible backends the assistant is moving to. Off `main` (no dependency on the AI-provider PRs).

## What
- **`general.help` no longer runs adaptive thinking.** It was the single biggest latency source on the broad `/chat` surface; adaptive thinking doesn't exist on OpenAI-compatible/local backends; and the ground rules already forbid answering regulatory questions from memory. Backend-only change.
- **`AgentChat` no longer auto-fires an empty first turn on open.** Opening `/chat` with nothing to say spent a model call generating a greeting, and per the audit 27-75% of those conversations were never typed in. The explicit **seed-message path** (an "ask about this" starter the user chose) still fires; the empty case opens ready for input.

## Please review the one visible change
The auto-fire change is **visible** — the chat opens without an auto-greeting. There are no component tests for `AgentChat` (Vitest scope is `lib/` + `app/api/`), so **please eyeball `/chat` open on the preview before merging** (per the standing UI sign-off rule). Everything else is backend and tested.

## Verification
- `general.help` read-only + redirect guards and the `run-turn` thinking guards unchanged and green; `lib/agent` suite green; `tsc`, `check:guards`, `check:lint` clean. No migrations.

## Summary by CodeRabbit

* **Bug Fixes**
  * Empty chat sessions no longer automatically send an empty message or start an unnecessary agent turn.
  * Chat sessions now begin with a response only when an initial message is provided.

* **Improvements**
  * General help responses now use a streamlined single-response experience without extended thinking.

### DIFF
--- a/components/agent/AgentChat.tsx
+++ b/components/agent/AgentChat.tsx
@@ -296,19 +296,21 @@ export default function AgentChat({
     // the composer is already replaced by the upgrade note.
     if (!hasAi) return
 
-    // Seed-message path: render the user's pre-baked starter in the timeline
-    // and send it as the first turn's user_message (skips intent.capture +
-    // promptTemplate). Empty seed runs the normal capture-driven flow.
+    // Seed-message path: the user arrived with a pre-baked starter (e.g. an
+    // "ask about this" button), an explicit intent to ask something: render it
+    // and send it as the first turn's user_message. This still auto-fires
+    // because the user chose to.
+    //
+    // No EMPTY auto-fire (audit 2026-08-18): opening /chat with nothing to say
+    // used to fire a full turn to generate a greeting, and 27-75% of those
+    // conversations were then never typed in: a model call and a wait for a
+    // greeting nobody asked for. The chat now opens ready for input; the first
+    // turn runs when the user actually sends something.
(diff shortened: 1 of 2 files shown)


jakobwennberg-oss:
Closing: the goal shipped through the RIP-3 cutover (#1759 a071a0220, #1762 3f6f1ab06, #1769 148ec0ce8). On main every general.help entry renders AskConsole, which never auto-fires an empty turn and does not use extended thinking. The AgentChat hunk here would now regress main: AgentChat is mounted fresh with no seedUserMessage for all openAgentSheet intents (verifikation.draft, inbox.bulk-book, transaction.categorization, onboarding.intake), and the empty-first-turn branch this PR removes is what starts them. The dead thinking config line in lib/agent/intents/general-help.ts can go in a one-line follow-up if wanted.

PR #1763, opened 2026-08-20
fix(import): let provider re-sync re-import an earlier fiscal year after data deletion

### LINKED ISSUE #1667: Fix provider re-sync failing to re-import an earlier fiscal year after deletion
**Priority:** medium

After partially deleting imported data, re-running the Fortnox sync could not bring back the previous fiscal year: the `sie_imports` completed-status watermark survives data deletion, the replace path aborts the whole year if the prior import row cannot be resolved, and a leftover fiscal period with residual entries or a lock throws an overlap error. The result is a customer stuck between states: the current year re-imports, the prior year refuses.

## Relevant code
- `lib/import/sie-import.ts:142-162`: `checkDuplicatePeriodImport` watermark survives manual deletion
- `lib/import/sie-import.ts:1994-2038`: replace-mode path aborts the year when `replaceSIEImport` fails
- `lib/import/sie-import.ts:378-444`: `ensureFiscalPeriod` overlap error for leftover periods
- `extensions/general/arcim-migration/index.ts:894-970`: prior-import detection uses `.limit(1).maybeSingle()` with no ordering; multiple overlapping rows break it
- `extensions/general/arcim-migration/lib/sie-fetcher.ts:104-133`: per-year fetch failures land in `failedYears` quietly

## Next steps
- Make the replace path resilient when the referenced prior import's data is already gone (treat as a fresh import)
- Order and handle multiple overlapping `sie_imports` rows for the same year
- Surface per-year failures loudly in the wizard instead of a quiet failedYears list
- Reproduce: import year N and N-1, delete parts of N-1's data, re-sync

---
**Gmail thread ID:** `1a00fc17693482a1`

mattssonn:
# What and why

Closes #1667. After deleting a year's data, provider re-sync refused to re-import an earlier fiscal year: the replace pre-check only looked at the single newest overlapping import row. `findOverlappingPeriodImports` now returns ALL completed overlapping imports newest-first, and `replaceSIEImport` returns typed codes (`not_found` / `not_completed` / `rpc_error`). The pre-check fails closed: only PGRST116 (zero rows) maps to `not_found`; any other query error aborts with "Kunde inte läsa tidigare import" instead of risking duplicate verifikationer (BFL 4:1). The Arcim migration workspace labels per-year errors.

## Checklist

- [x] Unit tests pass locally (14 tests, `sie-import-resync.test.ts`)
- [x] No UI string changes beyond existing message keys
- [x] No migrations
- [x] Core builds with zero extensions enabled
- [x] Commits are signed off

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved multi-year SIE import handling by resolving all overlapping completed imports in the correct order.
  * Stale or incomplete import records are now skipped safely, while locked periods and genuine import failures still stop processing.
  * Import errors now identify the affected fiscal year and file.
  * Replace-mode imports report combined deleted-entry totals and the latest replaced import.

* **Documentation**
  * Documented provider re-sync behavior for overlapping and stale imports.


mattssonn:
Review findings addressed in ab360745b:

- **Fail closed + fetch every overlap** (CodeRabbit): `findOverlappingPeriodImports` now goes through `fetchAllRows` — query errors throw instead of returning `[]`, pagination passes the PostgREST row cap, and an `id` tiebreak keeps the order total.
- **Stale-watermark skip vs surviving entries** (compliance review): the skip is no longer trusted on its own. `replace_sie_import` deletes by `(company, fiscal_period, source_type='import')`, so entries can indeed outlive their `sie_imports` row — the loop now positively verifies zero posted import entries remain in the fiscal year before proceeding, and aborts (fail closed) on survivors or a failed check. Tests added for both paths.
- **Regex/RPC wording coupling**: contract comment added; note the failure mode of a wording drift is fail-closed (stale race reclassified as rpc_error → abort), never fail-open.
- **beforeEach hygiene**: suite now clears mocks and the event bus.

The behandlingshistorik point (whatever deletes `sie_imports` rows elsewhere should soft-transition, not hard-DELETE) is real but out of scope for this PR — it concerns the deletion path, not the re-sync path.

PR #1772, opened 2026-08-20
fix(import): let provider re-sync re-import an earlier fiscal year after data deletion

### LINKED ISSUE #1667: Fix provider re-sync failing to re-import an earlier fiscal year after deletion

mattssonn:
## What

Lets a provider re-sync re-import an earlier fiscal year after the user has partially deleted imported data, instead of aborting on a stale `sie_imports` watermark.

- `findOverlappingPeriodImports` returns ALL overlapping completed rows, newest first; `checkDuplicatePeriodImport` now picks deterministically.
- `executeSIEImport` replace mode resolves every overlapping row. A row that is gone or no longer `completed` (`replaceSIEImport` codes `not_found` / `not_completed`) is a stale watermark: skip it with a warning and import the year fresh instead of stranding the user. Locked/closed periods and RPC failures still abort the year.
- The arcim-migration wizard names the fiscal year in every per-file import failure and shows the newest prior import in the options step.
- Fail closed on pre-check errors: `replaceSIEImport`'s pre-check previously discarded the `.single()` error, so a transient query failure (statement timeout, network error, PostgREST 5xx) was indistinguishable from a genuinely absent row and got classified `not_found`, letting the replace loop import a year fresh while the prior import's verifikationer were still in the ledger (silent duplicates, BFL 4:1 risk). Only PGRST116 now classifies as `not_found`; any other pre-check error returns `rpc_error`, which aborts the year.

## Why

After partial data deletion, the `sie_imports` 'completed' watermark survives, the replace path aborted the whole year when the prior import row could not be resolved, and prior-import detection picked an arbitrary row when several overlapped the same year.

Fixes #1667

## Tests

New 351-line suite `lib/import/__tests__/sie-import-resync.test.ts`: stale-watermark skip vs abort classifications (`not_found` / `not_completed` / locked / `rpc_error`), deterministic newest-first prior-import selection, pre-check error fail-closed classification, and the `executeSIEImport`-level abort on `rpc_error`.


mattssonn:
Superseded: the identical fix (same branch content) merged as PR #1763 on 2026-08-20 and issue #1667 is closed.

PR #2044, opened 2026-08-30
feat(salary): recurring payroll lines per employee (#2042)

### LINKED ISSUE #2042: feat(salary): recurring payroll lines per employee (e.g. förmånscykel bruttolöneavdrag)
## Problem

A standard Swedish payroll setup today is a benefit bike via gross salary deduction (bruttolöneavdrag): the employee's payslip carries a recurring line like `Förmånscykel bruttolöneavdrag -670,17 kr` every month. Bokio, Fortnox and Visma all support recurring payslip rows.

Accounted has no recurring-line mechanism. `employee_benefits` only derives **positive taxable benefit rows** (`is_taxable: true`, `is_gross_deduction: false`, filter `monthly_value > 0` in `lib/salary/run-calculation.ts` step 8d). A recurring gross deduction must be re-added by hand as a manual `gross_deduction_other` line on every single run, which is easy to forget and silently produces a wrong payslip (overstated gross, tax and AGA).

## Proposal

A per-employee recurring line registry, mirroring the `employee_benefits` pattern:

- New table `employee_recurring_lines`: `item_type` (SalaryLineItemType subset), `description`, `amount` (signed), `is_taxable`, `is_avgift_basis`, `is_vacation_basis`, `is_gross_deduction`, `is_net_deduction`, optional `account_number` override, `valid_from`/`valid_to`, `is_active`, RLS by company.
- Derivation step 8e in `run-calculation.ts`: delete + re-derive rows with a `source_recurring_line_id` marker, same lifecycle as the benefit rows (survives recalculation, excluded from manual-line editing guards).
- CRUD API under the employee settings surface + UI section next to benefits.
- Validation: gross deductions must be negative, benefits-style positive rows rejected (use `employee_benefits` for those), net deductions only for net item types.

## Real-world case

Devmera AB (self-hosted): benefit bike leased via Bike Mobility/PEAC, 670,17 kr/month gross deduction, bike benefit itself tax-free (< 3 000 kr/year). Currently handled via a remembered manual step per run.

I'll follow up with a PR.

joakimhew:
Closes #2042.

## What

A standing per-employee payslip row that is derived into every salary run inside its validity window: the canonical case is a benefit-bike bruttolöneavdrag (-672,81 kr/month) that previously had to be re-added by hand on every run and silently produced a wrong payslip when forgotten.

Mirrors the `employee_benefits` pattern end to end:

- **Migration** `20260830140000_employee_recurring_lines.sql`: table with company RLS (all four policies), audit + updated_at triggers, validity/amount-sign/account-format CHECKs, and a `salary_line_items.source_recurring_line_id` back-link (partial index), mirroring `source_benefit_id`.
- **Engine** (`lib/salary/run-calculation.ts` step 8d3): delete-by-back-link + re-derive, exactly like the benefit rows; derived rows are excluded from the manual-line set in step 8e. Flags are computed from the item type (`lib/salary/recurring-lines.ts`): gross deductions reduce the tax and AGA bases (sick-karens convention), net deductions are post-tax only, so a stored row can never contradict the payslip math. `valid_to` is filtered in JS so the phantom-column scanner resolves the query.
- **API**: `GET/POST /api/salary/employees/[id]/recurring-lines` + `PATCH/DELETE .../recurring-lines/[lineId]` with the same 401/403/404/400 contract as the benefits routes (PGRST116 vs 23514 mapping, merged-state period validation, amount-sign re-check against the stored item type).
- **UI**: `EmployeeRecurringLinesPanel` on the employee page below the benefits panel; deductions are entered as positive numbers and stored negative. sv/en strings added; run-detail `li_*` labels already existed for all item types.
- **BFL**: table registered in the full-archive export.

## Item types

Reuses existing `SalaryLineItemType` members (no CHECK expansion on `salary_line_items`): `gross_deduction_pension`, `gross_deduction_other`, `net_deduction_union`, `net_deduction_benefit_payment`, `net_deduction_other`, `other` (recurring addition).

## Tests

29 new tests: pure flag/sign helpers, POST route (401/403/201/404, amount-sign 400s, period-order 400s, 23514→400 backstop, 500 passthrough), PATCH/DELETE route (404-vs-400 mapping, merged-period check, sign-vs-stored-type check). Phantom-column ceiling raised 388→390 with a dated reason (two variable-payload shapes the benefits code already carries).

`npm run lint`, `npm run check:types`, `npm run check:guards` and the full `npm test` suite pass.

## Summary by CodeRabbit

* **New Features**
  * Added recurring salary deductions to employee detail pages.
  * Create, view, update, and delete or deactivate recurring deductions.
  * Automatically includes active recurring deductions in salary calculations.
  * Added validity dates, amount validation, localized labels, and feedback messages.
  * Included recurring salary deductions in full master-data exports.

* **Bug Fixes**
  * Recurring deductions already used in salary calculations are safely deactivated instead of deleted.


jakobwennberg-oss (dismissed):
Thanks for this, it closes a real gap: a manual `gross_deduction_other` line added through the API inherits `is_gross_deduction=false` and `is_vacation_basis=true` from the schema defaults, so today it is not deducted at all. Mirroring `employee_benefits` is the right call, the composite FK closes the IDOR, and the gross deduction flows correctly into the tax base, the AGA base and AGI (`generate-declaration.ts:360` reads `gross_salary` / `avgifter_basis` from the engine output, so no extra mapping was needed).

Verified locally on 1dda3d2db: `npx vitest run lib/salary app/api/salary` 67 files / 873 tests green, eslint clean on the changed files, `check:guards` green, no new type errors.

## Blocker

1. **`other` (recurring addition) does not work.** `lib/salary/calculation-engine.ts` only treats `ADDITION_TYPES` (overtime / OB / bonus / commission, lines 321-328) as additions and never reads `isTaxable` / `isAvgiftBasis`. A positive `other` row is therefore not added to gross, tax base, AGA base or net. Booking (`lib/salary/salary-entries.ts:257-262`) still posts it to 7210 and shrinks the 7010 remainder. Net effect: the payslip shows "+500 kr Återkommande tillägg", the employee is not paid it, AGI omits it, and 500 kr of base salary is relabelled as 7210. The `recurring_addition_hint` string ("Läggs till som skattepliktigt tillägg varje månad") is therefore false. Please drop `other` from `RECURRING_LINE_ITEM_TYPES`, the migration CHECK and the panel. Teaching the engine to treat `other` as an addition is a separate change with its own engine tests.

## Before merge

2. **Rebase.** The only conflict is `tests/schema/no-phantom-columns.test.ts`. Both sides raised the ceiling; the merged tree measures 395 (I ran the scanner on the merge), not 393, so re-measure rather than sum the two deltas.
3. **DoD 3:** a `tests/pg/*.pg.test.ts` for the new RLS policies, the CHECKs and the composite FK (viewer read-only, cross-company insert refused, sign CHECK).
4. **One engine-level test** that runs a derived `gross_deduction_other` row through `calculateSalary` and asserts gross, tax base and `avgifter_basis` drop. The route tests cover plumbing; nothing pins the payroll math today, which is exactly the regression CodeRabbit caught in the first round.
5. **Delete race** (CodeRabbit thread still open): cheapest race-free fix is `ON DELETE NO ACTION` on `salary_line_items.source_recurring_line_id` and let the DELETE route fall back to deactivate on 23503. That also settles the provenance thread at the DB layer.
6. **Correction runs double the deduction.** `app/api/salary/runs/[id]/correct/route.ts:141-157` copies `salary_line_items(*)` without `source_benefit_id` / `source_recurring_line_id`, so recalculating a correction run keeps the copied row as "manual" and step 8d3 derives it again. Pre-existing for benefits, but this PR makes it the common case. Copy the back-links, or skip derived rows in the copy.
7. **Record the vacation-base choice** in `DECISIONS.md`: a bruttolöneavdrag does not reduce `is_vacation_basis` here, which diverges from the manual-line default and is a judgment call worth pinning.

## Nice to have

- Round `amount` on POST (a 3-decimal input is stored as-is today).
- The single-column `employee_id` FK is redundant next to the composite one; the dimensions tables keep only the composite.
- `lib/api/schemas.ts` still references migration 20260830140000.
- Panel: no edit (PATCH exists), and deactivated lines vanish with no way to see or reactivate them.
- Help text on the window semantics: inclusive against `payment_date`, no proration.

Migration version 20260831210000 is fine: no collision, and prod applies out-of-order versions.

PR #2101, opened 2026-09-01
feat(invoices): grouped sections and a grouping picker on the customer invoice list

joakimhew:
## What

The customer invoice list gets the sectioned layout the supplier list already has, plus an explicit grouping picker in the toolbar:

- **Default (Status)**: rows group into Utkast / Väntar på betalning / Betalda och avslutade sections with counts in the headers. Headers only render when more than one section has rows.
- **Kund / Månad / Ingen gruppering**: alphabetical customer groups, month groups (localized names, newest first), or the flat list. The choice is written back to the URL as `?group=` so views stay shareable and survive back-navigation.
- The column sort applies **within each section** and now cycles asc / desc / default (tri-state, matching the supplier list) so an applied sort can be released back to the invoice-date order.
- Paging and the detail pager (prev/next) follow the rendered group order.

## How

Rows are bucketed after search + fiscal-year scope + status view + sort, so section counts always equal what the section shows and an active sort governs order inside every bucket. Strings live in both `messages/sv.json` and `messages/en.json` under the `invoices` namespace.

## Testing

`npx eslint` clean on the page, `check:types` at baseline, full invoice suites green. Running in production on a self-hosted instance since 2026-08-31.

## Summary by CodeRabbit

* **New Features**
  * Added invoice list grouping by status, customer, month, or no grouping.
  * Added a grouping control to the invoice toolbar.
  * Added localized group labels and section headings in English and Swedish.
  * Grouped invoice lists now work with sorting, pagination, and detail navigation.

* **Improvements**
  * Column sorting now cycles through ascending, descending, and default order.


jakobwennberg-oss (dismissed):
Thanks. Two things need fixing regardless of what we decide about the feature:

1. **This reverts #2093.** `origin/main` keeps the row checkbox visible at muted opacity via `CHECKBOX_REVEAL_CLASS` (`opacity-50 ... group-hover:opacity-100`). The PR drops the import and inlines the old `opacity-0 group-hover:opacity-100 ...` classes, so checkboxes go back to hidden-until-hover. The same region now also conflicts with #2117 (shift-click range selection), so a rebase is needed anyway; keep `CHECKBOX_REVEAL_CLASS` and `useRangeSelect` from main.
2. **`'—'` placeholders** (`?? '—'`, `key !== '—'`): the repo bans em / en dashes in code (CLAUDE.md). Use an i18n string or a plain hyphen.

On the feature itself: the premise that the payment queue gets buried is weak. The status tabs (Obetalda / Förfallna / Utkast) already isolate it, and Fortnox, our default UX reference, keeps lists flat with filters. Kund / Månad grouping is a genuine convenience, so I am fine taking it as an opt-in, but not as the default view for every user:

3. **Default to `none`** (today's flat list) and keep `status` / `customer` / `month` as explicit choices. A sectioned default on the most-used list page is a design change that needs a visual pass first.
4. **Extract the bucketing.** The ~90 lines are duplicated verbatim in #2102. A pure helper, e.g. `lib/lists/group-rows.ts` (`groupRows(rows, keyOf, order)` returning flat rows + section meta) with vitest coverage, lets both pages render sections from one implementation and gives the feature the tests DoD 2 asks for.
5. **One "awaiting" predicate.** `statusGroupOf` uses `sent | overdue | partially_paid`, `matchesListTab('unpaid')` on the same page uses `sent | overdue`. Two predicates for the same concept will drift; derive one from the other.

Small: the section header `<tr>` inherits the `stagger-enter` animation with the data rows.

PR #2102, opened 2026-09-01
feat(supplier-invoices): payment-queue sections and a grouping picker

joakimhew:
## What

Companion to #2101 for the supplier invoice list:

- **Default (Status)**: the list groups into Väntar på betalning (registered / approved / overdue) on top and Betalda och avslutade below, with counts in the section headers, so the payment queue is never buried under settled rows. With no column sort active each section runs newest first.
- **Leverantör / Månad / Ingen gruppering**: alphabetical supplier groups, month groups (localized names, newest first), or the flat list. Written back to the URL as `?group=`.
- The tri-state column sort from #2091 governs order **inside** each group; the sections and the sort compose instead of competing.
- The detail pager follows the rendered group order.

## How

Buckets are built from the filtered + sorted rows, so counts match the visible rows and sort order carries into every group. Strings in both message catalogs under `supplier_invoices`.

## Testing

`npx eslint` clean on the page, `check:types` at baseline. Running in production on a self-hosted instance.

## Summary by CodeRabbit

- **New Features**
  - Supplier invoices can now be grouped by payment status, supplier, month, or displayed without grouping.
  - Grouped sections preserve invoice ordering and support selection and detail navigation.
  - Added a grouping selector and a link to the Payment files page.
- **Localization**
  - Added English and Swedish labels for grouping options, section headings, payment statuses, and missing-group fallbacks.
  - Month names are formatted according to the selected locale.

### DIFF
--- a/app/(dashboard)/supplier-invoices/page.tsx
+++ b/app/(dashboard)/supplier-invoices/page.tsx
@@ -1,9 +1,9 @@
 'use client'
 
-import { useState, useEffect, useRef } from 'react'
+import { Fragment, useMemo, useState, useEffect, useRef } from 'react'
 import dynamic from 'next/dynamic'
 import { useRouter, useSearchParams } from 'next/navigation'
-import { useTranslations } from 'next-intl'
+import { useLocale, useTranslations } from 'next-intl'
 import { Skeleton } from '@/components/ui/skeleton'
 import { Badge } from '@/components/ui/badge'
 import { Button } from '@/components/ui/button'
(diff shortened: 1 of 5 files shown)


jakobwennberg-oss (dismissed):
Same two fixes as #2101 first:

1. **Reverts #2093**: `CHECKBOX_REVEAL_CLASS` is replaced with the old hidden-until-hover classes. Rebase onto main (the same region conflicts with #2117) and keep the shared class.
2. **`'—'` placeholders**: em / en dashes are banned in code here (CLAUDE.md).

Feature-specific:

3. **Default order changes.** With `status` grouping and no column sort the list is re-sorted by `invoice_date desc, arrival_number desc` (`byNewestFirst`), replacing the API default of förfallodatum stigande. For a payment queue, due-date ascending is the order that matters: what is due first goes first. Keep the API order inside "Väntar på betalning".
4. **`colSpan={9}` is hardcoded**; the table has 8 columns when `canWrite` is false (no selection column). Use `canWrite ? 9 : 8`.
5. **`isAwaitingPayment` omits `partially_paid`**; `PAYABLE_STATUSES` in `lib/invoices/bulk-reconcile-supplier-vouchers.ts:60` includes it. A partially paid invoice belongs in the queue.
6. Same asks as #2101: default `none`, and one shared tested helper instead of a second copy of the bucketing. The IIFE inside the `<Fragment>` for the header row is harder to read than the index-based `prevKey` in #2101; pick one shape for both pages.

PR #2432, opened 2026-09-08
chore(scripts): date-only backfill for receipts left without invoiceDate

jakobwennberg-oss:
## What

Companion to #2429, which fixed the extraction prompt so receipts get their purchase date. That fix only helps documents read after it shipped. This script re-reads the receipts extracted before it and fills in the one missing field.

Scope on prod as of 2026-09-08: 1 313 receipts, all status `received`, none booked, all with a document, across 46 companies.

## First principles

1. **Why the problem occurred.** Same mechanism as #2429: the prompt told the model the date was invoice-only. The already-extracted rows are the residue of that prompt and nothing re-reads them on its own.
2. **What could be removed or simplified instead.** The obvious tool is the existing per-item "Tolka om" action, but it overwrites `extracted_data` wholesale, which would silently discard manual edits (typed totals, corrected suppliers) across 46 companies' inboxes. A date-only write is the smaller, safer operation. Stamping the upload date or today instead of re-reading was rejected: it invents a business-event date, which BFL 5 kap 6-7 § does not allow.
3. **Why this shape.** One field, written only where it is still null at write time, so a user typing the date by hand mid-run wins. Idempotent and resumable, dry-run by default, `--env` mandatory so the target database is a deliberate choice. Booked items are excluded (their verifikat already has the date), sandbox companies are excluded. The document-level mirror gets the same null-guarded patch, and each write appends a `DocumentExtractionRetried` processing-history event (BFNAR 2013:2 p. 9.16 trail).

## Verification

- eslint clean; type ratchet shows only the pre-existing `@types/nodemailer` miss in an unrelated module.
- Dry run against prod, three items, no writes: 3/3 would be dated. The receipt from the original report (date printed twice, field empty) resolves to 2026-06-08, exactly what is printed.
- Plausibility guard: real calendar date, not before 2015, not after the day the document reached the inbox.

## Execution status

EXECUTED on prod 2026-09-08, 17:20 to 17:55 UTC, on the founder's instruction, in five harness-tracked runs of `--apply --limit 400 --concurrency 6` (the last one a retry over the leftovers).

| Outcome | Items |
|---|---|
| Dated | 1 276 |
| No date readable even with the fixed prompt (stable across two runs) | 35 |
| Implausible date rejected (read as 2005 on a 2026 upload) | 1 |
| Errors, download failures, skipped media | 0 |

Verified afterwards by SQL: the undated-receipt count went from 1 313 to 36, exactly 1 276 `DocumentExtractionRetried` history rows carry `backfill = receipt-purchase-date`, and every backfilled item's document-level mirror carries the same date. 69 mirrors that still lack a date all belong to items the script never touched (dated by hand in the UI before this run, which does not update the mirror); pre-existing, out of scope.

Re-running is safe and would only re-read the 36 leftovers.

https://claude.ai/code/session_015bTBgZrofpGCgSUfAKN2H3

### DIFF
--- a/scripts/backfill-receipt-purchase-date.ts
+++ b/scripts/backfill-receipt-purchase-date.ts
@@ -1,0 +1,331 @@
+#!/usr/bin/env npx tsx
+/**
+ * Backfill invoice.invoiceDate on receipts that were extracted before
+ * PR #2429 (2026-09-08).
+ *
+ * Until that PR the extraction prompt described invoice.invoiceDate as a
+ * bare ISO date under the invoice block, next to a purchaseTime rule marked
+ * "receipts only", and the model returned null for the date on roughly half
+ * of all receipts (three quarters of the WhatsApp ones). Those rows never
+ * re-extract by themselves: the prompt fix only helps documents read after
+ * it shipped.
+ *
+ * DATE-ONLY BY DESIGN. The script re-reads the stored document with the
+ * fixed prompt and writes exactly one field, invoice.invoiceDate, and only
+ * where that field is still null at write time. Nothing else in
+ * extracted_data (manual edits, matched supplier, totals, line items) is
+ * touched. The per-item "Tolka om" action overwrites extracted_data
+ * wholesale, which is fine for one item a user is looking at and wrong for
+ * a bulk run over 46 companies' inboxes.
+ *
+ * Idempotent and resumable: a second run finds nothing to do for rows that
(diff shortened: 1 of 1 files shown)


jakobwennberg-oss:
Closing: the backfill has to re-read each receipt with the model (about 1,300 calls), because the date was never extracted the first time. Not worth it for a field new receipts now get right; affected receipts can use 'Tolka om' individually.

PR #2467, opened 2026-09-10
feat(settings): opt-in auto-approve for supplier invoices

olofpinzke:
## Summary
- Adds `company_settings.auto_approve_supplier_invoices` (default `false`) so limited companies can skip the separate Approve (payment attest) step after registering a supplier invoice.
- Adds an Invoicing settings toggle and wires the dashboard create flow to call `POST /approve` when `isEF || setting`.
- Sole traders keep always-on auto-approve in the dashboard; aktiebolag keep the review dialog and only gain optional auto-attest after register.
- Server create behaviour is unchanged: invoices stay `registered` on create, and approve remains a separate endpoint (v1/MCP callers are unaffected).

## Why this problem occurred
Payment attest (`approved_at`) was only auto-applied for enskild firma via a hardcoded client path. Aktiebolag had no company setting to opt into the same post-register behaviour, so every invoice required a separate Godkänn click even when the registerer and attester are the same person in practice.

## What could be removed or simplified instead
Considered baking approve into the create API when the setting is on. Rejected: that would change v1/MCP semantics and add a second write path for attest. Reusing the existing approve endpoint keeps one attest path and leaves API callers explicit.

Also considered making the flag control EF as well (default true via backfill). Rejected: EF already behaves correctly client-side; gating EF on the column would need a backfill or become a breaking change for companies that never open settings.

## Why this solution
A boolean setting plus `shouldAutoApprove = isEF || setting` preserves current EF UX without a backfill, gives AB an opt-in, and keeps create/approve boundaries intact for API surfaces.

## Test plan
- [ ] Enable the toggle under Settings → Invoicing; confirm it persists after reload
- [ ] As AB with the setting **off**: register a supplier invoice and confirm it stays unapproved until Godkänn
- [ ] As AB with the setting **on**: register (including after the review dialog) and confirm it is approved without a separate Godkänn click
- [ ] As EF: register still auto-approves with the setting off
- [ ] Privately paid / person-payer path still skips approve
- [ ] Register-and-match (bank pick) still works for EF and for AB with the setting on
- [ ] `npx vitest run lib/supplier-invoices/__tests__/auto-approve.test.ts components/supplier-invoices/__tests__/supplier-invoice-defaults.test.ts lib/pending-operations/schemas/__tests__/company-settings.test.ts`

## Not live yet
- Migration `20260910105521_company_settings_auto_approve_supplier_invoices.sql` still needs to be applied to staging/prod after merge
- Feature is unreachable in hosted until that migration is on the target database

## Summary by CodeRabbit

- **New Features**
  - Added an invoicing setting to optionally auto-approve supplier invoices immediately after registration.
  - Limited companies retain manual approval by default; sole proprietorships continue to auto-approve, with the setting unavailable.
  - Added clear feedback when saving the setting fails.
- **Documentation**
  - Changes to this setting are now included in the processing history.
- **Tests**
  - Added coverage for default settings and approval behavior.


jakobwennberg-oss:
Thank you! Closing because the supplier invoice flow is about to be reworked as a whole. The opt-in auto-approve idea is noted as an input to that rework rather than landed on the current flow.

PR #2496, opened 2026-09-11
feat(zettle): editable store display name for Orders Butik

olofpinzke:
## Summary
- Zettle's `users/self` only returns `organizationUuid`, so Orders showed that UUID under **Butik** (`store_label || store_scope`).
- New connections default `organization_name` to `Zettle` on OAuth activate; sync uses the same default when the stored name is still null.
- Settings panel adds an editable store name (never seeded with the UUID). `POST /organization-name` saves it and backfills `webshop_orders.store_label` for that connection.

## Why this problem occurred
The column and UI already preferred `organization_name`, but the OAuth callback always wrote `null`, so sync left `store_label` null and the UI fell back to the org UUID.

## What could be removed or simplified instead
Considered scraping an undocumented Zettle org-profile API. Rejected: no documented name field; brittle vs a merchant-owned label. Also considered only changing the Orders fallback string without persistence; rejected because filters/lists would still key off the UUID visually.

## Why this solution
Accounted already stores `organization_name` and copies it to `store_label` on sync. Owning the label (default + rename + backfill, and never treating the UUID as a display name) fixes existing and future rows without depending on Zettle for a name.

## Test plan
- [ ] Connect (or use existing) Zettle: settings show a store name field defaulting to `Zettle`, not the UUID
- [ ] For a connection with `organization_name = null`, save `Zettle` (or a custom name): Orders **Butik** updates for existing rows
- [ ] Rename to e.g. "Café Norr": Orders and settings show the new name
- [ ] Empty name is rejected
- [ ] New OAuth connects default the label to `Zettle` until renamed
- [ ] `npx vitest run extensions/general/zettle/__tests__/organization-name.test.ts extensions/general/zettle/__tests__/api-routes.test.ts extensions/general/zettle/__tests__/order-sync.test.ts app/api/extensions/zettle/callback/__tests__/route.test.ts`

## Not live yet
- Rows already synced with `store_label = null` keep showing the UUID until the merchant saves a name (backfill) or the next purchase sync rewrites `store_label` with the default/custom name.

## Summary by CodeRabbit

- **New Features**
  - Added the ability to edit the Zettle store display name from settings.
  - Store names are validated, normalized, and limited to 80 characters.
  - Existing order labels are updated to reflect renamed stores when possible.
  - New Zettle connections default to the display name “Zettle.”
  - Store labels no longer fall back to displaying an organization UUID.

- **Bug Fixes**
  - Improved store-name fallback behavior.
  - Added warnings when existing order labels cannot be updated automatically.

- **Localization**
  - Added English and Swedish translations for store-name management.


jakobwennberg-oss:
Merged origin/main (8ab784752; conflicts were imports and i18n keys next to main's new backfill feature, both kept) and addressed the Major review thread in 68397071b: the rename and the store_label backfill now run in one security-invoker RPC, rename_zettle_store (migration 20260923170000). The connection UPDATE holds the row lock while the orders are backfilled, so concurrent renames serialize; a failed backfill runs in its own savepoint, so the rename still commits and the route keeps its warning response. New pg-real test tests/pg/zettle-rename-store.pg.test.ts covers the backfill scope, the outsider case, and two concurrent renames. Unit tests updated for the RPC.

PR #2502, opened 2026-09-11
feat(ui): make the ⌘K search visible and derive its entries from the nav, report catalog and settings

jakobwennberg-oss:
## What

The command palette (⌘K / Ctrl+K) now has a visible way in, and its entries are derived from the same sources the sidebar uses instead of a hand-kept list.

- **Trigger**: a "Sök ⌘K" row under the brand in both sidebars (v2 and v1, plus a rail icon when v1 is collapsed) and a "Sök" row at the top of the mobile menu sheet. The chord label is ⌘K on Apple platforms and Ctrl K elsewhere (`useSyncExternalStore`, so SSR and hydration agree). Because `LazyCommandPalette` mounts the bundle on first use, triggers dispatch a window event that mounts or opens it.
- **Entries**: pages come from `NAV_V2_TOP` / `NAV_V2_COMPANY` through `passesNavGates` (extracted from `DashboardNav` into `components/dashboard/nav-gates.ts`, used by both), reports from `REPORT_CATALOG` (a report that already is a page becomes an alias of that page instead of a second row), settings sections from `useSettingsNavItems`. Plus the create actions and a few routes deliberately off the sidebar (Regler, Skattekonto, Återkommande, ROT & RUT, Tillägg, Hjälp, Inställningar).
- **Labels** read from the `nav` / `reports` / `settings_nav` namespaces, so the palette follows the user's locale like the sidebar does. New `command_palette` namespace in both `sv.json` and `en.json`.
- **Ranking**: exact name > prefix > substring > vocabulary (label or alias), sections ordered by their best hit, diacritic-insensitive via `reportMatchesQuery`. "resultat" selects Resultatrapport, "ny kund" selects Ny kund, "stäm av" still lands on Avstämning (DECISIONS 2026-07-30).
- **Deep links**: `/customers?new=1` and `/bookkeeping?new=1` now open the create dialog and strip the flag, like `/invoices`, `/supplier-invoices`, `/salary/employees` already did.

## First principles

Why it drifted: the palette was a second hand-written copy of the navigation. Every nav change since July (v2 tree, Konton, Motparter, Betalfiler, Utlägg, Årsbokslut, the report catalog growing to 22) had to be repeated in `CommandPalette.tsx` and never was: 5 of 22 reports, ~20 pages missing, Löner shown to non-employers, Swedish-only beside a bilingual nav, "Lägg till kund" landing on a closed dialog. The fix removes the copy: the palette reads the nav tree, the catalog and the settings list, gated by one shared function, and a test fails on any palette href without a page file. What was removed: the inline `passesGates` / `ALWAYS_ENABLED` closures in `DashboardNav` (now `nav-gates.ts`), and the hand-listed `ACTION_ENTRIES` / `PAGE_ENTRIES`.

## Verified

- `npm test` (973 passed), `check:types` (533 = baseline), `check:lint` (0 = baseline), `check:guards` green; `i18n/__tests__/message-keys` green with the new namespace.
- In the browser on a local dev server (v2 shell, logged in): sidebar row shows "Sök ⌘K"; clicking it opens the palette; ⌘K toggles it; "resultat" ranks Resultatrapport first with category hints; "ny kund" → `/customers` with the Lägg till kund dialog open and a clean URL; "ny verifikation" → `/bookkeeping` with Ny verifikation open. No console errors from the change (the four dev-overlay issues are PostHog script loads).
- Not clicked through by a human: the mobile sheet row and the v1 (Standard) sidebar row / rail icon; both reuse the same trigger component and existing row markup.

https://claude.ai/code/session_01Q7DLM3CbHws6am3D5j4HFt


jakobwennberg-oss:
Merged origin/main (070eae98a). Beyond the src/ move, the palette now takes its gates from the current DashboardNav: nav-gates.ts uses main's employer rule and adds the quotes and Arkiv gates, and DashboardNav, the palette and the layout all read the same NavGateContext. The v1 shell no longer exists on main, so the shell prop and the collapsed-rail trigger are gone; the start page is labelled Att göra and 'Hem' is kept as an alias. The section-order test now derives its expectation from nav-v2.ts instead of a hand list, and a new test checks that every opt-in gate in the tree also hides the destination in the palette. The customers/bookkeeping ?new=1 handling is resolved against main only (the open declutter PRs #2946-#2950 may conflict on those two pages later). Component, dashboard, reports and i18n tests pass.

PR #2526, opened 2026-09-11
fix(ai): echo Gemini thought signatures through the OpenAI-compatible tool loop

tibbis:
## Summary
- Gemini 3 via Google's OpenAI-compatible endpoint 400s Fråga CFO on the second tool-loop turn: the model requires `tool_calls[].extra_content.google.thought_signature` to be echoed, and `@ai-sdk/openai-compatible` stores it under the custom provider name (`accounted-byo`) while only serializing `providerOptions.google` (vercel/ai#18962).
- A fetch wrapper below the SDK harvests that field from JSON and SSE completions and reattaches it on the next request. Endpoints that never send `extra_content` (Berget, llama.cpp, Qwen) are unchanged.
- Does not rename the provider to `google`, bump the pinned SDK, or inject `skip_thought_signature_validator` unless a signature was already seen.

## Fix from first principles
1. **Why did the problem occur?** The thought signature lives on the wire as `extra_content.google`, but the SDK keys provider metadata by `name: 'accounted-byo'` and only writes `providerOptions.google` back. The bug is a serialization gap between capture and the next request, not a bad model id.
2. **What could be removed or simplified instead?** Renaming the provider to `google` would make the SDK serialize the field, but it would also lie about Berget/llama.cpp. Bumping the SDK is pinned and audited. The wrapper is one code path under the existing fetch hook.
3. **Is this the best solution, or just the proposed one?** A fetch wrapper is better than a provider rename (Berget stays honest) and better than pinning a newer SDK without asking. `skip_thought_signature_validator` is last-resort only when a signature was harvested but this id is unknown.

## Test plan
- [x] `npx vitest run lib/ai/__tests__/gemini-thought-signatures.test.ts lib/ai/__tests__/openai-compatible.test.ts` (26 passed)
- [ ] After deploy: set `AI_MODEL=gemini-3.8-flash` and ask Fråga CFO a question that uses a tool (e.g. resultaträkning)

## Summary by CodeRabbit

- **New Features**
  - Added PWA push-notification settings, including subscriptions, test notifications, and event preferences.
  - Added an optional app-icon badge for pending worklist items.
  - Added company-aware browser tab titles and updated default branding to Accounted.
  - Added scheduled push notifications for supported events.
  - Improved AI assistant responses with the current Swedish date and Accounted identity.

- **Bug Fixes**
  - Improved notification badge updates and push-subscription recovery.
  - Added time limits to scheduled background requests.
  - Improved reliability when processing AI tool interactions.


jakobwennberg-oss:
Thank you for digging into this! Closing because the in-app assistant this fixes is heading for removal (customers will connect their own harness), and the branch also carries unrelated PWA/CI commits and fails the build. If the assistant ends up staying, a clean 3-commit PR against current main (src/ layout) would be welcome.

PR #2527, opened 2026-09-11
fix(vat): read EG account labels as the union, like their EU spelling

### LINKED ISSUE #2525: [VAT] SIE import misclassifies accounts labelled "EG" instead of "EU"
## What happened?

`suggestVatTreatment` in `lib/vat/account-vat-treatment.ts` matches EU-related account labels with `/\beu\b/` and `/utanför eu/`. Real-world charts still carry accounts labelled with the older term **EG**, and the matcher silently misclassifies every one of them — falling through to a different branch and presenting the result with a confident "Trolig" badge.

The chart that surfaced this (SIE4 from Spiris Bokföring, ex-Visma eEkonomi, company created 2021) contains **both spellings side by side**: 3109/3309 say "till annat EU-land" and 4531/4532 say "utanför EU", while 3041–3058 and 4056–4059 say "EG". Likely the older accounts kept names from a chart predating the terminology change and were never renamed, while later additions picked up current BAS names. Whatever the cause, a chart can't be assumed to use one spelling consistently — and the accounts still saying "EG" are exactly the EU-trade accounts the rule exists to catch. 8 reviewable rows in one file were affected.

### Root cause

The class 3 branch tests in order:

- `/export|utanför eu/ && /var/` → `export_goods`
- `/export|utanför eu/ && /tjänst|tjanst/` → `export_services`
- `/\beu\b/ && /var/` → `reverse_charge_eu_goods`
- `/\beu\b/ && /tjänst|tjanst/` → `reverse_charge_eu_services`
- …
- `/momsfri|utan moms/` → `exempt`

"EG" matches none of the first four, so evaluation reaches `momsfri` and returns `exempt`. The ordering is correct — EU/export deliberately precede `momsfri` — the term list is what's incomplete.

The class 4–6 branch has the same three occurrences (`/utanför eu|import/`, `/\beu\b/ && /var/`, `/\beu\b/ && /tjänst/`). There the fallthrough returns `null`, so the row renders as "Använd BAS-standard". Since 4056–4058 aren't in `ACCOUNT_TO_BOX` (`lib/vat/moms-box-mapping.ts` carries 4515–4517 for ruta 20, not these numbers), the static fallback doesn't catch them either — those amounts drop out of the declaration entirely rather than landing in the wrong box.

### Why this reaches past the suggestion

`default_vat_treatment` is stored per account, and `fetchDynamicVatAccounts` (`lib/reports/vat-revenue-accounts.ts`) reads the chart live when a declaration is rendered. A treatment accepted during import therefore misstates **every** period, including the historical ones the import just created, and feeds wrong figures to the periodisk sammanställning. Nothing is frozen per verifikat.

`syncMappedAccounts` also writes confirmed treatments onto accounts that already exist in the chart (`existingVatUpdates` in `lib/import/account-sync.ts`), so a wrong confirmation can overwrite a treatment the user had already set correctly. "Bekräfta alla föreslagna" makes it one click to accept the whole set.

### Proposed fix

Extend the term match to cover both spellings. The expression appears six times, so lifting it out seems worthwhile:

```ts
const UNION = /\b(?:eu|eg)\b/
const OUTSIDE_UNION = /utanför\s+(?:eu|eg)/
```

### Steps to reproduce

1. Export a SIE4 file containing class 3–6 accounts whose `#KONTO` names use "EG" rather than "EU" (example names in the table below).
2. Go to Importera / exportera → SIE-import, upload the file, continue past Förhandsgranskning.
3. At step 3 (Kontomappning), filter on "momskoder att granska".
4. Every EG-named account shows a wrong suggested momskod, badged "Trolig" or defaulted to "Använd BAS-standard".

I have only one such file, so I can't say how common the spelling is across source systems or chart vintages. The failure mode itself is deterministic and readable from the regexes alone.

### Expected behavior

EG-named accounts should get the same treatment as their EU-named equivalents:

| Account | Name in file | Suggested | Ruta | Expected |
|---|---|---|---|---|
| 3048 | Försäljn tjänst EG momsfri | `exempt` | 42 | `reverse_charge_eu_services` (39) |
| 3058 | Försäljn varor EG momsfri | `exempt` | 42 | `reverse_charge_eu_goods` (35) |
| 3045 | Försäljn tjänst utanför EG momsfri | `exempt` | 42 | `export_services` (40) |
| 3055 | Försäljn varor utanför EG momsfri | `exempt` | 42 | `export_goods` (36) |
| 4056 | Inköp varor 25% EG | none | — | `reverse_charge_eu_goods` @ 0.25 (20) |
| 4057 | Inköp varor 12% EG | none | — | `reverse_charge_eu_goods` @ 0.12 (20) |
| 4058 | Inköp varor EG 6% | none | — | `reverse_charge_eu_goods` @ 0.06 (20) |

Two further rows are wrong but ambiguous rather than mechanically fixable: 3046 "Försäljn tjänst till EG 25%" and 3057 "Treparts försäljn varor till EG 25%" both fall through to the `/försälj/` + percent branch and get `standard_25` / ruta 05.

### Deployment

Hosted (app.gnubok.se)

### Browser and OS

Safari 27

### Relevant logs or screenshots

Image

Image

valpet:
# What and why

SIE charts that label accounts **"EG"** (Europeiska gemenskapen, the pre-Lisbon name for the union) had every EU-trade account misclassified by `suggestVatTreatment`:

- **Class 3**: fell through to the `momsfri` branch and was suggested as `exempt` / ruta 42, presented with a confident "Trolig" badge.
- **Class 4-6**: fell through to no suggestion. Since 4056-4058 are not in `ACCOUNT_TO_BOX` (`lib/vat/moms-box-mapping.ts` carries 4515-4517 for ruta 20), the static fallback did not catch them either, so those amounts left the declaration entirely.

This reaches past the suggestion: `default_vat_treatment` is stored per account and `fetchDynamicVatAccounts` reads the chart live, so an accepted treatment misstates **every** period, including historical ones the import just created, and feeds the periodisk sammanställning.

Closes #2525

## Why the problem occurred

Not a missing word. The union term was spelled out as a **literal in six places across two branches**, so the vocabulary had no single definition and a term added to one branch could be forgotten in another. That is the mechanism, and it is why the two branches failed *differently*: class 3 had a `momsfri` catch-all below the EU tests, class 4-6 had nothing.

The fix defines the vocabulary once, and all six sites now share it:

```ts
const UNION = /\b(?:eu|eg)\b/
const OUTSIDE_UNION = /utanför\s+(?:eu|eg)\b/
```

`OUTSIDE_UNION` must be tested before `UNION`, since "utanför EU" also satisfies `UNION`. That ordering was already correct and is preserved.

Six literal regexes collapse into two named constants, so adding the next synonym is a one-line change in one place rather than six correct edits. Nothing else needed to be added: no new treatment, no new mapping table, no config.

The trailing `\b` on `OUTSIDE_UNION` is new and fixes a **second latent bug** in the same expression: the old `/utanför eu/` matched **"utanför Europa"**, which was then zero-rated as an export. Same expression being rewritten, so it is not a drive-by.

## Why this solution rather than the proposed one

The issue proposed the same shape, and I kept it. Two deliberate departures:

**Rejected: matching on account number instead of label.** BAS numbers map to rutor deterministically, which is more reliable than label regex. It does not work here: the whole reason `suggestVatTreatment` exists is non-BAS charts, and the accounts in this report (3041-3058, 4056-4059) are exactly that. Label matching cannot be removed.

**Knowingly not fixed: 3057 "Treparts försäljn varor till EG 25%".** This moves from ruta 05 to ruta 35; the correct box is **38**. There is no `trepartshandel` value in `ACCOUNT_VAT_TREATMENTS`, and the column is guarded by `chart_of_accounts_default_vat_treatment_check`, so a correct suggestion needs a new treatment plus a migration, a pg-real test, and UI/i18n. That is a feature, not a term fix. **Tracked as #2534**, which also records two related gaps found while confirming it: the BAS 2026 dataset carries no trepartshandel accounts at all, and `moms-box-mapping.ts` maps ruta 38 to an account number that is not in that dataset.

Ruta 35 was preferred over the alternatives, and this is the call most worth a second opinion:
- over **null** (which drops the amount from the declaration entirely),
- over the current **ruta 05** (which asserts Swedish output VAT on an intermediary sale).

Class 3 rows always sit behind the import review gate, so the row is shown before anything persists. Happy to swap this for a `trepart` guard returning null if you would rather it not carry a confident suggestion.

**3046 "Försäljn tjänst till EG 25%"** does change, from ruta 05 to ruta 39, which is right for services to an EU business under huvudregeln. The "25%" in the label contradicts that, but the row is reviewed before it persists, so 39 is the better default.

## Verification

| Check | Result |
|---|---|
| The 7 mechanically-fixable rows from the issue | 7/7 resolve to the expected treatment and ruta |
| Behaviour change across the BAS 2026 chart | **0 of 1286 accounts** |
| BAS names matching `eg` that did not already match `eu` | **0** (26 already matched `eu`) |
| `npm test` | 22,867 passed |
| `npm run lint` | 0 errors |
| `npm run check:types` | at baseline (533) |
| `npm run build` | passes |
| `npm run check:guards` | passes |

Ruta mappings were checked against the `swedish-vat` reference (35/36/39/40 for class 3; ruta 20 via 4515-4517 for purchases), not from memory.

## Checklist

- [x] New or changed UI strings exist in both `messages/sv.json` and `messages/en.json` (none changed)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved VAT treatment suggestions for Swedish labels using “EG” (Europeiska gemenskapen), matching “EU” behavior.
  - Correctly distinguishes labels referring to locations outside the EU or EG from unrelated place names.
  - Prevents affected revenue and purchase accounts from receiving incorrect VAT-free treatment.
  - Improved SIE import VAT mapping for EG-labeled accounts, including correct VAT rates and declaration mappings.
  - Keeps ambiguous mappings available for review instead of assigning an automatic treatment.


valpet:
@jakobwennberg-oss could you hit "Approve and run workflows" on this one when you get a moment?

It is a fork PR from a first-time contributor, so every workflow is parked at `action_required` with 0s runtime: Core Build, pg-real tests, CodeQL, Compliance diff, zizmor and PR Agent. None of them have executed, which also means `gh pr checks` here looks short and green while nothing has actually run. The Vercel check fails for the same reason ("Authorization required to deploy").

Verified locally in the meantime, so this is not arriving untested: `npm test` (22,867), the zero-extension `npm run build`, `check:lint`, `check:types`, `check:guards`, plus `skills:check`, `taxonomy:check`, `apiskill:check`, `validate:packs` and `validate:registry`. pg-real was run against a local throwaway database (252 files, 1947 tests) before the fixture commit.

No rush, just flagging that CI cannot speak for itself until someone clicks.

PR #2538, opened 2026-09-12
fix(import): keep a row in the VAT review list while it is being edited

### LINKED ISSUE #2537: [Import] Setting a momskod removes the row from "momskoder att granska" before its rate can be set
### What happened?

In SIE import, step 3 (Kontomappning), the "momskoder att granska" filter is the natural place to work through the accounts that need a VAT code. It is unusable for any row whose rate also needs changing, because the row disappears from the list on the first interaction.

Each class 3-6 row has two controls: a momskod select and a separate sats select. Changing **either** one calls `applyVatTreatmentReview`, which sets `vatTreatmentReviewed: true` (`lib/import/account-vat-treatment.ts:72`). The filter predicate is `m.requiresVatTreatmentReview && !m.vatTreatmentReviewed` (`components/import/AccountMappingStep.tsx:117`), so the row stops matching immediately and vanishes before the second control can be touched.

This matters because picking a treatment sets a default rate that is often wrong for the account. `defaultRateForVatTreatment` returns 0.25 for any reverse-charge treatment on class 4-6, so an account like `4057 "Inköp varor 12% EG"` becomes `reverse_charge_eu_goods @ 25%` the moment you choose the momskod, and the row is gone before you can correct the 25% to 12%.

#### Where the row goes

Nowhere reachable. It does not appear under "manuellt satta" either: that filter is `m.isOverride` (`AccountMappingStep.tsx:120`), and `isOverride` means the user remapped the account to a different **target account** (`lib/import/account-mapper.ts`, and `types.ts:209` "User manually set this"). It is not set by a VAT edit, and it should not be: the two are different concepts. There is simply no filter for "momskod I set by hand".

So after the first click the row is only visible under "Visa alla", where it sits among every account in the file (243 in the chart that surfaced this), across paginated pages.

#### Why this is a migration blocker in practice

Working around #2525 before that fix lands requires setting momskod **and** sats by hand on eleven EG-labelled accounts, three of which need a non-default rate. The review filter is exactly the tool for that job, and this bug removes it. The workaround for the workaround is to switch to "Visa alla" before touching anything and use the search box for each account number, one at a time.

### Steps to reproduce

1. Start a SIE import with any chart containing a class 4 account whose rate is not 25 percent. The fixture at `lib/import/__tests__/fixtures/eg-eu-bas97.se` (added in #2527) has `4057 "Inköp varor 12% EG"`, or use any chart of your own.
2. Continue past Förhandsgranskning to step 3, Kontomappning.
3. The "momskoder att granska" filter is selected by default when there are rows to review.
4. On row 4057, pick a momskod in the first select, for example "EU-varor".
5. The row disappears from the list before the sats select can be touched. It is now `reverse_charge_eu_goods @ 25%`, which is wrong for a 12 percent account.
6. Click "manuellt satta": the row is not there either. Only "Visa alla" shows it.

### Expected behavior

A row being edited should stay visible until the user navigates away from the step or changes filter. Setting a momskod is the start of reviewing a row, not the end of it.

Suggested fix, smallest version: keep a local set of source accounts edited during this step in `AccountMappingStep`, and let the `vat_review` filter match `requiresVatTreatmentReview && (!vatTreatmentReviewed || recentlyEdited.has(sourceAccount))`. The chip counter keeps using the live predicate, so it still counts down as rows are completed, and "Bekräfta alla föreslagna" still clears the list in one go because it does not go through the per-row handler.

Alternative considered: treat the filter as a worklist snapshotted when it is selected, so the whole set stays stable until you leave the step. That is arguably better for working through a long list, but it changes behaviour for people who rely on rows disappearing as they finish them, so the narrower fix seems right unless a maintainer prefers the worklist.

Not suggested: setting `isOverride` on a VAT edit. It would put the row under "manuellt satta", but `isOverride` means the account was remapped to a different target, and overloading it would misreport the mapping and risk persisting a mapping override that the user never made.

### Deployment

Hosted (app.gnubok.se)

### Browser and OS

Reported from Safari 27 on macOS; mechanism confirmed by reading the filter predicates, applies to any browser.

valpet:
# What and why

On the SIE import mapping step, the "momskoder att granska" filter was unusable for any row whose rate also needed changing: the row left the list on the first interaction.

Each class 3-6 row carries a momskod select **and** a separate sats select. Changing either calls `applyVatTreatmentReview`, which sets `vatTreatmentReviewed: true`, and the list filtered on `requiresVatTreatmentReview && !vatTreatmentReviewed`. So the row vanished before the second control could be touched.

That is worse than an annoyance, because picking a treatment also assigns a default rate: `defaultRateForVatTreatment` returns 0.25 for any reverse charge on class 4-6. `4057 "Inköp varor 12% EG"` became 25% and left the list before the 25 could be corrected to 12. The row did not appear under "manuellt satta" either, since that filter is `isOverride`, which means the account was remapped to a different target and is a different concept. Only "Visa alla" showed it, among every account in the file.

### Why this can block a correct import

The wrong rate does not stay in the UI. `enrichAccountMappingsWithVat` feeds `account-sync.ts`, which writes both `default_vat_treatment` and `default_vat_rate` onto `chart_of_accounts` at import (lines 87-88, 110-111, 275-276). `lib/reports/vat-revenue-accounts.ts` then reads those columns **live** from the chart when building the momsdeklaration (line 124) and derives the rate from the treatment, with no period scoping. A rate accepted at import therefore misstates **every** period, including the historical ones the same import just created.

The step's own gate does not catch it either. `confirmVatReview` blocks Continue on `mappings.some(needsVatTreatmentReview)`, and the row counts as reviewed the moment the first select is touched. So the sequence is: set a momskod deliberately, the row disappears carrying a 25% you never chose, the gate stops objecting, and the import completes looking reviewed.

The only recovery inside the step is to switch to "Visa alla" before touching anything and search each account by number, which nobody discovers unprompted. After the import it takes `PUT /api/bookkeeping/accounts/{number}` or the `gnubok_update_account` MCP tool, per account.

This compounds with #2525: EG-labelled charts are exactly the ones needing manual VAT edits on many rows at once, so the two bugs meet on the same import.

Closes #2537

## Why the problem occurred

The predicate was written inline at the point of use, **six times across four files**, so nothing kept the places that ask "does this row still need review?" in agreement. The list and the chip counter had already drifted apart; the confirm button was a third copy waiting to do the same.

This is not a hypothesis. I first consolidated two of the six and shipped a fix that visibly misbehaved for exactly that reason: the row stayed in the list while its confirm control, still keyed to its own copy of the predicate, disappeared. The row read as unfinished with no way to finish it. The bug class reproduced itself inside the fix for the bug class.

Six inline copies collapse to two named predicates in `lib/import/account-vat-treatment.ts`:

- `needsVatTreatmentReview`: the narrow "still outstanding" question. Drives the chip counter, the default-filter check, the show-the-step and block-continue gates in `import/page.tsx`, the auto-skip in `ArcimMigrationWorkspace`, and `applyVatTreatmentReviewAll`.
- `isInVatReviewList`: visibility, deliberately wider. Used by the list and by the confirm button, so the control and the row it belongs to cannot drift apart again.

Moving them into `lib/` also puts them inside the repo's test scope. The component could not be unit-tested at all under the existing convention.

No new state was introduced beyond one component-local `Set` of rows touched through the two selects during the current visit to the filter.

## Why this solution rather than the alternatives

**Rejected: widen the filter to `requiresVatTreatmentReview` alone.** The list would be stable, but the chip counter would stop counting down and the user loses any sense of progress through a 57-row worklist.

**Rejected: snapshot a worklist when the filter is selected.** Arguably nicer for working through a long list, but it changes behaviour for people who rely on rows disappearing as they finish them. The narrower fix earns its place; this is the one to revisit if the worklist feel is wanted.

**Rejected outright: set `isOverride` on a VAT edit** so the row lands under "manuellt satta". `isOverride` means the account was remapped to a different target account. Overloading it would misreport the mapping and risk persisting a mapping override the user never made.

**Confirm semantics.** The per-row confirm and "Bekräfta alla föreslagna" deliberately do **not** feed the sticky set, and confirm releases a row from it. Those actions mean "done with this row", and the row leaving is the point. Without the release the button is inert on an edited row, since the set holds the row visible regardless of what the button does. That was a real bug in the first iteration, caught by clicking it rather than by any test.

## Verification

UI, on the running app:

| Step | Row 4057 | Confirm button |
|---|---|---|
| Before | visible | present |
| After setting momskod | **visible** | **present** |
| After clicking the check | **gone** | gone |

The chip counted down correctly throughout (4 to 3 to 2), and the wrong 25% default was reachable and correctable while the row was held.

| Check | Result |
|---|---|
| `npm test` | 22,895 passed |
| `npm run build` (zero extensions, as CI runs it) | passes |
| `npm run check:lint` / `check:types` / `check:guards` | pass |
| `skills:check`, `taxonomy:check`, `apiskill:check`, `validate:packs`, `validate:registry` | pass |
| No core imports from `@/extensions/` | pass |

Honest limit: the sticky-set release lives in component state, which is outside the repo's test scope (`lib/` and `app/api/` only, no component tests). The two predicates are unit-tested, including the regression that the row survives between setting momskod and sats; the confirm-releases behaviour is verified in the UI only.

## Not in this PR

`ConfidenceBadge` still renders in the VAT review filter. It describes the **account mapping**, not the momskod, so a "Trolig" badge next to a momskod the user just chose reads as an opinion about the momskod. Filed separately as #2539, which digs one level down: the badge grades a certainty that was never measured. `suggestMappings` emits only three confidence values, and the 0.7 that renders as "Trolig" marks an identity mapping that will be created, not a doubtful match. Fixing that dissolves the concern here without moving the column.

The hardcoded Swedish badge strings are covered there too: the import mapping UI is not on the "stays Swedish" list in `.claude/rules/i18n.md`, so that is a plain gap rather than a deliberate exemption.

## Checklist

- [x] New or changed UI strings exist in both `messages/sv.json` and `messages/en.json` (none added)
- [x] Core builds with zero extensions enabled (verified, not assumed)

## Summary by CodeRabbit

- **Bug Fixes**
  - VAT treatment review rows now remain visible while users edit account treatment and rate values.
  - Review progress and confirmation controls consistently reflect outstanding VAT treatment reviews across import workflows.
  - Individual and bulk confirmation actions now correctly remove completed items from the review list.
  - Filtering and review controls behave more consistently during VAT treatment mapping.

- **Tests**
  - Added coverage for VAT review visibility during editing, filtering, outstanding reviews, and individual or bulk confirmation.


valpet:
@jakobwennberg-oss same request as on #2527: this fork PR's workflows are parked at `action_required` and have never run, so "Approve and run workflows" is needed before CI can say anything.

Verified locally: `npm test` (22,895), the zero-extension `npm run build` as CI runs it, `check:lint`, `check:types`, `check:guards`, `skills:check`, `taxonomy:check`, `apiskill:check`, `validate:packs`, `validate:registry`, and no core imports from `@/extensions/`. The behaviour change itself was verified by driving the running app, since the sticky-set part lives in component state and is outside the repo's test scope.

jakobwennberg-oss:
Merged current main into this branch (5aa5e2c2f), no conflicts after the src/ move. Lint clean on touched files, account-vat-treatment tests pass.

PR #2541, opened 2026-09-12
fix(import): replace the confidence badge with what the mapper actually did

### LINKED ISSUE #2539: [Import] Konfidens column grades a certainty that was never measured: "Trolig" on identity mappings
### What happened?

The Konfidens column on the SIE import mapping step reports a confidence that does not exist, and the value it shows most often is wrong.

`suggestMappings` produces exactly **three** confidence values across the whole codebase: `1.0`, `0.7` and `0` (`lib/import/account-mapper.ts:97`, `:109`, `:132`, `:182`, `:274`). `ConfidenceBadge` then thresholds that with `>= 0.9` for "Exakt", `>= 0.7` for "Trolig" and "Osäker" below, so a three-valued enum is being dressed as a float and re-derived by inequality. `matchType` already carries the same distinction as a proper string (`'exact'`, `'bas_range'`), is passed into the badge, and is explicitly discarded:

```ts
matchType: string  // Keep for potential future use
```

#### "Trolig" does not mean what it says

`0.7` is assigned in one branch only, and the comment above it says what it means:

> Fallback: if the account is a valid BAS-range number (1000-8999), self-map it using the name from the SIE file.

That is an **identity mapping**: 4057 to 4057, created at import. There is nothing uncertain about it. The number is identical to itself, and the only reason it is not `1.0` is that the number is absent from the BAS 2026 reference. The badge turns "this account will be created" into "we are probably right about this mapping", which is an invented doubt.

It is also the common case, not an edge case. In the fixture committed in #2527 (`lib/import/__tests__/fixtures/eg-eu-bas97.se`), **11 of 12 accounts** resolve to `bas_range @0.7` and therefore read "Trolig"; only 1930 is in BAS 2026. Any chart from a source system that keeps sub-accounts outside BAS 2026, which is most real charts, lights up almost entirely in an advisory colour that means nothing.

#### Two related problems in the same column

- **Threshold drift.** The UI counts "osäkra" as `confidence < 0.7` (`AccountMappingStep.tsx:119`, `:201`) while the importer's own statistics use `< 0.5` (`account-mapper.ts:204`, `:238`). Today both answer the same, purely because no value exists between 0 and 0.7. Introduce a single 0.6 and the chip and the stats disagree. Same duplicated-predicate shape as #2537.
- **Hardcoded Swedish.** `"Manuell"`, `"Exakt"`, `"Trolig"` and `"Osäker"` are literals in `ConfidenceBadge`, while the rest of the component goes through `useTranslations`. The import mapping UI is not on the "stays Swedish" list in `.claude/rules/i18n.md` (which covers invoice PDFs, the year-end wizard, the journal entry editor, INK2/NE/SRU, SIE export, BAS names, momsrutor, AGI/KU and engine domain errors), so this looks like a plain gap rather than a deliberate exemption.

### Steps to reproduce

1. Start a SIE import with any chart containing accounts outside the BAS 2026 reference. The fixture at `lib/import/__tests__/fixtures/eg-eu-bas97.se` (#2527) works, or any real export from Fortnox, Visma or similar.
2. Continue to step 3, Kontomappning, and choose "Visa alla".
3. Read the KONFIDENS column. Every account that will be created reads "Trolig", including rows whose source and target account numbers are identical.

Or without the UI:

```ts
const m = suggestMappings(parsed.accounts, BAS_REFERENCE)
// fixture: { 'bas_range @0.7': 11, 'exact @1': 1 }
```

### Expected behavior

Render what the mapper actually knows, which is `matchType`, and drop the numeric threshold from the UI entirely.

| Case | Today | Proposed |
|---|---|---|
| Present in BAS 2026 | "Exakt" | nothing, or a quiet "BAS" marker |
| Valid BAS-range number, will be created | **"Trolig"** | **"Skapas"** or "Nytt konto" |
| No target | "Osäker" | "Ej mappad" (already has its own filter and chip) |
| Remapped by the user | "Manuell" | "Manuell" |

The column then states what the import is going to do, which is the useful thing when reviewing a migration, instead of grading a certainty that was never measured.

Keep the `confidence` field itself: it is persisted in `sie_account_mappings` and feeds `averageConfidence` in the import statistics. Only the presentation is wrong.

While in the file, the two thresholds should agree (or disappear along with the numeric comparison), and the four badge strings should move to `messages/sv.json` and `messages/en.json`.

#### Alternatives considered

- **Rename only the 0.7 band.** Smallest possible change and it removes the false claim. It leaves a fake numeric scale behind, so the drift and the discarded `matchType` stay.
- **Hide the column when the filter is `vat_review`.** This was the original suggestion, because a "Trolig" badge next to a momskod the user just chose reads as an opinion about the momskod. It treats the symptom: with the table above, "Skapas" next to a momskod is not confusing, so the problem dissolves without moving the column.
- **Do nothing.** Defensible if the column is considered low-stakes chrome, but it is the only signal in that row about what the import will do to the account, and it currently says the wrong thing on most rows.

This is copy and design-system territory under `.claude/rules/design.md`, so the wording above is a proposal rather than a decision.

### Deployment

Hosted (app.gnubok.se)

### Browser and OS

Any; the behaviour is in the component, not browser-specific.

valpet:
# What and why

The Konfidens column on the SIE import mapping step graded a certainty that was never measured, and the value it showed most often was false.

`suggestMappings` emits exactly three confidence values: `1.0` (exact BAS 2026 match), `0.7` (valid BAS-range number self-mapped) and `0` (no match). `ConfidenceBadge` then thresholded those at `>= 0.9` and `>= 0.7`, re-deriving by inequality a three-valued enum that `matchType` already carried as a string. `matchType` was passed into the badge purely to be discarded, with a `// Keep for potential future use` comment.

`0.7` rendered as **"Trolig"**, and it is assigned in exactly one branch: a valid BAS-range number, self-mapped to itself, created at import. That is an identity mapping, 4057 to 4057, with nothing uncertain about it. In the fixture committed in #2527 it is **11 of 12 accounts**; any chart keeping sub-accounts outside BAS 2026, which is most real charts, lit up almost entirely in an advisory colour that meant nothing.

"Osäker" never rendered at all: the only zero-confidence case carries an empty `targetAccount`, and the badge was guarded on the target being present.

Closes #2539

## Why the problem occurred

A derived label was computed from a proxy instead of from the fact. The mapper knows precisely what it did and records it in `matchType`; the UI ignored that and inferred a grade from a number that was only ever a stand-in for the same three cases. Once a number stands in for an enum, thresholds accumulate around it and drift: this column had **three** separate `< 0.7` / `< 0.5` comparisons across the component, the page and the mapper's own statistics.

- The numeric threshold is gone from the mapping table; the badge reads `matchType`, `isOverride` and whether the mapping is an identity.
- The **"osäkra" filter chip and its `low_confidence` FilterType are deleted** (second commit, droppable on its own). Its predicate was `targetAccount && confidence < 0.7`, and the only zero-confidence rows are precisely the ones with an empty `targetAccount`. The two halves were mutually exclusive, so the count was structurally `0` and the filter could never select a row. Every import has been showing "0 osäkra" next to a control that did nothing.
- The intro prose said "granska de osäkra nedan" and lost its referent; it now says "granska de markerade nedan", which matches how rows are actually marked (the warning border on the momskod select).

`confidence` itself stays on the type: it is persisted in `sie_account_mappings` and feeds `averageConfidence`. Only the presentation was wrong.

## Why this solution rather than the alternatives

**The first cut showed a badge only for the two non-identity cases and left every other row blank.** That was wrong for a different reason: in a migration an empty cell reads as "nothing is known about this row" rather than "this one is fine", which is the opposite of what you want when checking that a whole chart landed correctly. All four states are now named.

| State | Label | What it claims |
|---|---|---|
| In the BAS 2026 reference | **BAS-konto** | The number is recognised; name and type come from the standard |
| Valid number, not in the reference | **Från filen** | The number is kept; name and type come from your upload |
| Group-header redirect | **Omdirigerad** | Postings land on a different number |
| User picked the target | **Manuell** | Your choice, not ours |

**"Från filen" rather than "eget konto" is the load-bearing choice.** `bas_range` covers two different things: a source system's invention (a Fortnox 4599) *and* a perfectly standard sub-account our reference does not carry. `1241 Personbilar` is verifiably not among its 1286 entries. Calling either one "the company's own account" asserts something the mapper cannot know, which is the same mistake "Trolig" made with a friendlier word. "Från filen" says only what is true.

The Målkonto column corroborates it: a BAS-konto row shows the BAS 2026 name (`1930` becomes "Företagskonto / checkkonto" even though the file said "Företagskonto"), while a Från filen row keeps the name the file supplied.

**Rejected: rename only the 0.7 band.** Smallest possible change and it removes the false claim, but it keeps a fake numeric scale and a discarded `matchType`.

**Rejected: hide the column inside the `vat_review` filter.** This was the original suggestion, since a "Trolig" badge next to a momskod the user just chose reads as an opinion about the momskod. It treats the symptom; with the labels above, "Från filen" next to a momskod is not confusing.

**Rejected: "BAS 2026" as the label.** Precise, but wrong the day the reference moves to BAS 2027.

**Deliberately left alone:** the wider `lowConfidence` surface in `SIEPreviewStep`, `import/page.tsx` (which carries a third copy of the 0.7 threshold), `account-mapper`'s own statistics (which use 0.5) and `types.ts`. That is a larger cleanup across the preview step's warnings and the same duplicated-threshold shape as #2537, so it wants its own pass.

## Verification

Driven in the running app with a file covering all four states:

| Account | Målkonto | Matchning |
|---|---|---|
| 1241 Personbilar (real BAS, absent from the reference) | 1241 (skapas) | Från filen |
| 4599 Eget inköpskonto (genuine invention) | 4599 (skapas) | Från filen |
| 3058 Försäljn varor EG momsfri | 3058 (skapas) | Från filen |
| 1930 | 1930 Företagskonto / checkkonto | BAS-konto |
| 2640 | **2641** Debiterad ingående moms | Omdirigerad |

The chip row after the removal reads "momskoder att granska / ej mappade / nya konton skapas / manuellt satta / Visa alla", with no "0 osäkra".

| Check | Result |
|---|---|
| `npm test` | 22,890 passed |
| `npm run build` (zero extensions, as CI runs it) | passes |
| `check:lint` / `check:types` / `check:guards` | pass |
| `skills:check`, `taxonomy:check`, `apiskill:check`, `validate:packs`, `validate:registry` | pass |
| sv/en key parity | no keys missing on either side |

Honest limit: this is a component change, and the repo's test scope is `lib/` and `app/api/` with no component tests, so the badge logic is verified by driving the app rather than by a unit test.

## Note on copy

The four labels are my wording, not a maintainer's. They are one i18n entry each (`match_bas`, `match_from_file`, `match_redirected`, `match_manual`, plus `match_column` for the header) and trivial to change. The reasoning behind "Från filen" is the part worth keeping even if the words change.

## Checklist

- [x] New or changed logic in `lib/` or `app/api/` has tests (no `lib/` logic changed; component only)
- [x] Core builds with zero extensions enabled (verified, not assumed)

## Summary by CodeRabbit

- **New Features**
  - Added match-type labels to the account-mapping table for BAS accounts, file matches, moved accounts, and manual mappings.
  - Added English and Swedish translations for the new column and labels.

- **Bug Fixes**
  - Match indicators now reflect the actual mapping type instead of confidence scores.
  - Removed the non-functional “uncertain” filter and updated the surrounding description.
  - Updated the redirected-match label to “Moved” in English and “Flyttad” in Swedish.


valpet:
@jakobwennberg-oss third fork PR in this batch (with #2527 and #2538), so it needs "Approve and run workflows" the same way. Its workflows are parked at `action_required` and have never run.

Verified locally: `npm test` (22,890), the zero-extension `npm run build`, all three ratchets, the five generator/validator gates, and sv/en key parity. The behaviour itself was checked by driving the running app, since this is a component change and the repo's test scope is `lib/` and `app/api/`.

jakobwennberg-oss:
Merged current main into this branch (3faa08264), no conflicts after the src/ move. Lint clean on touched files, account-mapper tests pass.

PR #2603, opened 2026-09-14
docs(design): read the v2 shell from origin/main before any UI work

jakobwennberg-oss:
## What

Adds a first section to `.claude/rules/design.md` stating that the v2 shell (SidebarV2 + nav-v2 + the 48px top bar, full-bleed panel) is the only shell, that v1 was removed in #2578, and that any UI work or mockup must fetch and read the shell from `origin/main` when the checkout is behind. CLAUDE.md gets a one-line pointer under "When Uncertain". DECISIONS.md records why.

## First principles

The failure was not a missing rule about the shell (design.md on main already describes v2) but a session matching a stale checkout: the feature branch it ran on was based on 2026-08-21, so `DashboardNav.tsx`, the old MainContainer and the old design.md were all still present locally and looked authoritative. The guard therefore targets the checkout, not the shell description: fetch, compare with `git merge-base --is-ancestor`, and read from `origin/main` when behind.

Docs only, no code.

https://claude.ai/code/session_01HuRqdB5zXmtLcSdcMpqPms


jakobwennberg-oss:
Closing: design.md on main already describes the v2 shell (SidebarV2, 220px nav), and the claim that DashboardNav.tsx was removed is wrong (it still exists as the wrapper that renders SidebarV2). A guard written inside the checkout also can't reach the stale checkout it targets; the rule lives in the agents' memory instead.

PR #2640, opened 2026-09-15
fix(ui): make a labelled InfoTooltip reachable by keyboard

yummymaya:
`InfoTooltip` has two shapes. Without children it renders a `<button>` that sits in the tab order. With children, which is how a piece of text gets an explanation attached, it renders a bare `<span>`.

Radix opens a tooltip on focus as well as hover, but a span cannot take focus, so that half never worked. Nine of sixteen call sites use that branch, six of them the skattetabell and jämkning explanations in `EmployeeTaxCard`. All of it was mouse-only.

One attribute, plus a focus ring.

<details>
<summary><b>Why a span with tabIndex and not a button</b></summary>

`TooltipTrigger` with `asChild` forwards props but never adds `tabIndex`; verified in the installed package rather than the docs. Without it the span is not focusable and Radix's focus handler cannot fire. On touch the content needs a long-press users have no reason to guess at.

Six of the nine call sites sit inside a `<label htmlFor>`, where a `<button>` swallows the click that focuses the field. That rules out the obvious fix.

No `role="button"` either: there is nothing to activate. Focus is the whole interaction, and Radix already points `aria-describedby` at the content while it is open, so a screen reader reads it on arrival.

The ring uses the repo's `focus-visible` tokens and `rounded-sm`, the leaf tier of the radius ladder. The cost is one tab stop per tooltip, which is the point.

WCAG 2.1 asks for keyboard access to content shown on hover; `.claude/rules/design.md` asks for keyboard navigability with visible focus rings.
</details>

<details>
<summary><b>Verification</b></summary>

23 856 tests passing, types at baseline 533, eslint clean on the file, and the warning count is identical with and without the change.

Not seen rendering: none of the nine call sites draw in the data at hand, so this rests on the source change and the gates rather than on a screenshot. The same defect class on a file picker was reproduced and fixed in the browser while reviewing the source-chart branch.
</details>

## Summary by CodeRabbit

* **Accessibility**
  * Info tooltips can now be reached using keyboard navigation.
  * Added visible focus styling and a help cursor to make tooltip interactions clearer.
  * Improved tooltip usability for keyboard and pointer interactions while preserving the existing tooltip behavior.


yummymaya:
Flagging that the next step here is a maintainer's.

Every GitHub Actions workflow is at `action_required` on the current head (1b377b899): Core Build, Checks, CodeQL, pg-real tests, Compliance diff, zizmor. None has run, so the green marks above are only the third-party bots.

This is the smallest of the four open ones, +22/-1 across 2 files, an accessibility fix that makes a labelled `InfoTooltip` reachable by keyboard. There are no unresolved review threads. It needs one approval on the workflow runs to become mergeable or not, and I have `pull` only, so I cannot give it or request a reviewer.

jakobwennberg-oss:
Merged origin/main into this branch. The component now lives at src/components/ui/info-tooltip.tsx and git carried the change over. DECISIONS.md merged without a conflict, and both sides' lines are kept. Lint and typecheck are clean for the file.

PR #2644, opened 2026-09-15
fix(onboarding): a journey step never clips its own primary action

### LINKED ISSUE #2642: Onboarding: the primary action is clipped when a step outgrows the step box
## Original report

"In the onboarding flow, the button is cut off." Screenshot: the Klart step (Valkommen, <name>. + company profile card with 7 rows) on a 1327x768 window; the "Fortsatt" button is sliced in half by the bottom of the step area, and nothing on the page scrolls to reach it.

## Source

Founder, screenshot, 2026-09-15.

## Mechanism

`.jny-qstep` (components/onboarding/journey/journey.css) is a clipped scroll box (`max-height: 100%; overflow-y: auto`) inside `.jny`, which is `height: 100dvh; overflow: hidden` so the page itself never scrolls. On the done step the title plus the profile card already fill that box, and `Reveal` (components/onboarding/journey/OnboardingJourney.tsx:1407) mounts the action row inside a *second, nested* `.jny-qstep`, so the button is laid out past the clip edge of the outer one with no scrollbar reachable from where the pointer is. Any step whose answer surface outgrows the box can clip its own primary action the same way.

mattssonn:
# What and why

The Klart step of onboarding showed the profile card with the Fortsätt button sliced in half, and on a shorter window the button was gone entirely, with empty space still on screen below it. Closes #2642.

**Why it happened.** `.jny-qarea` took `flex: 1`, i.e. flex-basis 0, so the column inside `.jny-center` never went into deficit and `.jny-balance` (the optical-centering spacer, 180px) kept every pixel while the question area was handed only what was left over. A step taller than that got clipped by `.jny-qstep`, which is a scroll box inside a page that deliberately does not scroll, so the action row sat below the clip edge with only a detached scrollbar to reach it. The done scene made it worse: the delayed continue button was wrapped in a second `.jny-qstep`, a scroll box inside a scroll box.

**What was removed or simplified instead.** No new height budget, no per-step measurement, no scroll-into-view timer. The spacer's own comment already promised it "collapses first when a tall step needs the room"; giving the question area a content-based basis is what makes that true. The nested scroll box is deleted rather than compensated for, and the address step's wrapper div became a fragment instead of gaining a rule of its own.

**Why this rather than the shape the report suggested.** The obvious patch is to shrink the done card until it fits a 768px window. That fixes one card size and leaves every other tall step (the BankID roster, the source grid, a first-fiscal-year step with notes) one row away from the same bug. This fixes the class: the spacer yields space before a step has to clip, and if even that is not enough, the action row stays seated at the bottom of the step while the answer surface scrolls under it.

**Migrations:** none. **Strings:** none added. **Tests:** `components/onboarding/journey/__tests__/journey-layout.test.ts` guards the four invariants the layout now rests on; it fails if the sticky rule loses its `.jny-qarea` scope (which would leak into the books act, where the page itself scrolls), if it loses the direct-child combinator, if the question area goes back to flex-basis 0, or if the address step re-wraps its action row.

**Verified** by rendering the real stylesheet in headless Chrome at 1327x900, 768, 620 and 500: before, the button is half-cut at 768 and entirely below the clip edge at 620; after, it is fully visible down to 620 and pinned in view at 500. Short steps are pixel-identical above 640px tall.

## Checklist

- [x] `npm run check:lint`, `npm run check:types`, `npm run check:guards` and the touched tests pass locally
- [x] New or changed logic in `lib/` or `app/api/`: none (CSS and one component wrapper)
- [x] New or changed UI strings: none
- [x] Migrations: none
- [x] Core builds with zero extensions enabled

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved onboarding layouts so longer steps are less likely to be clipped on smaller screens.
  - Kept action controls visible while scrolling through onboarding steps.
  - Improved spacing and visibility around action controls on short screens.
  - Corrected address-step layout behavior to support sticky actions.

- **Style**
  - Added a smoother reveal transition for onboarding content.


mattssonn:
CodeRabbit's one finding is fixed in 2df821be9: the address-step guard now counts how many JSX elements enclose the action row and requires zero, so an attributed wrapper fails it too (verified 0 as written, 1 with `<div className="wrapper">` spliced in). No other findings: compliance swarm and the Swedish accounting review both reported clean.

PR #2648, opened 2026-09-15
fix(onboarding): the bank picker lists accounts another company books instead of hiding them

### LINKED ISSUE #2647: Onboarding bank step shows an empty account list when every account is already booked by another company
## Original report

A user got stuck in the onboarding wizard at the bank step. The page said "Swedbank är ansluten. Vilka konton ska in i bokföringen? Inget hämtas förrän du valt." and then showed nothing at all: no accounts, no message, no primary button, only "Hoppa över". The same wizard worked fine the day before.

## Confirmed

Reproduced from prod data, 2026-09-15.

- Connection `cefe2715-dd56-4cb5-8ca4-d7d0798c2ebc`, company Diesel & Chill AB, Swedbank, created 10:29 UTC. 52 verifikat in the books at that moment, which matches the wizard's step label in the screenshot.
- The PSU authenticated with their own BankID, so the consent returned two accounts that both belong to sibling companies of the same user. Both IBANs already had bank-linked `cash_accounts` rows in those companies, so the callback's cross-company guard stamped `claimed_by_company_id` on both.
- `components/onboarding/books/steps/BankStep.tsx:247` does `.filter((a) => !a.claimed_by_company_id)` on `accounts_data`, so every account was removed and the list rendered empty. `accounts_data` was a non-empty array, so the `bank_no_accounts` message never fired either.
- The user escaped to /settings/banking, whose picker lists claimed accounts with "Synkas redan i X" and lets them be ticked, and saved there at 10:31:34 (`initial_lookback_days: 90`, the settings dialog default, not the wizard's after-last-entry date).

## Root cause

The callback guard already records the right thing: the account is stored `enabled: false`, flagged with the claiming company, and kept out of the cash-account mirror. The wizard's picker threw that information away by filtering the rows out instead of rendering them, so the one screen a new user cannot leave went blank with no explanation and no way forward.

## Decision

Warn, never block. Founder direction 2026-09-15: the app cannot know which of two companies is the right home for an account, so both pickers show every account the consent returned, claimed ones unticked with a quiet notice naming the claiming company, and the user decides.

Explicitly not in scope: no database constraint on IBAN, no takeover-as-move, no cleanup of the 40 IBANs that currently sit in more than one company, no change to the callback guard.

## Source

Founder report with screenshot, 2026-09-15. Investigated in prod (Supabase read-only) and Vercel runtime logs for deployment `dpl_6d9u6DCzwGHgMqcZk5FgCAmHv7Kn`.

mattssonn:
# What and why

A user connected Swedbank in the onboarding wizard and landed on a blank step: "Swedbank är ansluten", and under it nothing at all. No accounts, no message, no primary action, only "Hoppa över". Prod, 2026-09-15, connection `cefe2715` on Diesel & Chill AB, 52 verifikat in the books, which is what the step label in his screenshot said.

**Why it happened.** The Enable Banking callback's cross-company guard stores an account that another of the user's companies already books as `enabled: false` with `claimed_by_company_id`, and keeps it out of the cash-account mirror. That is the whole policy, and it is correct. `BankStep` then filtered those rows out of the picker, so the guard's information was thrown away instead of shown; a consent where every account is claimed rendered an empty list. `accounts_data` was a non-empty array, so the existing `bank_no_accounts` message never fired either. Both of his Swedbank accounts are booked by his Arcim companies, so all of them vanished.

**What was removed rather than added.** The filter. Nothing new decides anything: the picker renders what the guard already recorded. The mapping moved into `lib/onboarding-books/picker-accounts.ts` as a pure function, so "never dropped, never pre-ticked, free accounts first" has tests, which a component in this repo cannot have.

**Why this and not the fix first proposed.** The first instinct was to refuse the connect when the bank is already connected to another company. The guard is per account, not per bank, and which accounts a consent covers is only known after the BankID round trip, so a pick-time block would refuse legitimate second-company connects at the same bank (an EF and an AB at Swedbank, say). Making the block real in the database, a unique IBAN across companies, was rejected too: 40 IBANs already sit in more than one company in prod, so it would need a data cleanup first. Founder direction 2026-09-15 is warn, never block: one bank account belongs to one legal entity, but the app cannot know which of two companies is its right home, so it names the claimant and leaves the choice.

Changes:

- claimed accounts render with "Bokförs redan i X" under the name, unticked, sorted behind the free ones
- ticking one shows a single quiet line about the consequence: a notice, not a dialog, nothing to dismiss
- the settings picker, which labelled claimed accounts but let a tick through silently, gets the same line
- an empty `accounts_data` array now ends like a null one, back to the bank list with "Banken lämnade inga konton", instead of a blank page

Second commit fixes three findings from `/skeptic` on the first, all in its own new code: the claim line kept its muted colour on the inverted ticked pill (~3:1 contrast exactly when selected), both label lines truncated with an ellipsis so a long claimant was cut, and the claim predicate keyed on the flag alone while `partitionByClaim` requires flag AND `enabled === false` and documents that conjunction as load-bearing.

Migrations: none.

Tests: `lib/onboarding-books/__tests__/picker-accounts.test.ts` (5 cases). `npx vitest run lib/onboarding-books extensions/general/enable-banking app/api/extensions/enable-banking` green, `check:lint`, `check:types`, `check:guards` at baseline.

Closes #2647

## Checklist

- [x] `npm run check:lint`, `npm run check:types`, `npm run check:guards` and the affected vitest suites pass locally; the full build is left to CI's "Build (zero extensions)"
- [ ] Commits are signed off (`git commit -s`, see [DCO](../DCO)): they carry Co-Authored-By, not Signed-off-by, and no DCO check runs on this repo

## Summary by CodeRabbit

- **New Features**
  - Bank account selection now shows accounts already booked by another company, including the claimant’s name.
  - Selecting such an account displays a warning that it will appear in both companies’ books.
  - Account details now include clearer labels, account numbers, currencies, ledgers, and balances.

- **Bug Fixes**
  - Accounts are no longer hidden solely because another company has booked them.
  - Empty account results return users to the bank selection step with an appropriate message.

- **Style**
  - Selected year and account indicators now use a muted color, with improved wrapping for account labels.


mattssonn:
Resolution pass, three findings, one fix.

**Swedish Accounting Compliance Review, double-booking under BFL 5 kap: answered, decision recorded.** Correct that one bank account belongs to one bokföringsskyldig entity. The disagreement is about where the rule is enforced. Nothing here books anything: enabling an account imports unbooked `transactions` rows, auto-booking is off outside dev/test (`lib/transactions/ingest.ts:1161`), and a verifikat only exists once the user books the row. The app cannot tell which of the user's companies is an account's right home, so it names the claimant and leaves the choice, which is a founder decision (2026-09-15) after two alternatives were rejected: a pick-time block (the guard is per account, and which accounts a consent covers is only known after the BankID round trip, so it would refuse a legitimate second company at the same bank) and a unique IBAN across companies in the schema (40 IBANs already sit in more than one company in prod, so it needs a cleanup first). Recorded in `DECISIONS.md` in 0cd9cdcc0, which is the only code change in this pass.

**Compliance Swarm A.8.3, cross-tenant disclosure of the claimant name: declined, not cross-tenant.** Claims are built only from rows owned by the same user: `fetchCrossCompanyAccountContext` filters `.eq('user_id', userId)` throughout (`extensions/general/enable-banking/lib/session-sharing.ts:157, 238, 372, 411`) and resolves company names via that user's own `company_members`. The name shown is the viewer's own other company, never a third party's, and the settings picker has rendered exactly this string since #2116. Naming it is the point: "bokförs redan i ett annat bolag" does not tell the user which of their companies to check.

**CodeRabbit, use `bank_claimed_notice` in `AccountPickerDialog`: declined.** That file imports no `useTranslations` at all and carries 17 hardcoded Swedish lines, including the "Synkas redan i X" label this notice explains. Localizing one sentence would leave an English user with one translated line in a Swedish dialog, and the key lives in the onboarding `books` namespace. Translating that dialog is its own change, not this PR.

/skeptic on the frozen commit returned MERGE; its three findings are already fixed in c2e54b206.

PR #2649, opened 2026-09-15
fix(bookkeeping): close field-lock gap on posted-to-cancelled journal entries

mattssonn:
## Problem

`enforce_journal_entry_immutability()` locks fields on posted -> reversed (an enumerated field comparison) but the posted -> cancelled branch had no such check: it fell straight through to `RETURN NEW`. Any company writer could change other fields (description, entry_date, ...) in the same statement that cancels a posted entry, with no audit trail. The trial balance excludes cancelled entries, so the entry simply disappears from the balance without a linked correction.

Reported privately per SECURITY.md scope (accounting data integrity / journal entry immutability).

## First principles

1. **Why did the problem occur?** The two structurally identical terminal transitions (posted->reversed, posted->cancelled) had the field-lock written by hand for one and never written for the other. The invariant ("no other field may change during a terminal transition") lived inside one branch's enumerated field list rather than as a shared rule, so the second branch could be added without it.

2. **What could be removed or simplified instead?** Merged both branches into a single full-row comparison excluding `status`/`updated_at`, which replaces the enumerated field list entirely. That is less code than adding a second list, and new columns are covered automatically instead of requiring a third hand-maintained list.

3. **Is this the best solution, or just the proposed one?** This closes "change a field while cancelling in one statement." It deliberately does not address the broader issue the reporter also raised: any company writer can cancel an ordinary posted entry outright, without proving it is a genuine cleanup. Closing that needs a transaction-local GUC (same pattern as `gnubok.allow_metadata_rattelse`), set only by `reverseEntry()`'s orphan cleanup and `cancelOrphanedPaymentEntry()`, and required by the trigger for the whole cancelled branch. That is a larger change across three call sites plus the trigger, and a product question about whether standalone cancellation should exist at all. Left as a separate founder decision.

## Testing

- New pg-real tests in `lib/bookkeeping/__tests__/engine.pg.test.ts`: status-only cancel still allowed; cancel + field change together now rejected.
- Verified on the staging Supabase branch: both new cases behave as expected, and the three untouched legitimate paths (notes-only annotation, opening-balance source-type retag, metadata rättelse) are unaffected.
- `npm run lint` clean. `npm run test:pg` was not run against local Docker Postgres (unavailable in this environment); staging verification stands in for it.

https://claude.ai/code/session_01T1cFNeW76mWPeNfJ9oHSLf

## Summary by CodeRabbit

- **Bug Fixes**
  - Journal entries can now be cancelled by changing only their status.
  - Additional edits to a posted journal entry during cancellation are blocked, preserving the entry’s recorded details.
  - Cancellation now consistently enforces field-level immutability for posted entries.


mattssonn:
## Findings disposition (`f0d35cf`)

### pg-real: 21 failures — FIXED

The first pass replaced both terminal-transition branches with one full-row comparison. That was wrong for reversal: the storno path legitimately writes `reversed_by_id` in the same statement as the status, which the original enumerated list permitted. Every failure was `Cannot modify fields of a posted entry during reversed`.

Reversed branch is now restored verbatim; the full-row comparison applies to `cancelled` alone, where every legitimate call site (`reverseEntry()` orphan cleanup, `cancelOrphanedPaymentEntry()`, the draft-cleanup paths) writes `status` and nothing else. Added a regression test that pins the storno shape (`status = 'reversed'` + `reversed_by_id`) open so this cannot be reintroduced.

### Swedish Accounting Compliance Review

**1. `posted -> cancelled` is not a valid rättelse track (BFL 5 kap 5 §) — acknowledged, not fixed here.**

This is correct and it is a real gap, but it is pre-existing behaviour that this PR does not introduce: the bare `posted -> cancelled` transition was already permitted before this diff. What this PR changes is strictly narrowing, it stops *other fields* being altered during the cancel.

Closing the underlying issue means requiring a transaction-local GUC (the `gnubok.allow_metadata_rattelse` pattern) for the whole cancelled branch, set only by the two legitimate cleanup call sites. That is a larger change across three call sites plus the trigger, and it carries a product question: whether standalone cancellation of an ordinary posted verifikation should exist at all. Flagged in the PR body as a founder decision and left out deliberately rather than silently.

**2. Guard-flag branches do not verify period-lock state in the trigger — verified as fine upstream.**

The reviewer's own condition ("If the calling RPCs enforce this check upstream, this is fine") holds. Checked against the live function bodies: `correct_entry_metadata`, `correct_entry_lines_inline`, and `mark_entry_as_opening_balance` all test period-lock state (`is_closed` / `locked_at` / lock date) before setting the GUC. The `enforce_period_lock()` trigger is also still in the path. No change made; these branches are recreated verbatim by the `CREATE OR REPLACE` and are not modified by this diff.

**Notes (un-reversal path, audit record on cancel):** both describe pre-existing behaviour outside this diff. They fold into finding 1's founder decision rather than being separately actionable here.

### Other bots

CodeRabbit: no actionable comments. PR Agent: no major issues, no security concerns. Compliance Swarm: no active findings.

https://claude.ai/code/session_01T1cFNeW76mWPeNfJ9oHSLf

PR #2656, opened 2026-09-15
fix(errors): let a domain error keep the sentence it was written with

yummymaya:
Two failures in one migration afternoon, both in the SIE import, both raised with a finished Swedish sentence naming exactly what to change, and both replaced on the way out.

A paused import said *"Nya SIE-importer är tillfälligt pausade"* and arrived as **"Förfrågan innehåller ogiltiga uppgifter"**, which sends its reader looking for a fault in their own file. The truth was an unset environment flag.

A klarmarkerat räkenskapsår said *"Öppna det igen under Inställningar > Bokföring > Räkenskapsår"* and arrived as **"Ett oväntat serverfel uppstod. Försök igen senare."** with a 500, on a state the user undoes themselves in two clicks.

An error can now say that its message is meant for the reader, and keep it.

<details>
<summary><b>Why it happens, and why the fix is opt-in</b></summary>

`errorResponse` answers a code by looking it up in the registry and using ITS sentence. That is right for the codes the registry knows: a Postgres 23505 has no useful message of its own, and `INTERNAL_ERROR` must never echo whatever a library threw.

It is wrong for the handful of errors raised WITH the sentence the reader needs. Those lose the only part that could have helped, and the second case loses the status too: an expected, self-serve state reported as a server fault.

`userFacing()` marks a message as ready to show. The code still decides status and classification, and a code the registry does not know drops to the ordinary path rather than inventing a status.

Opt-in, because letting every `Error` speak would leak developer English and library internals the first time something unexpected threw. The marker is a symbol and non-enumerable, so it never rides along into a details payload.

Swedish in both locales, like the engine's other domain errors: `.claude/rules/i18n.md` keeps regulatory wording untranslated because the English equivalents are ambiguous (verifikat, klarmarkerad, räkenskapsår).
</details>

<details>
<summary><b>Where it is marked</b></summary>

Two sources, both already writing for the reader:

`SIEJobValidationError`, whose every message is a finished instruction: *"Välj vad som ska importeras"*, *"Kontomappningarna täcker inga konton i SIE-filen"*, *"SIE-verifikation A12 (2021-03-05) ligger utanför räkenskapsåret"*.

The fiscal-period refusal that `closedPeriodRefusal` composes per state, which already distinguishes a klarmarkerat year from one closed by a year-end run from one that is merely locked, and gives each its own remedy.

Nothing else is marked. The rest of the codebase behaves exactly as before.
</details>

<details>
<summary><b>Verification</b></summary>

23 872 tests passing, types at baseline 533, lint and `check:guards` unchanged.

Six new tests hold both halves: the sentence survives and the status is right; an unmarked error still says nothing (an `ECONNREFUSED` stays behind `INTERNAL_ERROR`); an empty sentence is refused, being worse than the canned one; an unknown code falls back; and the marker does not appear to anything that enumerates the error.
</details>

## Summary by CodeRabbit

- **Bug Fixes**
  - SIE imports now show clear, actionable Swedish messages when blocked by paused imports or finalized fiscal years.
  - Expected import validation errors now return a client error instead of appearing as generic server failures.
  - Technical error details remain hidden unless an error is explicitly marked for display.
- **Error Handling**
  - User-facing error messages now preserve the intended error classification and localized response.
  - Validation details are retained alongside custom user-facing messages.


yummymaya:
Flagging that the next step here is a maintainer's, not mine.

Every GitHub Actions workflow on this PR is sitting at `action_required` on the current head (6e9b571dc): pg-real tests, Core Build, Checks, CodeQL, Compliance diff, zizmor. None has run, so the only green marks above are the third-party bots. Until someone with write access approves the runs, neither of us can tell whether this is mergeable: the PR body reports 23 872 tests passing and types at baseline locally, but that is my word, not CI's.

I have `pull` on this repo, so approving the runs is not something I can do, and I cannot request a reviewer either. A comment is the only channel I have.

One approval on the runs is all that is needed to turn this into a real review decision.

*(Separately, and not per-PR: the failing Vercel check is `Authorization required to deploy`, an account authorization on the Vercel side. It fails identically on every open PR here, including #2641 and #2677, so it is not a signal about any individual branch.)*

yummymaya:
Valid, and taken: a6f8c8ea4.

`buildResponse` rewrites both locale messages from `details.issues` so API consumers reading only `error.message` get the actionable field summary. The user-facing branch this PR adds reaches that same function with the sentence its author wrote in exactly those fields, so a marked error classified as `VALIDATION_ERROR` while carrying `details.issues` would have had its sentence replaced by the summary. That is the one thing this PR exists to prevent, at the one place it could still happen.

It now carries an explicit `authoredMessage` flag, set by the user-facing branch and by `errorResponseFromCode` when `ctx.messageSv` / `ctx.messageEn` were supplied. The summary still runs for everything else, and `details.issues` is untouched either way, so API clients keep the full array and the stable code.

## On reachability

Agreed that it is conditional, and worth writing down how conditional. It is not reachable today: no marked error populates `issues`, and the only writer of that shape is `DimensionValidationError`, whose code is `DIMENSION_VALIDATION_FAILED` rather than `VALIDATION_ERROR`, so the branch never fires for it. The Zod path in `lib/api/validate.ts` does use `VALIDATION_ERROR` with issues, but has no authored message to lose, which is exactly the case the summary is for.

Guarded anyway, because reaching it is one caller pairing `userFacing(err, 'VALIDATION_ERROR')` with a route that forwards Zod details, and the failure would be silent: a correct-looking envelope with the wrong sentence inside it. That is the kind that survives review.

## On the flag being explicit

Deliberate. Inferring it by comparing the entry against the registry would work most of the time and fail precisely when an author writes a sentence that reads like the canned one, which is the case least likely to be noticed by anyone.

## Tests

Two, as asked: a marked error with `VALIDATION_ERROR` plus `details.issues` keeps its sentence in both locales and still carries the issues array; an ordinary Zod failure with no authored sentence still gets the field summary. Checked that they bite by removing the guard, which fails the first and only the first.

jakobwennberg-oss:
Merged origin/main into this branch. The new files moved to src/lib/errors/ and DECISIONS.md merged cleanly. Tests for src/lib/errors, src/lib/import and src/app/api/import pass locally (one mappings test timed out under load and passed on a re-run).

Note for review: #2728 has since landed src/lib/import/sie-job-validation-response.ts, which already keeps the SIEJobValidationError sentence on the execute route and in the arcim wizard (and deliberately keeps the registry's bilingual text for codes with their own entry, such as SIE_IMPORT_UNSUPPORTED_ACCOUNT_CLASS). The half of this PR that is still new on main is the klarmarkerat/closed fiscal-period refusal in ensureFiscalPeriod, which still throws a plain Error and ends up as a 500. You may want to narrow the SIEJobValidationError marking to code === 'VALIDATION_ERROR' so the two paths agree.

PR #2683, opened 2026-09-17
chore(docs): commit the Next.js agent-rules block AGENTS.md keeps regenerating

yummymaya:
# What and why

`AGENTS.md` has been showing as modified in a clean checkout, and the change comes back after every build. It is not ours: `next dev` and `next build` upsert a `nextjs-agent-rules` block into it on every run.

The generator is `node_modules/next/dist/server/lib/generate-agent-files.js` (Next 16.3.4). Its rule is simple: if `AGENTS.md` exists it writes there and skips `CLAUDE.md`; only when neither exists does it scaffold both. Ours exists, so it wins the block. There is no env var, flag or ignore file that suppresses it.

## Why this and not deleting it

Deleting the block is what everyone tries first, and it is what the block itself warns against: the next `npm run dev` or `npm run build` puts it straight back. That is how it came to sit uncommitted across a whole working session, widening every `git status`, hiding real changes in the noise, and standing one `git add -A` away from being swept into an unrelated commit.

Committing it is the only state where the working tree is actually clean.

**Verbatim on purpose.** The generator re-upserts the text between the `BEGIN:`/`END:` markers, so any rewording of ours would be overwritten on the next build. Editing it would produce exactly the churn this PR removes.

## What was checked rather than assumed

The block's claims are load-bearing guidance for future agents, so they were verified, not taken on the block's own word:

- `node_modules/next/dist/server/lib/generate-agent-files.js` exists and is the writer.
- `node_modules/next/dist/docs/` exists and carries the guides the block points at (`01-app`, `02-pages`, `03-architecture`, `04-community`, `index.md`).
- `CLAUDE.md` does not already host the block, so there is no duplicate to reconcile.

## Alternative considered

**Adding `AGENTS.md` to `.gitignore`.** Rejected: the file is a curated pointer to `.claude/rules/*` and is checked in on purpose. Ignoring it to silence a vendor block would stop tracking our own guidance too.

## Checklist

- [x] Documentation only; no code, no tests, no migrations
- [x] No imports from `@/extensions/`

## Summary by CodeRabbit

* **Documentation**
  * Added guidance for development tools and contributors working with the project’s Next.js version.
  * Guidance highlights potential breaking changes and directs readers to the relevant framework documentation before making updates.
  * Clarified that the guidance is automatically maintained during local development to keep project changes consistent.


yummymaya:
Closing: the current arrangement looks deliberate, not accidental. AGENTS.md being the file Next.js upserts into (and CLAUDE.md therefore staying untouched) is a reasonable thing to have chosen on purpose, and if so the reappearing block is a known cost rather than something to commit away.

Leaving the call to the maintainers. The branch stays up if anyone wants the commit: 4a9703d0e.

For whoever picks this up later, the facts as checked: the writer is `node_modules/next/dist/server/lib/generate-agent-files.js` (Next 16.3.4), it writes to AGENTS.md whenever that file exists and skips CLAUDE.md, and it exposes no env var or flag to suppress the block.

Issue #2585, opened 2026-09-13
Migration: prefill account momskod from the source system's VAT code (Fortnox /accounts VATCode) instead of guessing from the account name


mattssonn:
## Original request

Discord, 2026-09-13 (migrating user, name withheld):

"Jag försökte byta till accounted här om dagen från Fortnox men tyckte det var svårt att förstå konto-mappningen mot momskoder etc. Fortnox exponerar konton genom sitt API /accounts. kan inte accounted själv hämta korrekta momskoder för samtliga konton?"

"i åtminstone ett par fall såg jag att accounteds föreslagna momskod på ett konto skiljde sig från den befintliga momskoden om jag gick in och kollade på kontoplanen i Fortnox"

Another participant: "Momskoder är rätt centralt, särskilt om man har en kontoplan som inte är identisk med BAS".

## Source

Discord community channel, 2026-09-13. Relayed by Emil.

## Investigation so far

- The provider migration pulls a SIE4 file from Fortnox (`/sie/4`). SIE carries no per-account momskod, so the account mapping step falls back to `suggestVatTreatment`, a regex over the account name (`lib/vat/account-vat-treatment.ts`). Any account not named "25%", "EU", "omvänd" etc. gets a guess or nothing, which is the mismatch the user saw.
- Fortnox exposes the per-account momskod as `Account.VATCode` on `/accounts`. `mapFortnoxToAccountingAccount` already reads it into `AccountingAccountDto.vatCode`, and the `AccountingAccounts` resource is registered in `lib/providers/fortnox/config.ts` under the `bookkeeping` scope we already request. Nothing consumes it. Visma (`VatCodeId`), Björn Lundén (`vatCode`) and Briox (`vat_code`) mappers carry the same field, equally unused.
- Related: #1457 (per-account VAT treatment picker, closed) noted "SIE4 carries no per-account momskod, so import can at best suggest treatments". That is true for manual SIE upload; for API migrations the source system's code is one call away.

mattssonn:
```
VERDICT   BUILD-SMALLER: Fortnox only; Visma/BL/Briox expose opaque code ids whose semantics are unverified
WHAT      Migrating from Fortnox, the mapping step's momskod column comes prefilled from the account's VAT code in Fortnox ("Från Fortnox: MP1"), already marked reviewed, instead of a guess from the account name.
[...]

PR #2701, opened 2026-09-17
fix(import): let a settled "ingen momskod" stay settled

### LINKED ISSUE #2700: A deliberate "ingen momskod" answer is asked again on every later import
## What happens

Importing 2021 (SIE + Spiris chart) left 57 rows in "momskoder att granska", which were answered and imported. Importing 2022 into the same company then asks again about **13 of them**:

`4059, 4549, 3740, 3960, 3970, 3971, 3980, 4000, 4400, 4500, 4700, 4900, 4960`

Every one is a class 3 or class 4 account whose correct answer is "Använd BAS-standard / Ingen", and every one was already answered that way during the 2021 import. Only 3 of the 16 rows in the 2022 list are genuinely new (`4535` newly coded in the file, plus `3730` and `4599`).

Observed in production on a real two-year Spiris import, after #2641 merged.

## Why it happens

`chart_of_accounts.default_vat_treatment` is nullable, and NULL carries two different meanings that nothing distinguishes:

- nobody has taken a position on this account, and
- someone took a position, and the position is "this account has no VAT handling".

The short-circuit that spares a settled account from being asked again reads the treatment itself (`lib/import/account-vat-treatment.ts:164`):

```ts
const existing = existingByNumber.get(mapping.targetAccount)
if (existing?.default_vat_treatment) {
  return { ...mapping, vatTreatmentReviewed: true, requiresVatTreatmentReview: false }
}
```

A null treatment is falsy, so the row falls through to

```ts
requiresVatTreatmentReview: accountClass === 3 || accountClass === 4 || Boolean(suggestion)
```

and is asked again. `buildSIEVatDefaults` (`lib/import/account-sync.ts:115`) does persist the answer, but an answer of "none" persists as `default_vat_treatment: null`, which is the same bytes as never having answered.

So the review list can only ever shrink for accounts that end up WITH a treatment. Accounts that correctly have none come back on every subsequent import: import 2023 and 2024 and the same 13 rows appear each time.

## Why it matters

Nothing is booked wrong: an account with no treatment reaches no ruta either way, and the hardcoded `ACCOUNT_RUTA` still carries the accounts it covers.

The cost is to the control itself. "Momskoder att granska" is the gate that stands between a source system's momskod and this company's VAT declaration, and #2641 made it load-bearing. A gate that re-asks a question already answered, every year, trains the user to clear it with "Bekräfta alla föreslagna" without reading, which is exactly the habit the gate exists to prevent. The noise also grows with the number of years imported while the signal does not.

## Shape of a fix

The conflation is between *decided* and *undecided*, not between *has a treatment* and *has none*, so the fix should record the former rather than widen the latter. Two candidates:

1. **A persisted "settled" flag on `chart_of_accounts`.** One column, written where `buildSIEVatDefaults` already writes the treatment, read in the one short-circuit above. The treatment enum stays a pure domain concept and no other consumer changes.
2. **A sentinel member in `AccountVatTreatment`** (e.g. `'none'`). Reuses the existing short-circuit with no schema change beyond the CHECK, but every consumer of the enum (`resolveVatTreatmentRuta`, `defaultRateForVatTreatment`, the declaration's dynamic path, the MCP tools, the account editor) would have to learn a member that means "not a treatment", most by mapping it back to the null behaviour they already have.

(1) looks right for that reason, but the choice belongs in the PR.

yummymaya:
Closes #2700.

# What and why

Importing 2022 into a company that already had 2021 imported asked again about **13 of the 16** rows in "momskoder att granska". Every one was an account answered "Använd BAS-standard / Ingen" the year before:

`4059, 4549, 3740, 3960, 3970, 3971, 3980, 4000, 4400, 4500, 4700, 4900, 4960`

Only 3 of the 16 were genuinely new. Found on a real two-year Spiris import in production, after #2641 merged.

## Why the problem occurred

`chart_of_accounts.default_vat_treatment` is nullable, and NULL carried two meanings nothing could tell apart:

- nobody has taken a position on this account, and
- someone took a position, and it is that this account has no VAT handling.

The short-circuit that spares a settled account from being asked again read the treatment itself (`lib/import/account-vat-treatment.ts`), so the second kind was indistinguishable from the first. `buildSIEVatDefaults` did persist the answer, but an answer of "none" persists as `default_vat_treatment: null`, which is the same bytes as never having answered.

So the review list could only ever shrink for accounts that end up **with** a treatment. Accounts that correctly have none came back on every later import: import 2023 and 2024 and the same 13 return each time.

## Why it was worth fixing

Nothing was booked wrong. An account with no treatment reaches no ruta either way, and the hardcoded `ACCOUNT_RUTA` still carries the accounts it covers.

The cost was to the control itself. "Momskoder att granska" is the gate between a source system's momskod and this company's momsdeklaration, and #2641 made it load-bearing: it is where a `23` on the wrong account, or a 25 % default on a 6 % basis, is supposed to be caught. A gate that re-asks a question already answered, every year, teaches people to clear it with "Bekräfta alla föreslagna" without reading, which is exactly the habit it exists to prevent. The noise grows with each year imported while the signal does not.

## What it does

One column, `chart_of_accounts.vat_treatment_reviewed_at`, written where the reviewed answer is already persisted (`buildInsertRow` and the existing-account update pass in `account-sync.ts`) and read in the one short-circuit:

```ts
if (existing?.default_vat_treatment || existing?.vat_treatment_reviewed_at) {
```

A timestamp rather than a boolean, because "when was this decided" is the question an auditor asks about a setting that steers a declaration, and `IS NOT NULL` is the same test a boolean would give.

**A settled "none" takes the same precedence a settled treatment already has.** "Chart treatment > provider code > label, because a treatment the company set in Accounted is a deliberate later edit and the provider code is the user's own configuration in the system they are leaving" (DECISIONS 2026-09-14, #2585). A settled "none" is the same kind of deliberate edit, so it is not given a weaker rank than a settled treatment.

An earlier revision of this description claimed the answer "still yields to a file that newly speaks". That is only true when `applySourceVatCodes` runs last, which is the chart-upload flow; when `enrich` runs last (`enrichChangedAccountMappingWithVat`, after the user re-points a target) the settled answer wins silently, for both kinds of answer. That asymmetry predates this PR and deserves its own issue rather than a change here, since resolving it means choosing which of the two orders is the intended rule.

## What was considered instead

**A `'none'` member in `AccountVatTreatment`.** It reuses the existing short-circuit with no new column, only a CHECK change. Rejected because the conflation is between *decided* and *undecided*, not between *has-a-treatment* and *has-none*: a sentinel widens a domain enum with a member meaning "not a treatment", and every consumer (`resolveVatTreatmentRuta`, `defaultRateForVatTreatment`, the declaration's dynamic path, the MCP tools, the account editor) would have to learn it, most by mapping it straight back to the null behaviour they already have. One column read in one place is the smaller change and leaves the enum a pure domain concept.

**Persisting the review state on the import job.** Wrong layer: the answer is a fact about the account, not about one import.

## No backfill, deliberately

The reader accepts either signal, so an account settled to a real treatment before this column existed is still recognised, exactly as before. The rows this cannot rescue are the ones already answered "none", which are precisely the rows indistinguishable from never-asked: that is the bug. They are asked once more and stick from then on. A backfill would mean guessing which NULLs were answers.

## Tests

`lib/import/__tests__/account-vat-treatment.test.ts`: an account settled to no treatment stops being asked; the same account without the stamp is still asked (the other half of the pair); an account settled before the column existed is still recognised, which is why no backfill is needed.

`lib/import/__tests__/account-sync.test.ts`: a reviewed mapping with a null treatment stamps the review; an unreviewed one does not.

Checked that they bite: reverting the change fails exactly those three and nothing else.

## Checklist

- [x] `npm run lint` (0 errors), `npm test` (24 067 passed), `npm run build` pass locally; `npm run check:types` at baseline (533); `npm run check:guards` passes
- [x] New logic has tests
- [x] Migration is a new file; no existing migration edited. A nullable column with no constraint, trigger, RLS policy or default, so it carries `-- pg-test: skip` with the reason
- [x] No imports from `@/extensions/`

## Summary by CodeRabbit

- **Bug Fixes**
  - Accounts explicitly reviewed as requiring no VAT treatment are no longer repeatedly returned for VAT review.
  - VAT review status is now preserved for both new and existing accounts.
  - Settled VAT choices take precedence over conflicting imported provider information.

- **Data Updates**
  - Added review timestamps to distinguish accounts intentionally set to no VAT treatment from accounts that have not yet been reviewed.
  - Accounts without a recorded VAT decision remain eligible for review.


yummymaya:
The behandlingshistorik finding rests on "unless such history is already captured elsewhere outside this diff (unsure)". It is, and that is why this PR adds a column rather than a log table.

## The history already exists, per change, immutable and dated

`chart_of_accounts` carries the `write_audit_log()` trigger on INSERT, UPDATE and DELETE (`20240101000018_audit_triggers.sql:94`). It has never been dropped; the only later mention disables and re-enables it inside the account-deletion RPC. The trigger stores the **whole row before and after** as `old_state` / `new_state` jsonb with `created_at`, and `audit_log` is append-only: `audit_log_immutable()` blocks UPDATE and DELETE on it.

So every change to `default_vat_treatment` and `default_vat_rate` already produces a dated, immutable record of what the value was and what it became. Not the most recent timestamp: the full sequence.

That record is already surfaced in the statutory report. `behandlingshistorik.ts` lists `chart_of_accounts` in `AUDITED_TABLES` (line 203), and `accountAuditEvent`'s UPDATE branch renders a field diff over `ACCOUNT_FIELDS`, which names exactly these two:

```ts
default_vat_rate: 'Momssats',
default_vat_treatment: 'Momshantering',
```

Change an account's VAT default twice and behandlingshistorik shows two "Konto ändrat" events, each dated, each naming the old and new Momshantering. The reconstruction the finding asks for is available today.

## Why a separate log table would be worse

A `vat_treatment_review_log` would be a second record of a fact `audit_log` already holds, with its own write path that can fail, drift or be forgotten, while the statutory report keeps reading the first one. This repo has been bitten by exactly that shape more than once. One source of truth, already immutable and already reported, is the stronger position.

## What the new column is, and is not

`vat_treatment_reviewed_at` is not the history. It is a single current-state fact the import needs to answer one question: has a person taken a position on this account at all. It is deliberately **not** in `ACCOUNT_FIELDS`, so re-confirming an account without changing its treatment renders nothing in behandlingshistorik. That is intentional and follows the precedent stated in the `SETTINGS_FIELDS` comment just above it: fields that change on ordinary operation ("next_invoice_number alone would otherwise add a row per issued invoice") are kept out of the statutory history. Re-affirming an unchanged behandlingsregel is not a change to it; the audit row still exists underneath if anyone needs it.

## On the no-backfill note

The settlement dates for previously-answered rows are not unrecoverable either. Those accounts were written by an earlier import, and that write is in `audit_log` with its date and its before/after state. What is missing is only the convenience column, which is the thing being introduced, not something being lost.

## On the RLS note

Fair to ask, and it holds. RLS in Postgres is row-level: a policy's `USING` / `WITH CHECK` predicate is evaluated per row and does not enumerate columns, so a new column is covered by the existing `chart_of_accounts_select` / `_insert` / `_update` policies without any change. The one mechanism that would need extending is column-level `GRANT`s, and there are none on this table: grants are table-level throughout these migrations. Nothing about the policies changed, and nothing needed to.

No code change from this review. If you would rather see re-confirmations in behandlingshistorik as their own event type, say so and I will add it as a separate action rather than as a field diff, but I would argue against it on noise grounds.

yummymaya:
Not taken, because it asks to invert a rule this repo decided deliberately, and I have corrected my own PR description instead. Detail below, because the finding is a good one: it found a real asymmetry, just not a defect.

## The precedence is deliberate and recorded

`DECISIONS.md`, 2026-09-14 (#2585):

> Precedence is chart treatment > provider code > label, because a treatment the company set in Accounted is a deliberate later edit and the provider code is the user's own configuration in the system they are leaving.

The settled-account branch returning before it looks at `providerVatTreatment` **is** that rule. It is pinned by an existing test, `applySourceVatCodes (#2585) > yields to a treatment the company already set on the account`, which runs `applySourceVatCodes` first and then `enrichAccountMappingsWithVat`, and asserts the company's treatment wins with no review.

I implemented the suggestion before checking. It failed that test, which is how I found the decision. Reverted in b22c37545, and the branch now carries the reason so the next reader does not repeat the attempt.

This PR's change does not weaken the rule, it extends it consistently: a settled "none" is the same kind of deliberate later edit as a settled treatment, so it takes the same precedence rather than a weaker one. The commit adds a test pinning the "none" half next to the existing one for the treatment half.

## Where the finding is right

The reachable path is not the one named. On the standard SIE import, `enrichAccountMappingsWithVat` runs at `import/page.tsx:835` **before** any chart is applied, and SIE4 carries no VAT codes, so there is no `providerVatTreatment` on the row to compare against. Inside `applySourceChartCsv` the provider fields are explicitly nulled before enrich runs (`apply-source-chart.ts:131`) and `applySourceVatCodes` runs after. The Fortnox path never calls enrich at all: the arcim-migration extension calls `applySourceVatCodes` server-side and the workspace uses the result directly.

The path that does reach it is `enrichChangedAccountMappingWithVat` (`account-vat-treatment.ts:242`), called from the mapping step when the user re-points a target account. By then the chart's codes are on the row and nothing runs after to re-open it.

So both orders are reachable and they disagree:

| runs last | on disagreement |
|---|---|
| `applySourceVatCodes` (chart upload) | reopens for review |
| `enrich` (after a re-point) | settled answer wins, silently |

That asymmetry is real and predates this PR. It deserves its own issue rather than a change here: resolving it means choosing which of the two is the intended rule, and one of them is written down in DECISIONS while the other is not.

## What I did change

My PR description claimed "a settled 'none' still yields to a file that newly speaks", which is only true when `applySourceVatCodes` runs last. That over-claim is corrected in b22c37545 and in the description.

yummymaya:
Filed the ordering asymmetry as #2702, with the two reachable orders, the recorded decision that favours one of them, and the argument that favours the other.

Noted there that whichever way it is resolved should apply to both kinds of settled answer, so this PR's extension does not pre-empt it: it gives a settled "none" the rank a settled treatment already had, rather than inventing a rank for it.

yummymaya:
Not taken. The premise is that a remapped target can "later be treated as unanswered and asked the same settled decision again", and that cannot happen: a non-identity mapping is never asked in the first place, on any of the three paths.

## The exclusion is symmetric, not one-sided

`lib/import/account-vat-treatment.ts`:

- `enrichAccountMappingsWithVat` line 150 returns early for `sourceAccount !== targetAccount` with `vatTreatmentReviewed: true, requiresVatTreatmentReview: false`. It short-circuits **before** the settled-account check, so what the target row does or does not hold never enters into it.
- `applySourceVatCodes` line 107 returns the same mappings untouched, so a remapped row never even receives a `providerVatTreatment`.

`lib/import/account-sync.ts`:

- `buildSIEVatDefaults` skips them on the write side.

`components/import/AccountMappingStep.tsx` gates the momskod control on `requiresVatTreatmentReview && !vatTreatmentReviewed` (lines 167, 642), which is false for these rows, so no control is offered and there is no answer to persist.

Read, write and UI agree. The finding describes the write side in isolation, where it reads like a gap; paired with the read side it is the same rule stated twice.

## Why it is the right rule

The VAT treatment question belongs to identity mappings because that is where "this source account **is** this chart account" holds, so a treatment on the target describes the same account the file describes. On a remap they are different accounts: the file says something about 3057 and the user pointed it at 3107. Carrying the file's momskod onto 3107 would assert a VAT box for an account the source system never described, which is the move #2585 already refused for a weaker case:

> IT (24) on a 46xx underentreprenör account answers null because Accounted files class 4 outside 4425-4427 on ruta 23, and asserting a box Fortnox disagrees with is worse than falling back to the label.

Implementing the suggestion would write a reviewed VAT state onto an account the user was never asked about, and mark it settled so nobody is asked later either. That is a worse failure than the one it aims at, because it is silent and it is wrong rather than merely repeated.

## What does happen, and is correct

If a later import contains an identity mapping for 3107 (the file itself carries that account), that row **is** asked. Not a re-ask: nobody has answered for 3107 as an account in its own right, and the earlier remap was an answer about where 3057's amounts should land, not about 3107's VAT handling. Once answered, #2701 makes it stick, including when the answer is "Använd BAS-standard / Ingen".

No code change.

jakobwennberg-oss:
Merged current main into this branch and renumbered the migration to 20260923121000 so it sorts after main's latest (not applied on prod). Head ed291ba8a. Lint clean on touched files, import tests pass.

PR #2707, opened 2026-09-17
feat(arkiv): phase 9c, search in the app, asking through the customer's own assistant

jakobwennberg-oss:
Stacked on #2705 (phase 9b). Draft, nothing merges; review on this branch's preview or the local stack. Nothing user-visible without `ARKIV_COMPANY_IDS`.

## What this is

Plan step 3: search, not chat, in the app; asking goes through the customer's own agent.

- `lib/arkiv/search.ts`: the retrieval `gnubok_search_records` already did, lifted out so the page and the tool run the same search. Pages by text (best page first, the matching passage), agreements by title or counterparty, facts by value or by the Swedish name of the predicate ("momsperiod" finds `vat_period`).
- `GET /api/arkiv/search?q=&kinds=&limit=`: the same hits presented for a person. A page hit is titled like the table titles it (agreement title first, then the read fields, then the file stem), with the file name as a second line when it says more (duplicate uploads, scans). The passage loses its markdown markers, never its `<b>` marks. Every hit carries the record to open and, for a page hit, the original file at that page.
- `/arkiv`: one search field under the header. While a query is on, grouped hits (Dokument, Avtal, Fakta) stand where the graph and the table were; cleared, the page is as before. Next to the field: "Hellre fråga? Koppla din assistent, så läser den arkivet och svarar där du redan arbetar", pointing at `/settings/api`.
- `/arkiv/dokument/[id]?page=N`: the record opens its text at that page, marked, so a hit lands on the passage.
- The table keeps its own text filter only on the Myndighet page, where there is no search above it.
- Event `arkiv_searched` (kinds, query length, hit count; never the words).

## First principles

Why a search and not a chat: the in-app assistant is being removed, and the data pass showed people arrive with a document or a question about one, not a conversation. A search that answers with the page is the shortest path to a fact inside the app; anything longer than that belongs in the tool the person already talks to, which is why the field carries the connector setup as the way to ask.

Why one core: the MCP tool and the page were about to hold two copies of the same three queries. Now there is one, and the tool's behaviour is unchanged (its tests pass untouched).

## Verified

Unit: `lib/arkiv/__tests__/search.test.ts` (order, kinds, limit, wildcard neutralising, predicate label match), `app/api/arkiv/search/__tests__/route.test.ts` (404 outside the rollout, 400 short query and unknown kind, titles, hrefs, event), MCP tool and strict-schema suites unchanged and green. 1,565 tests across `lib/arkiv`, `app/api/arkiv`, `extensions/general/mcp-server`; type ratchet at baseline 533; `check:guards` green; `apiskill:check` up to date.

Local stack: eight queries answered in 100 to 360 ms; `kinds=party` 400; `kinds=agreement,fact&limit=2` honoured; MCP `gnubok_search_records` returns the same three hits as before. Headless UI: grouped results with page links, graph and table stepping aside, empty state, a hit opening the record scrolled to page 6 of a seven-page loan agreement.

Both languages have the six new strings.


jakobwennberg-oss:
Superseded by #2923, which replays this phase's commits onto today's main with the conflicts resolved (the read queue merged with the rollout-first order, the photo fit before the reader tier, the fence and the coded refusals from main). Merged there; closing this one. The branch stays.

PR #2720, opened 2026-09-18
feat(salary): avvikelseperiod, read absence and worked days from the previous month

jakobwennberg-oss:
## What

Adds the avvikelseperiod (deviation period) to salary runs, so a company can pay this month's fixed salary and take last month's absence, sick pay, VAB, tjänstledighet and worked hours: the standard Swedish "innevarande månads lön, föregående månads avvikelser" setup.

Reported by Frey (Simon Hällqvist, 2026-09-18): their next pilot on the public API stalled because absence registered in the previous month never reached the current month's run.

- `company_settings.salary_deviation_period`: `same_month` (default, today's behaviour) or `previous_month`. New select in Inställningar > Lön.
- `salary_runs.deviation_period_start` / `deviation_period_end`: the window snapshotted on each run at creation. `NULL` on both means the pay month, which is what every existing run meant.
- `POST /api/v1/.../salary-runs`, the dashboard one-click route and `gnubok_create_salary_run` all resolve the window (explicit `deviation_period_start`/`deviation_period_end` win, else the setting) and return it. The MCP preview shows the resolved window before approval.
- The calculation engine reads `salary_absence_days` and `salary_worked_days` from the run window; the fixed salary and its mid-month proration stay on the pay month. The AGI frånvarouppgifter (VAB/parental) follow the same window. The payslip prints "Avvikelseperiod" when it is not the pay month; the run header says "Avvikelser 2026-08-01 till 2026-08-31".
- A window that overlaps another live run of the company is refused at creation with `409 SALARY_RUN_DEVIATION_PERIOD_OVERLAP` naming the conflicting run. One date without the other, an inverted window or one over 62 days is `400 SALARY_RUN_DEVIATION_PERIOD_INVALID`.
- Correction runs inherit the window of the run they replace.
- Docs: cookbook section + pitfall, v1 endpoint pitfalls/example, MCP payroll skill, regenerated `skills/accounted-api`.

## First principles

Why it occurred: the run only knew its pay month, and every calendar-derived read (absence days, worked days, sick-pay lookback, AGI absence events) used the first and last day of that month. There was no concept of "which calendar this salary settles".

What was removed: nothing had to go, but the fix deliberately does not add a company-wide offset that the engine reads at calculate time. The window is frozen on the run so a settings change can never retroactively move an already-calculated run, and the overlap guard sits at creation because that is the only point where the caller can still choose a different window. See DECISIONS.md 2026-09-18.

Not in scope: `salary_deviation_period` on the v1 settings PATCH (that endpoint mirrors the MCP bank/contact tool field for field; integrators pass the dates per run), and arbitrary windows as a company setting (explicit per-run dates already cover e.g. 16th to 15th).

## Tests

- `lib/salary/__tests__/deviation-period.test.ts`: month/previous-month windows, legacy fallback, explicit-window validation, setting resolution, overlap guard.
- `create-run`, dashboard route, v1 route (201 / 409 overlap / 400 invalid / dry-run preview carries the window) and MCP staged tool (preview carries the window) updated.
- `supabase/migrations/__tests__/salary-deviation-period.pg.test.ts`: the CHECK constraints on both tables, run against a fresh migration replay locally.
- MCP tools/list payload stays under the 60 500 ceiling; paid for inside the payroll cluster (note in the bench file).

## Switch-on

The migration auto-applies on merge (Supabase branching). Nothing changes for existing companies: the default is `same_month`. Frey sets `previous_month` once in settings, or passes the two dates on every `POST /salary-runs`.

## Summary by CodeRabbit

- **New Features**
  - Configure salary deviation periods for the same or previous month.
  - Override the period for individual salary runs with custom dates.
  - View deviation periods in run details, headers, previews, and payslips.
  - Salary calculations and AGI absence data use the run’s saved period.
- **Bug Fixes**
  - Prevent overlapping live runs from processing the same days.
  - Validate incomplete, invalid, reversed, or overly long date ranges.
- **Documentation**
  - Updated payroll API references and guidance for deviation periods.


jakobwennberg-oss:
Review triage (986e92c):

**Fixed**
- Overlap guard now includes correction runs and only skips the corrected original (CodeRabbit deviation-period.ts:220 and the compliance bot). The period_year prefilter is gone as well (deviation-period.ts:223): an explicit window is not tied to the pay month.
- validateExplicitWindow rejects overflowed days such as 2026-02-30 by round-tripping the parsed date (deviation-period.ts:135).
- The salary_runs CHECK mirrors the two-month cap (compliance bot). Migration edited in place since it is unshipped; pg test extended.

**Not changed, with reason**
- Atomic overlap validation across insert paths (create-run.ts:54): the unique index on (company, period_year, period_month) already stops the realistic concurrent case (two creates of the same pay month). Two different pay months created concurrently with overlapping windows only happens during a mid-stream switch, which the setting deliberately does not support. An exclusion constraint over NULL-defaulted ranges is a heavier change than this PR warrants; noted in DECISIONS.
- NOT VALID + later VALIDATE for the CHECK constraints (migration:48): company_settings and salary_runs are a few hundred rows each; the validating scan under ACCESS EXCLUSIVE is milliseconds.
- Lookback anchored on the deviation window (compliance note): the sjuklön lookback must precede the days being classified, and those are the window's days, so anchoring on deviation.start is the correct reference.
- Worked-days writes under a booked run and payment-file archiving (compliance findings on #2722) are pre-existing for absence and for the dashboard routes; filed as https://github.com/erp-mafia/accounted/issues/2723 and https://github.com/erp-mafia/accounted/issues/2724.

PR #2726, opened 2026-09-18
feat(vat): momsruta per 26xx account so a non-BAS chart declares in the right box

jakobwennberg-oss:
## What

A 26xx VAT account can now carry the momsdeklaration box it feeds (`chart_of_accounts.vat_box`: 10-12, 30-32, 60-62, 48, or `none`; null keeps the BAS mapping by number).

- **Report**: the dynamic VAT account loader (`lib/reports/vat-revenue-accounts.ts`) treats an overridden 26xx account like a class 3-6 momskod: it leaves its BAS box and feeds the chosen one. The declaration, the filing gate and the close checks (`RC_OUTPUT_MISSING`, `IMPORT_OUTPUT_WITHOUT_BASE`) all read rutor, so they follow without further changes.
- **UI**: `EditAccountDialog` shows a "Momsruta i momsdeklarationen" select on 26xx accounts (not 2650). When the account name suggests a different box than the number gives (Fortnox layout: "varuförvärv EU" on 2615, "import" on 2616, "tjänster utanför EU" on 2617), the help text says which box the name implies. The hint never sets the box by itself.
- **API and MCP**: `PUT /api/bookkeeping/accounts/[number]`, `POST /api/bookkeeping/accounts`, the pending-operation schema and `gnubok_update_account` accept `vat_box` and refuse it outside 26xx.
- **DB**: migration `20260918153000_chart_of_accounts_vat_box.sql` adds the column and a CHECK mirroring `ACCOUNT_VAT_BOXES` and `isVatBoxAccount`; pg test included.

## Why

The declaration mapped every 26xx account by BAS number only. A chart laid out before the 2015 BAS change books the same VAT on other numbers, so ruta 30 and 60 came out wrong every month and an account BAS does not know (2617) fell out of the declaration entirely, with the close check reporting the follow-on errors. The `swedish-vat` reference notes that legacy migrations must remap for exactly this reason; renaming the accounts on import would rewrite the company's history, so the account carries the ruta instead.

## First principles

- **Why did this occur**: the box map is a static BAS table and the class-3-6 momskod cannot express "this account IS the output VAT", so there was no place for a company to say where a 26xx balance belongs.
- **What can be removed**: nothing yet. The import wizard could flag 26xx deviations with `vatBoxDeviationFromName` at import time; that is a follow-up, not part of this change.
- The `gnubok_update_account` schema carries no enum for `vat_box` and the account tool descriptions were shortened: the tools/list ceiling (`payload-size.bench`) sat at the limit and the guard's rule is to pay inside the same tools (DECISIONS.md).

## Context

Easy Online Stores evaluation brief 2026-09-16, finding F1 (ruta 49 off by exactly the 2617 balance every month, January to July 2026).

## Verification

- `npx vitest run lib/vat lib/reports lib/api lib/pending-operations app/api/bookkeeping/accounts extensions/general/mcp-server`: 3751 passed
- `npm run check:lint`, `npm run check:guards`: OK
- `npm run check:types`: only pre-existing `smtp-service.ts` (missing `@types/nodemailer` in the shared local node_modules, not this change)
- pg test `supabase/migrations/__tests__/chart-of-accounts-vat-box.pg.test.ts` runs in CI


jakobwennberg-oss:
Answers to the compliance review:

- `vat_box: 'none'` could drop a VAT-carrying account from the declaration: agreed, removed in 3949d5aa0. The override now only routes a 26xx balance to a box; the BAS mapping stays the fallback and there is no opt-out value. CHECK, Zod schema, MCP validation, UI and tests follow.
- Side hard-coded by box: it is the same rule as the static `ACCOUNT_RUTA` map (credit for output boxes, debit for ruta 48), and `rutorFromTotals` reads the signed balance, so a credit-note-heavy period comes out negative rather than lost. Same behaviour a BAS-numbered account already has.
- Not period-dated: same design as `default_vat_treatment`, which the declaration already reads as current state. Submitted declarations are stored as filed, so a later change does not rewrite history; it changes what the next period computes, which is what the account's owner intends when they set it.
- Audit trail: `chart_of_accounts` carries `audit_chart_of_accounts` (migration 20240101000018), so every `vat_box` change lands in `audit_log` with who and when.

PR #2727, opened 2026-09-18
fix(mcp): sie_import_status without import_id, skattekonto hint, compact list_accounts

jakobwennberg-oss:
## What

Three MCP findings from the Easy Online Stores evaluation brief, in one PR.

1. **`gnubok_sie_import_status` without `import_id`** failed with the Postgres error `invalid input syntax for type uuid: undefined`. The tool now lists the company's ten most recent imports (`kind: "list"`: import_id, filename, status, job_state, created_at, replaced_at) when the id is omitted, and rejects a malformed id with a clear message before any query. `import_id` is no longer `required` in the input schema. The output schema is a `oneOf` of the existing status shape and the new list shape.
2. **`Unknown account_key "skattekonto" for this company`** before Skatteverket is connected. One helper (`reconciliation-key-error.ts`) now explains every unresolved key: for `skattekonto` it says the account is not connected (with the tool to use) or connected but not yet synced; for any other key it lists the keys the company does have, and adds the format when the key is malformed. All five throw sites (status, items, match, signoff, residual) use it.
3. **`gnubok_list_accounts` at 88 kB for 379 accounts.** New opt-in `detail: "compact"` (account_number, name, class, is_active, default_vat_treatment) and `limit` / `offset`; `total` on every answer. Default output keeps the previous rows and columns.

## First principles

- Why did this occur: the dispatcher does not validate `required` against `inputSchema`, so a tool that trusts `args.import_id` forwards `undefined` to Postgres. The reconciliation services signal "cannot resolve" with a null that carries no cause, and each tool wrote the same anonymous message over it. `list_accounts` was written for a BAS chart an agent reads once, not a 379-account chart re-read on every turn.
- What can be removed: five hand-written throw sites became one helper. Nothing new in core.
- Ceiling: tools/list had 2 tokens of headroom. The new `list_accounts` fields are paid for by shorter descriptions on the tools this PR touches (measured 60 489 against the 60 500 ceiling, from 60 498); no read was demoted. See DECISIONS.md.

## Context

Easy Online Stores evaluation brief 2026-09-16, finding F6.

## Verification

- `npx vitest run extensions/general/mcp-server`: 129 files, 1449 tests pass. New: `reconciliation-key-error.test.ts`; new cases in `sie-import-status`, `reconciliation-tools`, `account-tools` and `call-tool-bridge` (the bridge test pins that an omitted `company_id` resolves to the API key default, so the company half was never the bug).
- `npx eslint` on touched files: 0 errors. `npm run check:lint`: OK.
- `npm run check:types`: the only reported error is `@types/nodemailer` missing from the shared local `node_modules`, unrelated to this change.

## Summary by CodeRabbit

- **New Features**
  - View recent SIE imports when no specific import ID is provided.
  - Browse chart-of-accounts results with compact details, totals, limits, and offsets.
  - Use clearer reconciliation account-key guidance, including Skatteverket connection and synchronization status.
  - Omit the company ID when calling supported tools to use the API key’s default company.

- **Bug Fixes**
  - Invalid import IDs and account-list pagination values now receive clear validation errors.
  - Reconciliation errors consistently explain unknown or unavailable account keys.


jakobwennberg-oss:
Compliance Swarm's low finding (swallowed lookup failures in `unknownAccountKeyError`): fixed in 615957795, both catch paths now log a warn line with the company id and error before falling back to the same user-facing answer.

PR #2728, opened 2026-09-18
fix(import): name the real failure instead of the theater step, and let a read key run SIE preflight

jakobwennberg-oss:
## What

Three imports of the same Fortnox export failed with "Kontrollerar balansen: något gick fel" and 0 verifikat. That text is the import theater's last narration line plus the generic fallback. The validator had refused the file over one line-less voucher (#BTRANS only) dated in the previous year, and both routes replaced its sentence with registry text: `VALIDATION_ERROR` ("Förfrågan innehåller ogiltiga uppgifter.") in `/api/import/sie/execute`, and `SIE_IMPORT_UNEXPECTED` (500, "Importens resultat kunde inte bekräftas") in the migration wizard's `/import-sie`, with the reason buried in `details.reason` and the logs.

**Part (a), the failure message**
- `validateSIEJobInput` skips line-less vouchers in the size and date checks. Preparation skips them as empty and never posts them, so they cannot refuse the file either. A voucher with lines outside the fiscal year is still named (series, number, date).
- New `sieJobValidationResponse` answers a `SIEJobValidationError` with the validator's own sentence, code and request id. Used by both routes. A code with its own registry entry (`SIE_IMPORT_UNSUPPORTED_ACCOUNT_CLASS`) keeps that bilingual text and details. The wizard's unmapped-accounts 400 is now a structured `SIE_IMPORT_UNMAPPED_ACCOUNTS` envelope instead of the legacy `{ error: 'validation' }` tag the UI printed verbatim.
- New pure helper `lib/import/import-failure.ts`: `describeImportResponseFailure`, `describeSIEJobFailure`, `formatImportFailure`. One shape for the sentence, the details the server attached (unmapped accounts, per-voucher errors, a Swedish `details.reason` a fallback code wrapped, never a technical one) and a reference line ("Referens: kod X, ärende Y, import Z"). Rendered by the SIE wizard's review step, the migration wizard's result step and the job progress card.
- `waitForSIEJob` throws a typed `SIEJobFailedError` carrying the job and its description. The migration wizard rethrows it as user-facing instead of routing it through the Swedish-pattern heuristic that swallowed misses into "Något gick fel. Försök igen.".
- `sie_job.states.failed` is "Importen misslyckades" / "Import failed": a job can fail after it started.

What the failing case now says, on either surface:

```
Räkenskapsår 2026: SIE-verifikation LESSLIE2 (2025-01-02) ligger utanför räkenskapsåret.
Referens: kod VALIDATION_ERROR, ärende <requestId>
```

and, after the validator change, that particular file goes through, with the empty voucher counted under "hoppades över".

**Part (b), preflight with a read key**
- `gnubok_sie_preflight` and `gnubok_create_sie_upload` require `reports:read` (the upload was `bookkeeping:write`, preflight was unlisted). Preflight is analysis; the upload URL only stages bytes in the documents bucket (50 MB cap at bucket and validator, two-hour TTL, per-company pending prefix). `gnubok_import_sie` and `gnubok_undo_sie_import` stay `bookkeeping:write`. Descriptions updated within the tools/list budget. Decision recorded in DECISIONS.md.

## First principles

Why did this occur: the validator wrote a precise Swedish sentence, but the error envelope builders only know registry codes, so a generic code meant a generic sentence. The client side then had two more places to lose it: the migration wizard's `displayError` heuristic, and `waitForSIEJob` throwing a plain `Error`. Three layers each assumed the layer below had said what was wrong.

What can be removed: the wizard's hand-rolled `{ error: 'validation', message, unmappedAccounts }` shape (now the standard envelope), and the special-casing of `SIEJobValidationError` with details in the execute route (one helper for every validation error).

Not removed on purpose: `providerFailureResponse` for genuinely unexpected errors. Those still get the fallback code, but the reference line now carries the request id so support can find them.

## Tests

- `lib/import/__tests__/import-failure.test.ts` (helper), `sie-job-client.test.ts` (typed failure), `sie-job-input.test.ts` (empty voucher outside the year passes, a voucher with lines still refuses)
- `app/api/import/sie/execute/__tests__/jobs-routes.test.ts` and `extensions/general/arcim-migration/__tests__/import-sie-validation-reason.test.ts` (both routes carry the sentence; registry codes keep their text)
- `lib/auth/__tests__/scope-catalog.test.ts` (scope pins)
- Local: touched suites 263 files green, `npm run check:lint`, `npm run check:guards` pass. `npm run check:types` trips only on a pre-existing environment gap (`@types/nodemailer` not installed in the shared node_modules), not on this change.

## Context

Easy Online Stores evaluation brief 2026-09-16, finding F5.


jakobwennberg-oss:
On Superagent's P2 (`gnubok_create_sie_upload` under `reports:read`): kept as designed, with the bounds spelled out. The tool only mints a staging URL under the company's own pending prefix (tenant-scoped by the key's company), the bucket and the validator cap the object at 50 MB, objects expire after two hours, and the key is rate limited (100 requests per minute). Nothing reaches the ledger without `gnubok_import_sie`, which stays `bookkeeping:write` and stages a pending operation a human approves. A dedicated staging scope would add a fourth scope to explain for one read-side use (preflight of a file before a migration); the trade-off is recorded in DECISIONS.md.

PR #2729, opened 2026-09-18
fix(payroll): calculation corrections and categorized cutover balances

simhallq:
## Summary

General payroll calculation fixes and categorized opening-balance support. This is
a **draft for review**, not a claim of production readiness or payroll-provider
parity. The contribution contains synthetic tests only, with no customer source
data, reports, imports, credentials or local operational tooling.

- Fix signed absence double deductions, partial-hour absence, first-day sick pay
  and capped/cross-month waiting deductions.
- Preserve manual vacation payments using explicit engine-row provenance; avoid
  compensation accruing on itself.
- Add explicit employer calculation policies, partial-month display consistency,
  verified one-off tax percentages and effective-tax YTD/run totals.
- Preserve distinct paid, extra-paid, saved-by-year, unpaid and advance vacation
  balances at an independent cutoff, with dated withdrawals and review holds.
- Keep unknown historical net and monetary vacation liabilities null; expose the
  distinction in the API and native payslip PDF.

## Why did the problem occur?

Signed derived absence rows were also consumed as gross deductions. Recalculation
identified vacation compensation by item type, which conflated manual settlements
with generated payments. Some absence calculations counted dates rather than
scheduled hours, while YTD/run totals did not consistently honor effective tax
overrides. Flat vacation fields could not represent independent source cutoffs or
the distinct remaining-day categories; unavailable monetary history became zero.

## What could be removed or simplified?

Use a shared base-pay calculation for both display and calculation, one explicit
provenance marker for generated payments, and one authoritative categorized
remaining-day snapshot. Remove double consumption and implicit zero coercion
rather than add employee-specific corrections. Legacy flat fields remain for API
compatibility but cannot contradict a categorized snapshot.

## Why this solution rather than a workaround?

Hardcoded employee exceptions and target-total adjustments would hide underlying
errors and would not generalize. A global switch in calculation conventions would
also change unrelated employers. Explicit, snapshotted company policies retain
defaults while making alternatives reviewable. An alternative is a narrower PR
containing calculation fixes only; keeping the cutover work here shows the
integration boundaries and known blockers for maintainer review.

## Validation

- 1,679 targeted payroll, schema, API and PDF tests pass across 92 files.
- 11 rollback-only PostgreSQL constraint/persistence tests pass with synthetic data.
- Changed TypeScript files lint without errors; one pre-existing unused-variable
  warning remains in the schema test file.
- Antipattern guards and whitespace checks pass.
[...]

PR #2731, opened 2026-09-19
fix(salary): absence rows deducted once, sjuklön from day one with a capped karensavdrag, hour-weighted partial days

jakobwennberg-oss:
## What

Reported by Frey (Simon Hällqvist) after running the deviation-period build against a larger Fortnox payroll, with #2729 as a proposed fix. This PR takes the calculation part of that PR, with credit, and leaves the rest of #2729 (calculation-policy toggles, calendar-day long leave, one-off tax, categorized cutover balances, seven migrations) in review as its own change.

- **Double deduction.** The engine summed absence rows by type in Step 3 and then subtracted every row flagged `is_gross_deduction` again in Step 4; the derived sick, VAB and parental rows carry that flag. 31 500 kr with one VAB day gave 28 500 instead of 30 000. Step 4 now excludes the rows Step 3 consumed. The flag stays on the rows: it also drives the booking split in `salary-entries.ts`, so flipping it would have moved sjuklön and VAB between accounts in the same change.
- **Sick day one.** Sjuklön is paid from the first day and one karensavdrag of 20 % of a week's sjuklön is deducted per sjuklöneperiod, never more than the sjuklön the period yields; the remainder carries into the next sick day, also across a month boundary (SjLL 6 §, 2019 wording). The previous code was the pre-2019 karensdag shape and overpaid 0.2 day per period.
- **Partial days.** An absence row's hours against the scheduled hours per day (`hours_per_week / workdays_per_week`, default 8) weight the deduction, capped at one day. Reported day counts stay whole because AGI reports dates.

## First principles

Why it occurred: two independent conventions for "this row reduces gross" (a type list and a flag) were both consumed by the engine, and the derived rows satisfied both. The day-one shape was a simplification from the module's first version that never got the 2019 wording.

What was removed: nothing structural; the fix is one exclusion in Step 4 and the sjuklön model corrected in the derivation. Considered and rejected: dropping the flag from derived rows (changes booking), and a calculation-policy switch for the day-one convention (the law has one answer).

## Prod impact

Checked before the fix: one booked run with a double-deducted row, in the internal test company, no AGI filed. Ten draft runs recalculate on their next `:calculate`.

## Tests

Engine: Simon's exact scenario (31 500, one VAB day, gross 30 000) and a genuine bruttolöneavdrag next to absence rows. Derivation: day one included, karens cap on a one-hour first day, cross-month carry, no second karens when the lookback already absorbed it, hour-weighted VAB/parental/unpaid days, schedule-length day, existing expectations updated to the day-one model.

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented absence-related deductions from being counted twice in salary calculations.
  - Sick leave now includes sick pay from the first day, with qualifying deductions capped and carried forward when necessary, including across month boundaries.
  - Partial-day absences now calculate pay and deductions proportionally to scheduled working hours.
  - Improved absence handling for sick leave, care of sick child, parental leave, and unpaid leave.
  - Corrected reported sick-leave costs by accounting for qualifying deductions.
  - Improved high-risk protection handling within the same month.
  - Preserved vacation qualification based on whole absence dates.


jakobwennberg-oss:
Review triage (9b27594):

**Fixed**
- Step 4 exclusion keyed on one `ReadonlySet` of absence item types instead of array identity (compliance bot, finding 1).
- FK499 sjuklönekostnad now subtracts the karensavdrag rows, since the day 1-14 row includes day one (note 2; the AGI generator was the one consumer that priced that row).
- `hours_per_week` / `workdays_per_week` of 0 fall back with an explicit `> 0` check (note 3).
- Added the cap test: two new sjuklöneperioder in one month past the högriskskydd cap are both suppressed and the day 1-14 row still counts both days (note 4).

Other consumers of `sick_day2_14` (account mapping 7281, the payslip line label, the schema enum, the run-calculation type list) do not depend on day one being excluded.

jakobwennberg-oss:
Review triage (dfda08785):

**Fixed**
- 120-day semestergrundande cap counts whole dates while the deduction uses weighted days (derive:391). Test added: 119 dates YTD plus two half days is 121 dates, not semestergrundande.

**Not changed, with reason**
- Persisting the karens amount consumed per sjuklöneperiod (derive:299): correct observation, but it needs a per-period state table, which is a schema change and does not belong in this bug fix. Today the carry is reconstructed from prior sick hours at the current rate; it is exact unless salary or schedule changes inside one period that straddles a month boundary. Noted next to the existing TODO on persisting per-period karens state.

jakobwennberg-oss:
Review triage (dfda087):

**Not changed, with reason**
- Splitting the VAB/parental row at the 120th day (derive:386): correct, and pre-existing. The row has always been one aggregate per period with a whole-row qualification; this PR only made the count use dates instead of weighted hours. Filed as https://github.com/erp-mafia/accounted/issues/2732 so it lands as its own small change with its own tests.

Issue #2695, opened 2026-09-17
salary: removing an employee benefit leaves its derived payslip line on recalculate (ON DELETE SET NULL defeats the provenance filter)


joakimhew:
## Summary

Removing a benefit from an employee (Löner → Anställda → Förmåner → Ta bort) and then pressing "Räkna om" on a draft salary run does not remove the derived benefit line from the payslip. The taxable benefit, its tax deduction and the employer contributions stay on the run. Adding a benefit and recalculating works, and ending a benefit by setting "Gäller till" works, so only the delete path is affected.

Observed on main as of 2026-09-16 (self-hosted, synced to `91ab339a8` plus the six commits after it).

## Steps to reproduce

1. Create a draft salary run for a month.
2. On an employee, add a benefit (e.g. Bilförmån, monthly value 3 000 kr, "Gäller från" before the run's payment date).
3. Open the run and press "Räkna om": a `benefit_car` line appears. Correct.
4. Go back to the employee and remove the benefit with "Ta bort".
5. Open the run and press "Räkna om" again.

Expected: the `benefit_car` line is gone and gross, tax and avgifter are recomputed without it.
Actual: the line stays, with the benefit still taxed and still in the avgift basis. It now looks like a manually added line and can only be removed by hand.

## Root cause

Three pieces that are each reasonable on their own:

- `app/api/salary/employees/[id]/benefits/[benefitId]/route.ts`: `DELETE` hard-deletes the `employee_benefits` row.
- `supabase/migrations/20260512200100_employee_benefits.sql`: `salary_line_items.source_benefit_id uuid REFERENCES employee_benefits(id) ON DELETE SET NULL`. The hard delete therefore nulls `source_benefit_id` on every derived line that pointed at the benefit.
- `lib/salary/run-calculation.ts`, step 8d: derived benefit rows are replaced by first deleting `salary_line_items` where `salary_run_employee_id = sre.id AND source_benefit_id IS NOT NULL`, then inserting rows for the currently active benefits. After the `SET NULL`, the stale line no longer matches the delete filter, so it survives every recalculation.

The update path is unaffected: setting `valid_to` keeps the row, the `valid_to >= payment_date` filter excludes it, and the delete filter still matches the derived line by its intact `source_benefit_id`.

## Impact

[...]

PR #2734, opened 2026-09-19
feat(api): payroll gap-closure 5 on v1: employee benefits, recurring lines, correct a booked run

jakobwennberg-oss:
## What

The remaining employee-side inputs and the one lifecycle verb an external payroll operator (Frey) needs after booking, all under the existing `payroll:read` / `payroll:write` scopes.

- `GET/POST /companies/{id}/employees/{id}/benefits`, `PATCH/DELETE .../benefits/{benefitId}`: förmåner the engine reads at calculate time. Shared module `lib/salary/employee-benefits.ts`; the dashboard routes call it and keep their responses byte for byte.
- `GET/POST /companies/{id}/employees/{id}/recurring-lines`, `PATCH/DELETE .../recurring-lines/{lineId}`: standing monthly rows, with the amount-sign and validity rules moved into `lib/salary/employee-recurring-lines.ts`. Delete falls back to deactivation when a payslip line references the row, as the dashboard does.
- `POST /companies/{id}/salary-runs/{id}/correct`: the rättelsekörning. `lib/salary/correct-run.ts` extracts the dashboard's exact write order (storno every entry via `reverseEntry`, flip the original to `corrected`, revoke payslip links, insert the same-period draft, copy the roster) with a dry-run preview and the partial-failure states documented in the module doc.
- Onboarding cookbook documents the three; `skills/accounted-api` regenerated; spec snapshot 154 to 163 endpoints.

## First principles

Why it occurred: the v1 payroll surface was closed from the run outwards; these three registers were still dashboard-only and each had its logic inline in the dashboard route, so exposing them meant extracting a shared module first (the `absence.ts` shape).

What changed on purpose: a retried `:correct` now resumes past entries an earlier attempt already reversed (`CannotReverseNonPostedError` with status `reversed` counts as done). Before, a storno that failed halfway left a run that could never be corrected through the verb again. The internal recurring-lines GET answers 404 for a foreign employee instead of an empty list, matching every sibling. See DECISIONS.md 2026-09-19.

Known, not fixed here: #2695 (removing a benefit leaves its derived payslip line on recalculate) is stated in the benefits pitfalls.

## Tests

Module tests for all three libs, route tests for every verb (401, 403, 400, 404, 409, happy path, dry-run makes no write), the dashboard routes' existing tests unchanged, scope parity, OpenAPI examples and request schemas, spec snapshot reviewed.

## Summary by CodeRabbit

- **New Features**
  - Added employee benefits APIs for listing, creating, updating, deleting, filtering, and dry-run previews.
  - Added recurring salary-line APIs with validation, filtering, updates, deletion, and deactivation fallback when referenced by payroll records.
  - Added salary-run correction for booked runs, creating a linked draft correction run and supporting dry-run previews.
  - Added idempotency, scope checks, and structured error responses across new endpoints.

- **Bug Fixes**
  - Retried salary corrections now resume successfully after partially completed reversals.

- **Documentation**
  - Updated payroll API references and onboarding guidance for benefits, recurring lines, and salary-run corrections.


jakobwennberg-oss:
Review triage (aef983798):

**Fixed**
- `:correct` now also resumes when the original is already `corrected` but has no correction run (insert failed after the storno and flip): it creates the missing draft without touching the ledger (correct-run.ts:430). Test added.
- Benefit DELETE keeps and deactivates a row that a payslip line already derives from, instead of hard-deleting (compliance bot finding 1; the provenance FK is SET NULL so the database alone would not block it). Answers 200 with `{ employee_benefit_id, deleted, deactivated }`, aligned with recurring lines. DECISIONS.md updated to match (compliance bot finding 2).
- Empty benefit patch is a no-op returning the stored row (employee-benefits.ts:375).
- Recurring-line failure details carry the SQLSTATE only; the PostgreSQL message stays on `cause` for the dashboard (employee-recurring-lines.ts:115).
- Recurring-line create marked `idempotent: true` (recurring-lines/route.ts:256); `:correct` retry pitfall rewritten to the resume behaviour (correct/route.ts:71); cookbook no longer says to attach the roster again (cookbook:156).

Issue #2732, opened 2026-09-19
salary: split a VAB or parental absence row at the 120-day semestergrundande threshold


jakobwennberg-oss:
From CodeRabbit on #2731. Pre-existing behaviour: the derived VAB and parental-leave rows are one aggregate per pay period, and `is_vacation_basis` is decided for the whole row (`ytd + dates <= 120`, SemL 17 and 17 a §). When the 120th day falls inside the period, the dates up to and including day 120 still qualify and the ones after do not, so the row should be split in two (quantity and amount prorated by each date's weighted share), which also gives the vacation ledger the right basis.

Small, self-contained change in `lib/salary/derive-absence-line-items.ts` with tests for 119 YTD + two dates (one qualifying row, one not).

Issue #2724, opened 2026-09-18
salary: archive the generated payment file (pain.001 / LB) instead of only stamping the run


jakobwennberg-oss:
Raised by the Swedish compliance review bot on #2722. Inherited behaviour, not introduced there, but the new `POST /salary-runs/{id}/payment-file` widens third-party access to it.

**Today**: both dashboard routes and the v1 endpoint (`lib/salary/payment/build-payment-file.ts`) return the file inline and stamp `payment_file_format` / `payment_file_generated_at` on `salary_runs`. The bytes that went to the bank are not stored. If an employee's account or the company's IBAN changes, a regenerated file differs from the one transmitted and no copy of the original exists.

**Wanted**: store each generated file as a document attachment linked to the run (WORM, 7-year retention per BFL 7 kap 1 §), the way `supplier_payment_batches` keeps an immutable pain.001 snapshot for supplier payments (DECISIONS 2026-09-06). Return `document_id` on the v1 response so an operator can prove which file was sent.

Related: #2722.

Issue #2723, opened 2026-09-18
salary: absence and worked-days writes inside a non-draft run's window need a guard or a change history


jakobwennberg-oss:
Raised by the Swedish compliance review bot on #2722 for the new v1 worked-days endpoint; the same hole exists for absence (`lib/salary/absence.ts`, v1 and dashboard), so this is one fix for both registers.

**Today**: `PUT`/`DELETE` on `/employees/{id}/absence` and `/employees/{id}/worked-days` (and the dashboard routes) accept dates that a run in status review/approved/paid/booked already read through its avvikelseperiod window. The booked payslip figures are snapshotted in `salary_line_items` / `salary_run_employees`, so the verifikat does not change, but the register no longer matches what was paid and nothing records who changed it or when.

**Wanted**: either refuse writes on dates covered by a non-draft run's `deviation_period_start..end` (409 with the run id, like the opening-balance lock after the first booking), or keep a who/when history per row (the inline rättelse log pattern). Whichever, apply it in the two shared libs so dashboard, v1 and MCP behave the same.

Related: #2720 (avvikelseperiod), #2722 (worked days on v1).

jakobwennberg-oss:
Follow-up from #2735 (Superagent P2): the application-side lock is not atomic with the write. If a race ever shows up, the airtight form is a BEFORE INSERT/UPDATE/DELETE trigger on salary_absence_days and salary_worked_days that refuses dates inside a non-draft run's deviation window, with pg-real tests.

PR #2722, opened 2026-09-18
feat(api): payroll operator onboarding on v1: salary settings, worked days, payment file

jakobwennberg-oss:
## What

Closes the three public-API gaps an external payroll operator still had to do in the dashboard, so a bureau or an AI operator like Frey can provision and run a customer end to end over `/api/v1`. Stacked on #2720 (avvikelseperiod); base moves to `main` when that merges.

- `GET/PATCH /companies/{id}/salary/settings` (`payroll:read` / `payroll:write`): pay day, avvikelseperiod, payment file format, default bank, öre rounding, salary voucher series. Upserts a missing settings row; merges the voucher series into the per-source-type map. This is Simon's "kundprovisionering" item: the one place a customer's payroll behaviour is configured, settable before the first run.
- `GET/PUT/DELETE /companies/{id}/employees/{id}/worked-days` (`payroll:read` / `payroll:write`): the tidrapport register (`salary_worked_days`), so hourly staff and OB/shift premiums can be driven over the API. Logic extracted into `lib/salary/worked-days.ts` (mirrors `absence.ts`); the two internal worked-hours routes now call it.
- `POST /companies/{id}/salary-runs/{id}/payment-file` (`payroll:write`, idempotent, dry-runnable): the ISO 20022 pain.001 or Bankgirot LB salary payment file for a run, returned inline like `generate-agi`. Same preconditions and run stamping as the internal routes.
- Docs: new cookbook "Onboard a payroll customer and run the first month" (company, settings, employees, cutover balances, absence + worked days, run, payment file, book, AGI, year end), payment-file step in the existing payroll cookbook, regenerated `skills/accounted-api`, spec snapshot updated.

## First principles

Why it occurred: the v1 payroll surface was built run-first (Phase 5) and closed gaps from the run outwards (payslips, lines, absence, cutover, vacation). The operator's onboarding side (company settings) and the two "physical" ends of the loop (hours in, bank file out) were still dashboard-only, which is exactly where an API-only operator stops.

What was removed: the inline `salary_worked_days` queries in the two internal routes, replaced by one command module shared with v1, the same shape absence already had. Payment file generation stays in its pure generators; the routes only load and validate.

Not in scope (follow-ups, listed in the issue tracker if wanted): employee benefits CRUD on v1, recurring lines on v1, `:correct` (rättelsekörning) on v1, AGI status on v1 (submission stays BankID by design), a company-wide vacation balance list, and MCP tools for worked days.

## Tests

[...]

PR #2735, opened 2026-09-19
fix(salary): lock register dates under non-draft runs, archive every payment file, split absence rows at day 120

jakobwennberg-oss:
## What

Three follow-ups filed from the review bots on #2722 and #2731. All three were pre-existing behaviour that the new public endpoints made more visible.

- **#2723 locked dates.** Absence and worked-days writes on dates inside the avvikelseperiod of a run in review, approved, paid or booked are refused with `409 SALARY_REGISTER_DATES_LOCKED_BY_RUN`, naming the run and the locked dates. One helper (`lib/salary/register-locks.ts`) behind both command modules, so dashboard, v1 and MCP agree. Drafts and corrected originals never lock; the way out is revert-to-draft or `:correct`.
- **#2724 payment file archive.** Every generated pain.001 / Bankgirot LB file is archived as an immutable `salary_payment_files` row before it is returned (WORM trigger with the sandbox-teardown escape, RLS insert and select only, sha256 and byte_size over the exact bytes the download carries). The v1 response carries `payment_file_id` and `sha256`; `GET /salary-runs/{id}/payment-files` reads the archive back. Archive contract, migration-reset and account-erasure tests cover the new table.
- **#2732 day-120 split.** A VAB or parental-leave period that crosses the 120th semestergrundande date splits into a qualifying and a non-qualifying row, each priced on its own weighted days (SemL 17, 17 a §).

## First principles

Why they occurred: the registers were designed for a dashboard where a human sees the run state before editing; the API has no such glance. The payment file was returned inline because the dashboard downloaded it and nothing else needed it. The 120-day cap was evaluated per aggregate row because rows were whole days.

What was removed: nothing. Rejected: a who/when history for register rows (the payslip lines already snapshot what was paid; the guard removes the divergence instead of recording it). See DECISIONS.md 2026-09-19.

## Migration

`20260919105035_salary_payment_files.sql`, new table only, verified on a fresh replay of all migrations locally; auto-applies on merge.

## Tests

Lock helper unit tests, locked cases on both lib modules, both v1 routes, the dashboard routes and the MCP executors; archive builder tests (row shape, sha256 per charset, insert failure fails the build, dry run archives nothing), list route tests, `tests/pg/salary-payment-files.pg.test.ts` (RLS, WORM for superuser and member, cross-company refusal, digest and charset checks, run delete blocked, sandbox purge); split cases for VAB and parental leave.

## Summary by CodeRabbit

- **New Features**
  - Generated salary payment files are now archived immutably with checksums, encoding, and file metadata before delivery.
  - Added an API endpoint to view and paginate archived payment files.
  - VAB and parental-leave calculations now split entries at the 120-day qualifying limit.

- **Bug Fixes**
  - Absence and worked-day changes are blocked for dates read by salary runs in review, approved, paid, or booked status.
  - Blocked changes return HTTP 409 details identifying the locking run and dates; drafts remain editable.
  - Payment files are withheld if archival fails to prevent untracked bank submissions.

- **Documentation**
  - Updated API documentation for register locks, payment-file archiving, and correction workflows.


jakobwennberg-oss:
Review triage (3370148):

**Fixed earlier in this PR**
- `salary_payment_files.user_id` is `ON DELETE RESTRICT`, with a pg test that a hard delete of the author is refused while the archive references them (Superagent P1 on the migration).

**Accepted with reason**
- Lock check and register write are not atomic (Superagent P2, absence.ts:171): the window is the milliseconds between the `salary_runs` read and the upsert, and the only way through it is a run of the same company moving from draft to review at that instant. The guard's job is to stop the everyday case (editing August after September's run was approved), which it does; the airtight version is a database-side trigger on both register tables, which is a migration plus pg-real tests of its own. Noted in #2723 as the follow-up if it ever bites; not blocking this fix.

PR #2740, opened 2026-09-19
feat(salary): calculation policies for Fortnox parity, one-off tax per line, engine-owned vacation rows

jakobwennberg-oss:
## What

The calculation part of #2729 (Simon Hällqvist, Frey), ported onto today's main with Simon as co-author, following the founder's decision to offer every convention it proposes as a company policy (Fortnox parity). Every default reproduces the historical engine byte for byte; tests assert deep equality of engine and derivation output with and without the default policy.

- `company_settings.salary_calculation_policy` (jsonb, key set and enums closed by CHECK): `partial_month` (workdays | annual_calendar_days), `sick_rate` (daily_divisor | annual_hourly), `long_leave` (workdays | calendar_after_five_workdays) with `leave_context` (all_registered | through_deviation_end), `net_rounding` (up | nearest), `one_off_tax_rounding` (truncate | nearest). Read at `:calculate`, snapshotted into `calculation_params`, exposed on `GET/PATCH /salary/settings` (merged key by key, so one convention cannot reset the others) and as one select per convention under Inställningar > Lön.
- `salary_line_items.one_off_tax_percent`: engångsskatt on a payslip line, withheld per rate group with öretal bortfaller by default; internal, v1 and MCP line writes validate it.
- `salary_line_items.calculation_source`: the engine stamps its own semesterersättning row and deletes only that on recalculation, so a manual vacation payment (slutlön) survives. Pre-existing engine rows are backfilled by their exact signature; no hard block on legacy rows.
- Engine additions now count signed taxable `other`/`correction` rows (they were displayed but ignored before) and semesterersättning is excluded from the vacation basis (SemL 16 §). The displayed Grundlön row prorates a partial month the way the engine already did (display only).

## First principles

Why it occurred: the engine had one convention per question, hard-coded, and the module's first version simplified where Fortnox offers a choice. The founder chose parity, so the conventions become explicit, snapshotted policies rather than code.

What was removed: nothing structural. Not taken from #2729, on purpose: `vacation_movements`, the effective-tax override in totals and YTD, nullable `ytd_net`, and the hard `assertVacationLineProvenance` block (a UI-less dead end for legacy rows); those belong to the cutover track, which follows as its own PR stacked on this one.

## Migrations

Four new files (`20260919120000` to `20260919120300`; NOT VALID + separate VALIDATE for the two line CHECKs, the repo's existing split). Replayed on a fresh Postgres with every migration; all constraints present and validated. Auto-apply on merge.

## Tests

Policy schema, one-off helpers, calendar leave, `monthlyBaseSalary`; engine and derivation policy cases plus default-equality assertions; payslip lines, build-payslip-data, schemas, v1 settings route; pg test for the CHECKs. Full unit suite green (24,704 tests); guards, lint, types clean; i18n parity (26 keys each).


jakobwennberg-oss:
Review triage:

**Accepted with reason (compliance bot, both marked unsure)**
- `one_off_tax_rounding: 'nearest'`: the statutory path stays the default (`truncate`, öretal bortfaller) and every run snapshots the convention it used. `nearest` exists only because the founder chose Fortnox parity for companies migrating mid-year whose earlier payslips were computed that way; the docs and the settings help text call it a compatibility convention, not a statutory one.
- Operator-supplied engångsskatt percentage: same model as jämkning, deliberately. The percentage comes from Skatteverket's engångsbelopp table for the employee's estimated annual income, which the operator (or the employee's beslut) determines; the API validates range, sign, taxability and item type and refuses it on rows where it makes no sense. Deriving it from a table lookup is a possible follow-up once the annual-income estimate is a first-class field.

